PR #10, opened 2014-09-29
Using Express routes

nicola:
Using Express routes and componentize library.

I did: 
- rename server into index.js so this library can be imported in other projects
- create an Express router and expose that in `index.js`
- move all the server code into `examples/server.js`


nicola:
updated version is on #15

PR #13, opened 2014-10-18
avoid usage of global variables

nicola:
In addition, patchType and fileType are never used


timbl:
Folded these in a whole ago

Issue #9, opened 2014-09-29
Implementing a route


nicola:
The idea is to use routes instead of implementing it straight into the main toutes

```javascript
// in server.js or whoever will implement it
app.use('/what_i_desire', routes);
```

```
// in our library
var express = require('express');
var router = express.Router();

/* GET home page. */
router.get('*', function(req, res) {
});

router.patch('*', function(req, res) {
});

module.exports = router;
```

In this way, we can make something that any user can plug and play in their environment

This is only available in the latest `express`. Good idea to upgrade?

nicola:
(Updated version of this PR is on #15)

Sorry for the delay in this, I just got back from a long journey with little connection.

Express routes are very simple, think of them as a volume that you mount.
**Theory.** So, if you have on a drive `/A/B` then you can mount it on `/home/nicola` so it will have the final path `/home/nicola/A/B`. The equivalent in our case would be `/the_path_that_i_decide` mounts for example `app.get('*')`, meaning that I can mount the Linked Data Platform in top route that I decide.

**Practice**. As in my #15 PR, nothing really changes to the code written so far. The issue of the order and the middleware you mention still applies to the current implementation. What the PR does, it encapsulates the LDP code into a route that we can decide where to mount.

**Advantage**. The main benefit is that we can package the current implementation and it can be easily mounted, instead of hardcoding the path or re-editing the "demo".

```javascript
var ldp = require('ldp-httpd')
app.use('/here_I_want_ldp', ldp);
```

martinmr:
Speaking about upgrading to Express 4.x, I did it a few days ago to see if it worked. I didn't have to make any changes other than updating the package.json file.

martinmr:
Routes are already implemented. Closing issue.

nicola:
Can you discuss how they are implemented?

I can see in [server.js](https://github.com/linkeddata/node-ldp-httpd/blob/master/server.js) that I can add them one by one, but there is no such concept of router I described above, making this library strictly a demo (or a ready-to-use tool), but not a library I can implement in my current codebase.

nicola:
I found out something great related to this (that I am working on my fork right now).
An express app can be loaded as a router, this means that we just need to do app.use('/whereIwantMyLDP', ldp(opts))

martinmr:
Closing this issue. At this time, this feature has been implemented and it's pushed to master.

PR #15, opened 2014-12-04
Making it modular with express routes

nicola:
Should I make this working for liveUpdates instead?


nicola:
This is explained in #9

timbl:
I'm afrad I have just been workig in liveUpdates and there may be bits which the master branch does not have and should have.  Like I just put the proxy into liveUpdates.  We should merge liveUpdates back onto the main branch I think

nicola:
Ok, we then wait for liveUpdate to be merged for this merge

timbl:
Ok I have merged liveUpdate into master, do go ahead make it router-happy now

nicola:
So I have been implementing it and it worked apart from the new websocket.

I haven't had time until now (I have now substantially less workload at uni), but I have concern with our usage of the library `express-ws`, I have not clear whether our websocket server **must** be the same or if it can be on a different machine or on a different port.

So, assuming we want to achieve the best flexibility, I submitted an issue to the maintainer of express-ws, that might help, in there I suggest three options that differ from the one that I suggest in this PR.

https://github.com/HenningM/express-ws/issues/8

nicola:
working on a new version

PR #55, opened 2015-07-12
Use system-wide ACL notation in ACL module

deiu:
The ACL module was using a hardcoded notation (i.e. ".acl") for ACL suffixes, although the aclSuffix param is available at runtime.


nicola:
Since this could not automatically be merged and since I had to add them in the command line, I did a PR following what you did here, https://github.com/linkeddata/ldnode/pull/56

So, the reviewer should close yours and merge mine.

PR #58, opened 2015-07-13
Removing uriBase/uri

nicola:
Now the uriBase is set by the incoming request.

To load ldp on `http://nicola.io/LDP` and map it to /home/nicola, run

```
ldnode --mount /LDP --base /home/nicola
```

To load ldp on `http://nicola.io/` and your current working dir is `/home/nicola`, run 

```
ldnode
```

:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,7 @@ npm install
 ```javascript
 var ldnode = require('ldnode')
 
-var ldp = ldnode.createServer({
-  uri: "http://example.com/test/",
-  base: __dirname + '/test/'
-})
+var ldp = ldnode.createServer()
 ldp.listen(1234, function() {
   // Started Linked Data Platform
 })
@@ -46,7 +43,7 @@ You can integrate it with your existing express app
 ```javascript
 var ldnode = require('ldnode')
 var app = require('express')()
-app.use('/test', ldnode(opts))
+app.use('/test', ldnode())
 ...
 ```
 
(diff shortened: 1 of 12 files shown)


martinmr:
Does this cover every appearance of uriBase?

nicola:
yes absolutely, it fixes https://github.com/linkeddata/ldnode/issues/47

deiu:
The changes seem reasonable. Quick question though: why is this function still here: https://github.com/nicola/ldnode/blob/no_uriBase/container.js#L21?

martinmr:
I think the only issue that should be fixed before merging is the last one. I plan to remove the root container issue (along with all the other issues involving containers and the meta file) tomorrow and it's an entirely separate issue.

nicola:
overall comment: @deiu your changes were not rolled back and this is a PR for baseUri not for other issues:

1. that function is just a function that creates the root container. the `baseUri` mentioned in there is just the name of the variable, nothing to do with the `options.baseUri`. It is only used in by @martinmr in https://github.com/nicola/ldnode/blob/no_uriBase/index.js#L51
2. That exact acl suffix has not been mentioned in your earlier PR (we only touched here https://github.com/linkeddata/ldnode/commit/cfbc744f6b2e266837dc2608d832a1c9842db6aa), so I guess we both skipped it! This should be a different issue I just created let's discuss this elsewhere #59 #60
3. This should be discussed in #60 and in a different PR
4. No changes have been rolled back

PR #69, opened 2015-07-14
WebID delegation and tests

martinmr:
I added the code for WebID delegation and tests, as well as the cleanup tests.
One of the webID tests is commented out because of the fetcher issue and the fix should be part of other commit.

All the other tests pass just fine. This completes all the tests that are included in gold.


deiu:
3 failing tests? :)

martinmr:
I have no idea why it's happening. All the tests pass on my machine.

PR #70, opened 2015-07-14
Minor changes to ACL tests.

deiu:
* Renamed test file from `secure-test.js` to `acl.js`
 * Added 2 more depth levels (/dir1/dir2/) to defaultForNew test
 * Fixed cleanup tests -- they need to be synced, and it's faster to use the `fs` methods


nicola:
Finally a clean up, although there is area of improvement in this tests, the fact that they are waterfall tests  make them difficult to run individually. This means that each test that alter context, should revert the context back how it was before, see https://github.com/linkeddata/ldnode/blob/master/test/patch.js

PR #77, opened 2015-07-14
change session to generate unique session IDs and secret.

martinmr:


martinmr:
That's fine by me.

index.js
```diff
@@ -40,9 +41,12 @@ function ldnode (argv) {
     // Setting options as local variable
     app.locals.ldp = opts;
 
-    // Session [TODO]
+    // Session
     app.use(session({
-        secret: opts.secret || 'node-ldp',
+        genid: function() {
```
deiu: What is `genid` used for, and is it used anywhere else?

nicola: This genid is not needed since it is already implemented in here:
https://github.com/expressjs/session/blob/master/index.js#L445-L447

index.js
```diff
@@ -40,9 +41,12 @@ function ldnode (argv) {
...
+            return uuid.v1();
+        },
+        secret: opts.secret || uuid.v1(),
```
nicola: r+ I like this

nicola:
can you remove genid: and its function? This is a merge without it

martinmr:
Done.

PR #82, opened 2015-07-15
Removing unused code in `vocab/rdf.js`

nicola:


nicola:
Interesting, I will leave it for now

PR #84, opened 2015-07-16
Fixing abs/originalUrl path issue

nicola:


nicola:
The tests are ensure it partially does the job. I want to know what was @timbl 's initial issue.

timbl:
My initial issue was that the code incorrectly reconstructed the full request URI.  It used to use the --uriBase parameter, but then it switched to using   req.protocol + '://' + req.get('host') + req.baseURI or something where.baseURI was never set.   I removed the function from filestore.js as it is not filestore specific.   I replaced each call with an inline  req.protocol + '://' + req.get('host') + options.mount which seems to work...

Maybe the basic tests or a subset should be run with a non-empty --mount value to catch the code which deals with it.   The patch code was what failed, as in that case the base URI is important for the re-serializing

Tim

nicola:
`req.baseURI`  is set by Express js and it is the equivalent of `opts.mount`. The latter is the option, the first one is the actual setting on express.

In my PR I should have fixed that, same for the tests. In addition, I think the original code worked, but the logs were showing the wrong path. In facts, on express, `req.path` is not the main path, while `req.originalUrl` is. So, I fixed that too!

timbl:
Other things I rectified (see gitter.im) were: the function uriBase was in filestore.js which was (a) in filestore.js which was irrelevant to the filestore as opposed to any other store, and (b) was nonsense. Neither from is name nor any comments was it clear what it should actually do either. It was exprted as uriAbs though internally was uriBase.  Notice in 7eedf7270ca127960d376e5c704f92fae01da8c2 there are now two functions in filestore.js and there is a req.baseUrl used by never set.  I cleaned this up butit was reverted.

nicola:
a) I was going to comment on this. The file `file.js` would eventually be renamed or moved since the only thing it does is path conversion, this is why I decided to use it there.

b) Fixed that!

c) req.baseUrl is set by express and it is set when:

  ```javascript
app.use('/hello', ldnode())
// or
ldnode.createServer({mount: 'hello'})
// or
$ ldnode --mount /hello
  ```

  `req.baseUrl` is set by express to be '/hello' in all these cases, while opts.mount is only set in the last two, and it has no control on the first case. This is becase it is never passed. In fact opts.mount is used to decide where to mount (equivalent to `req.baseUrl`)

PR #87, opened 2015-07-17
Document fetcher

martinmr:
The acl document can now fetch documents so the groups tests all pass.

Note. I rebased the changes in master because I thought I had not merged the changes nicola made to the tests. I don't think this will affect anything since git states the changes can be merged, but if there's any problem I can create a new branch and recreate my work there.


nicola:
I think everything is valid, but we should avoid `sync`, and start to use callbacks.

acl.js
```diff
@@ -284,11 +286,46 @@ function allowOrigin(mode, userId, res, aclGraph, subject) {
...
+function fetchDocument(graph, uri, req) {
+    var options = req.app.locals.ldp;
+    var uriBase = file.uriBase(req);
+
+    var body = "";
+    if (S(uri).startsWith(uriBase)) {
+        try {
+            var oldPath = req.url;
+            var newPath = S(uri).chompLeft(uriBase).s;
+            var documentPath = file.uriToFilename(newPath, options.root);
```
nicola: `newPath` is the equivalent of `req.path`, can we use that instead?

martinmr: It's not. New path is the path to the document being fetched (e.g. the
group file) while req.path is the path to the current request (e.g
ABC.ttl). The code needs to change the URL attribute of req because allow
takes a req object as its argument.

nicola: Sorry, I just re-read the code. I just saw the `S(uri)`, my bad.

I was pointing out the following, but I realized now, it is not inherent!

```
original URL == 'https://localhost:2000/myMountPath/path/to/abc.ttl'
req.originUrl == '/myMountPath/path/to/abc.ttl'
req.path == '/path/to/abc.ttl'
file.uriBase(req) == 'https://localhost:2000/myMountPath'
```

r+

PR #88, opened 2015-07-17
Simplifying login

nicola:
Mainly it removes `setEmptySession` to avoid setting `req.session.webid = ""`, whose type is string, instead of `undefined`.

**Only merge on tests passing**


nicola:
Not a trivial fix for now

PR #92, opened 2015-07-17
Adding cors support and default

nicola:
- [x] cors on apps
- [x] tests
- [x] make it identical to gold
- [x] OPTIONS support

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,7 +51,8 @@ function ldnode (argv) {
         ],
         exposedHeaders: 'User, Location, Link, Vary, Last-Modified, Content-Length',
         credentials: true,
-        maxAge: 1728000
+        maxAge: 1728000,
+        origin: true
     }));
 
     // Session
--- a/index.js
+++ b/index.js
@@ -42,18 +42,6 @@ function ldnode (argv) {
     // Setting options as local variable
     app.locals.ldp = opts;
 
-    // Setting CORS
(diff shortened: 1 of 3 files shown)


deiu:
Where is the test? :)

nicola:
this is the right question, coming

nicola:
I just realized looking at Gold, that this library is not enough, since it only sets `Allow-Control-Origin: *`

index.js
```diff
@@ -41,6 +42,19 @@ function ldnode (argv) {
     // Setting options as local variable
     app.locals.ldp = opts;
 
+    // Setting CORS
+    app.use(cors({
+        methods: [
+            'OPTIONS', 'HEAD', 'GET',
+            'PATCH', 'POST', 'PUT', 'MKCOL', 'DELETE',
+            'COPY', 'MOVE', 'LOCK', 'UNLOCK'
```
deiu: Are all those methods supported by ldnode? MKCOL, COPY, MOVE, LOCK, UNLOCK are methods specific to webdav.

nicola: These methods are not supported yet. I just wanted to align them with gold, so I will just remove them. I am implementing options at the moment.

nicola: Actually cors() implements a trivial 'OPTIONS' handler by default - but we have to linkeddata-ize it!

PR #93, opened 2015-07-18
ACL path

martinmr:
Turns out I had correctly transcribed the code from gold into ldnode but go's Dir and node's dirname have different behaviour when the path has a trailing slash.

In gold
```
pathfile.Dir("/home/martinmr/) // returns "/home/martinmr
```
while in node
```
path.dirname("/home/martinmr/") // returns "/home"
```


nicola:
Perfect

PR #97, opened 2015-07-19
Decoupling fs interaction and express handlers to avoid `sync` operations

nicola:
In essence, this aims at:

- Removing `sync` operations
- Decoupling express from the ldp operation with the filesystem operation, meaning that it would be simpler  to eventually implement in other platforms (e.g. Hapi, etc..)
- Decoupling express from ldp operation with the filesystem, also, we can implement LDP that runs on a database (e.g. mysql or mongodb) as an alternative to the filesystem.
- Use a callback style (the nodejs way) of writing code
- Having a more structured code, e.g. LDP class, ACL class etc..

The subtasks are:

- [x] make a LDP Class in ldp.js using options.js
  I tackle this by making of `options.js` a `ldp.js`. From now on there are no options, but instead a `ldp = new LDP(argv)` that holds the state of our LDP. LDP instances also have methods for example: `ldp.listContainer`, `ldp.readFile`
- [x] bringing container.js and metadata.js functions into ldp.js as methods
- all the functions in handlers should not have be explicit on fs, and should use `ldp.*` functions instead
  - [x] get
  - [x] patch
  - [x] post
  - [x] put
- [x] async ACL
- [x] removing all the sync operations
- [x] align with master


nicola:
Although this pass the tests, this needs accurate testing.
In this commit I show how to handle the async/sync issue.
However, I only worked on `handlers/get`, and all the other should be written in the same way

nicola:
I am carrying out the last bit which is ACL in a separate branch that I will merge to this one, once done, https://github.com/nicola/ldnode/tree/ldpfs-acl

nicola:
ACL is now working (it passes all the tests). The change has been radical and a lot of code has been cut out. I followed line by line the gold implementation to make sure it was correct.

However, further tests and usage will be a solid way of testing this re-written version

timbl:
Also, for the solid spec a concise write-up of the algorithm so others implementing it will do the same.
Also for apps which set ACLs on behalf of their users, the way to follow your nose to the ACL file and modify the ACLs.  Apps might want to make a directory subtree and set it all the same, or do it file by file.

timbl:
Does removing sync file access mean you have to add file locks to the PATCH code -- it used to just run sync from reading the file to writing the patched file back.

nicola:
@timbl, this is a good question. I believe that in the sync version, you still would incur in this issue (that I have to investigate as well as write the tests for), since in the time between end of read and beginning of write, context-switch could occur and some other request could change the file. Although I say `async`, all the sequential flow is respected

nicola:
I am adding new tests for ACL

nicola:
I solved the issues pointed out by @deiu in my last commit (mainly by implementing `ldp.stat` and by using `ldp.stat` and `ldp.writeFile` instead of `fs`), in this way the status code is set by this function

PR #100, opened 2015-07-19
Added meta and acl links and basic tests.

martinmr:
Added code to add acl and meta links. Since it uses the options object, @nicola  would need to refactor it before merging #97 to avoid errors. 
I added some tests.


nicola:
It is good for me, this can be merged, no worries about my refactoring, I will resolve conflict in case

PR #107, opened 2015-07-30
Added proper content negotiation. Several smaller fixes.

deiu:
There are several fixes here, but the most important is the content negotiation part. It now properly parses the Accept header and sets the right content type in the response.

Also, this PR adds warp as default viewer (through a 303 redirect) for browsing resources.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -116,7 +116,7 @@ function get(req, res, includeBody) {
         }
 
         // redirect to file browser if we got text/html with highest priority
-        if (res.acceptTypes.indexOf('text/html') === 0) {
+        if (container && res.acceptTypes.indexOf('text/html') === 0) {
             return res.redirect(303, browseSkin + req.protocol + '/' + req.get('host') + req.originalUrl);
         }
 
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -20,6 +20,9 @@ var subscription = require('../subscription.js');
 var ldpVocab = require('../vocab/ldp.js');
 var turtleExtension = '.ttl';
 
+// this should be moved to options
+var browseSkin = 'https://linkeddata.github.io/warp/#/list/';
+
 function get(req, res, includeBody) {
     var ldp = req.app.locals.ldp;
     var uri = utils.uriBase(req);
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ldnode",
   "description": "Linked Data Platform",
-  "version": "0.2.8",
+  "version": "0.2.10",
```
nicola: just leave 0.2.8, I will update it through Github tags and npm releases

test/http.js
```diff
@@ -10,7 +10,7 @@ describe('HTTP APIs', function() {
 
   var emptyResponse = function(res) {
       if (res.text.length !== 0) {
-          error("Not empty response");
+          console.log("Not empty response");
```
nicola: I have a feeling this should be `throw new Error`, I will need to look into this

lib/handlers/get.js
```diff
@@ -109,6 +115,11 @@ function get(req, res, includeBody) {
             }
         }
 
+        // redirect to file browser if we got text/html with highest priority
+        if (res.acceptTypes.indexOf('text/html') === 0) {
+            return res.redirect(303, browseSkin + req.protocol + '/' + req.get('host') + req.path);
```
nicola: This won't work when ldnode is mounted on a route.
this should be instead:

```
res.redirect(303, browseSkin + req.protocol + '/' + req.get('host') + req.originalUrl);
```

`req.path` to `req.originalUrl`

nicola: actually, use
```
var utils  = require('../utils');
utils.uriAbs(req) + req.originalUrl
```

lib/header.js
```diff
@@ -74,25 +75,38 @@ function parseMetadataFromHeader(linkHeader) {
...
+    var negotiator = new Negotiator(req);
+    return negotiator.mediaTypes();
+}
+
+function parseAcceptRDFHeader(req) {
     var acceptFinalValue;
-    var acceptHeader = req.get('Accept');
-    if (acceptHeader === undefined) {
+    var acceptHeader = negotiateContentType(req);
+    if (acceptHeader === undefined || acceptHeader.length === 0) {
```
nicola: `if (acceptHeader)`

so, `undefined`, `false`, `''` will be false

PR #111, opened 2015-08-03
Making proxy work

nicola:
- [x] Proxy working
- [x] basic test
- [x] further tests
- [x] make sure cors works on proxy requests

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -54,7 +54,6 @@ function LDP(argv) {
 
   ldp.pathFilter = regexp().start(ldp.mount).toRegExp();
   ldp.proxy = argv.proxy;
-  ldp.proxyFilter = regexp().start(ldp.proxy).toRegExp();
 
   // Cache of usedURIs
   ldp.usedURIs = {};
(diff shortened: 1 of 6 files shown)


nicola:
Interesting travis might have some proxy for external resources, this needs further testing

lib/ldp.js
```diff
@@ -53,8 +53,8 @@ function LDP(argv) {
   }
 
   ldp.pathFilter = regexp().start(ldp.mount).toRegExp();
-  ldp.xssProxy = argv.xssProxy;
-  ldp.proxyFilter = regexp().start(ldp.xssProxy).toRegExp();
+  ldp.proxy = argv.proxy;
+  ldp.proxyFilter = regexp().start(ldp.proxy).toRegExp();
```
deiu: Quick question: what does proxyFilter do, and why do we need a filter for proxy?

deiu:
r+

lib/ldp.js
nicola: Good question, removing that variable. It created a regex based on `ldp.proxy`, however express handles paths as regexp already, so no need for that

PR #113, opened 2015-08-05
Set acl and describedBy (meta) Link headers for all requests.

### LINKED ISSUE #112: URIs in link: headers should be relative where possible
Link:		<http://www.w3.org/ns/ldp#Resource>; rel="type", </timbl/2015/08/02/poll/forms.ttl.changes>; rel="changes", <https://timbl.com/timbl/2015/08/02/poll/forms.ttl.acl>; rel="acl", <https://timbl.com/timbl/2015/08/02/poll/forms.ttl.meta>; rel="describedBy"

should be

      Link:		<http://www.w3.org/ns/ldp#Resource>; rel="type", <forms.ttl.changes>; rel="changes", <forms.ttl.acl>; rel="acl", <forms.ttl.meta>; rel="describedBy"

deiu:
ldnode was previously only setting the acl and describedBy (meta) Link headers only for GET. The PR sets the Link headers for all request methods. 

It also fixes #112

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -42,8 +42,8 @@ function uriBase(req) {
     return uriAbs(req) + (req.baseUrl || '');
 }
 
-function uriRelative(uri, base) {
-    var relative = path.relative(base, uri);
+function uriRelative(uri) {
+    var relative = path.basename(uri);
     return relative;
 }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -42,6 +42,11 @@ function uriBase(req) {
     return uriAbs(req) + (req.baseUrl || '');
 }
 
+function uriRelative(uri, base) {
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -42,6 +42,11 @@ function uriBase(req) {
     return uriAbs(req) + (req.baseUrl || '');
 }
 
+function uriRelative(uri) {
+    var relative = path.basename(uri);
```
nicola: I would not rely too much on this function. This function just gives the latest part in a path `hello/nicola/folder` will return `folder`. The idea of relative is that a link is relative to a particular directory. Although this solves the problem for the specific case, this function needs a second parameter that is the base from which we want to find the path of `uri` relative to the `base`

Issue #110, opened 2015-08-03
Custom error pages, especially 401


timbl:
I need to be able to set up a custom return HTML page for HTTP errors, especially error 401 "Unauthenticated".    In practice this is an important part of the on-boarding process, to lead someone who does not have a webid to grab one.

Apache does this with the ErrorDocument directive a la 
 ErrorDocument 404 /Lame_excuses/not_found.html
 ErrorDocument 401 /Subscription/how_to_subscribe.html

See https://httpd.apache.org/docs/2.2/custom-error.html

timbl:
(Note Apache in fact does replacement of many different variables within the page too.  This is less essential but obviously nice)

martinmr:
I found something relevant: http://www.hacksparrow.com/express-js-custom-error-pages-404-and-500.html

However, I think this method only works with 404 and 500 status and it won't work if a previous handler sends the response before these handlers are reached, which all handlers do now.

One idea is to have the handlers set only the status of the response without sending it (unless the status code is 200 or 201 in which case is safe to immediately send the response body), and instead calling next, which looks at the error code and returns the appropriate error page. I'll look into this to see if it's possible.

gobengo:
@martinmr Another good option instead: of just setting the status code to 400/500, pass an error object to next(). Later, you can have an error middleware that knows about the desired way of returning that error as an HTML page (if the client even wants HTML).

See https://github.com/shutterstock/node-common-errors#httpstatus

nicola:
In node you can intercept errors and handle the error page:

```
var ldnode = require('ldnode');
var ldpServer = ldnode();
ldpServer.use(function(err, req, res, next) {
  if(err.status !== 404) {
    return next();
  }
  res.render('Special 404 page.html')
});
ldnode.listen(3456, function () {
  console.log('Server started')
});
```

nicola:
Perfect, I have done some little thinking,
in order to respect @timbl's original idea of ldnode to be _as easy as possible_ I will add `--page-404` and `--page-500` as a command line params so that you can pass the page without coding

martinmr:
I was thinking instead there could be a special folder inside the ldnode
root where ldnode looks for error pages (named for example 404.html). That
way there's no need to create an option for every error. We could have the
folder be a default or a command line option.

In node you can intercept errors and handle the error page:

var ldnode = require('ldnode');
var ldpServer = ldnode();
ldpServer.use(function(err, req, res, next) {
  if(err.status !== 404) {
[...]

PR #114, opened 2015-08-06
Error pages

martinmr:
Custom error handling. Needs automated tests but I've tested on my machine and it works. If anyone spots a place where the error still needs to be forwarded to the error handler (instead of calling res,send()) please let me know.

Also updated rdflib to version 0.2.8. It seems that the Andrei made the change but somehow it got reversed and the json-ld tests were failing.


lib/error.js
```diff
@@ -0,0 +1,29 @@
+/*jslint node: true*/
```
deiu: Shouldn't this file be in lib/handlers/?

bin/ldnode.js
```diff
@@ -79,6 +79,14 @@ var argv = require('nomnom')
     help: 'Suffix for SSE files (default: \'.events\')',
     abbr: 'sE'
   })
+  .option('noErrorPages', {
+    full: 'no-error-pages',
+    help: 'Disable custom error pages (use Node.js default pages instead)'
```
nicola: This should be a flag,

flag: true

lib/acl.js
```diff
@@ -510,7 +509,11 @@ function allow(mode, req, res, next) {
     var reqPath = res && res.locals && res.locals.path ? res.locals.path : req.path;
     var acl = reqToACL(req);
     acl.allow(mode, reqPath, function(err) {
-        next(err);
+        if (err) {
```
nicola: Just write

next(err)

lib/acl.js
```diff
@@ -540,7 +543,11 @@ exports.allowAppendThenWriteHandler = function(req, res, next) {
         }
         // Append failed, maybe user can write
         allow("Write", req, res, function(err) {
-            next(err);
+            if (err) {
+                return next(err);
```
nicola: Same here

lib/error.js
nicola: Ths should be in handlers/ not in lib/

index.js
```diff
@@ -70,6 +73,11 @@ function createServer(argv) {
     var ldp = ldpApp.locals.ldp;
     app.use(ldp.mount, ldpApp);
 
+    //Error handling
+    app.use(errorHandler.handler);
```
nicola: Shouldn't this be mounted as a route?
Mounting this here, you could miss LinkHandler and Cors.

In fact ldpApp.locals.ldp already has the lpd instance

lib/handlers/delete.js
```diff
@@ -16,9 +16,7 @@ function handler(req, res) {
     ldp.delete(filename, function(err) {
         if (err) {
             debug("DELETE -- error: " + err);
-            return res
-                .status(err.status)
-                .send(err.message);
+            return next(err);
```
nicola: This comment is valid for all of your other changes,
Are you sure this is necessary?

It may be I am super wrong, but you could just intercept every time a red.send(404) or similar happens like in my example

test/resources/errorPages/404.html
```diff
@@ -0,0 +1,5 @@
+<html>
+<body>
+    404 Error Page
+</body>
```
nicola: Can you add all the other error pages?

nicola:
It looks good,
Merge only after tests and little changes

lib/error.js
martinmr: That's right. I'll move the file.

lib/acl.js
martinmr: Are you sure? The documentation said if anything is passed to next it will forward it to the error handler. I was not sure if a null value counts as a valid value so I made the check explicit just to make sure. But if you are sure the check is already being done by express itself I'll change it.

lib/handlers/delete.js
martinmr: I am sure. I tried your example first but the requests are not being intercepted. The request is immediately sent. You need to call next(err) for the error handler to be used.

test/resources/errorPages/404.html
martinmr: Yes. I will.

index.js
martinmr: I tried to mount it to the router first but it didn't work. I think it was because the handlers were calling res.sent instead of next(err) and the responses were being intercepted. I will attach it to the router again to check if it works now.

martinmr: Yeah, that was the issue. I got the error handler to work just on the router and not in the whole app.

deiu:
I suggest we wait to merge it until we've finished milestone 2.0 (tonight or tomorrow).

martinmr:
@deiu That's fine with me.

lib/acl.js
nicola: I will have a look at this before the meeting, I could be wrong, but I have been reasoning about this, and I think yours is better design, since we always pass a next.

nicola: Remove my comment you can just write next(err) if err is undefined, it passes to the next handler, no issues

lib/handlers/delete.js
nicola: I will have a look at this before the meeting, I could be wrong, but I have been reasoning about this, and I think yours is better design, since we always pass a next.

lib/handlers/post.js
```diff
@@ -69,7 +72,10 @@ function handler(req, res) {
     if (resourcePath === null) {
         ldp.releaseResourceUri(resourcePath);
         debug("POST -- URI already exists or in use");
-        return res.sendStatus(400);
+        var resourceErr = new Error();
+        resourceErr.status = 400;
+        resourceErr.message = "URI alredy exists or in use";
```
nicola: `already`

test/errors.js
```diff
@@ -0,0 +1,64 @@
...
+    noErrorLdp.listen(3458);
+
+    var noErrorServer = supertest(errorAddress);
+
+
+    describe('Error page test', function() {
+        it('Should return 404 custom page if flag set to true',
+           function(done) {
+               errorServer.get('/non-existent-file.html')
+                   .expect(/404 Error Page/)
```
nicola: can you read the file (you can use: var read = require('test-utils').read) and checking against it (in the future we might change that file)

nicola:
There is a scenario not described in tests that I would like to ensure you handle in this way.

Imagine I pass this command:

`ldnode --error-pages ./thisFolder`

Now, if that folder is empty, what error pages are you going to show default express ones or the default html one? I think the second one is the answer, unless I explicitly specify --no-(default)-error-pages

martinmr:
There are no default html error pages. It Will just print the error
message. What exists is a default folder on which to look for error
messages.

nicola:
Alright, so if I am doing a request that causes a 500, and I use:

```
ldnode --error-pages ./thisFolder
```

and the folder is empty, it will still return the default html 500 error page, correct?

nicola:
Perfect

nicola:
Mmm. Hold on, I don't want a default folder to exist, this is the problem addressed in the issue #110

martinmr:
I saw that. However, the default location of this folder is inside the
server's root, not inside ldnode, so no files need to be included with the
npm package. Basically if no custom error page folder is given, ldnode will
look at /error pages and serve the custom page if it exists. Otherwise it
send the default node error (just the message). It just saves the user from
having to type the error folder unless they want it on a different location
(outside the root for example).

nicola:
This is an implicit behavior :( I am not sure I am comfortable with that.
I was more comfortable to have some nice pages as default (existing in the repo)

martinmr:
Sure, but I don't see how

```
ldp.errorPages = argv.errorPages || ldp.root + 'errorPages/'
```
is different from doing
```
  ldp.mount = argv.mount || '/';
```
Both are implicit. I don't think there's anything necessarily wrong with that as long as the default values are documented. I think it's kind of weird to have the error handler read from files inside ldnode's source tree. Plus having the default error folder saves the user from having to explicitly set the command line option.

PR #115, opened 2015-08-06
Improved the function that makes Link URIs relative

deiu:
This patch fixes the relative URIs in Link headers. I've added tests to make sure this happens according to the spec. For example:

* `HEAD https://example.org/foo/bar` will return a corresponding ACL link `bar.acl`, then
* `HEAD https://example.org/foo/` will return a corresponding ACL link `.acl`, and finally:
* `HEAD https://example.org/` will return a corresponding ACL link `.acl`

The `basename()` function does not require a `base` param. It only returns two possible outcomes: either the path is for a file (in which case it returns just the filename), or it returns an empty string if path ends with a `/`.

### DIFF
--- a/lib/header.js
+++ b/lib/header.js
@@ -49,8 +49,8 @@ function linksHandler(req, res, next) {
     }
 
     // Add ACL and Meta Link in header
-    addLink(res, utils.basename(req.path) + ldp.suffixAcl, 'acl');
-    addLink(res, utils.basename(req.path) + ldp.suffixMeta, 'describedBy');
+    addLink(res, utils.basePathName(req.path) + ldp.suffixAcl, 'acl');
+    addLink(res, utils.basePathName(req.path) + ldp.suffixMeta, 'describedBy');
     // Add other Link headers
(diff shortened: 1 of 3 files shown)


nicola:
Good for me

lib/header.js
```diff
@@ -49,20 +49,9 @@ function linksHandler(req, res, next) {
...
-    var metaLink = utils.getResourceLink(
-        filename, uri,
-        ldp.root, ldp.suffixMeta,
-        ldp.suffixAcl);
-
-    addLink(res, utils.uriRelative(aclLink), 'acl');
-    addLink(res, utils.uriRelative(metaLink), 'describedBy');
-
-
+    addLink(res, utils.basename(req.url) + ldp.suffixAcl, 'acl');
```
nicola: Should we deprecate getResourceLink then?

deiu: I'm inclined to deprecate it, but let's give it a bit of time to see if we're not missing anything.

deiu:
Hold on. I found a bug.

PR #119, opened 2015-08-07
avoiding starting server in tests

nicola:
- [x] http
- [x] params
- [x] format
- [x] acl (however for simplicity we should leave this)


nicola:
Fix https://github.com/linkeddata/ldnode/issues/99

PR #124, opened 2015-08-08
Use the real error messages from node-webid

deiu:
Half the error codes in the switch() did not even exist anymore and/or were very cryptic. The new version of node-webid sends more verbose error codes, and we can just forward those directly to the debugger/printer.


nicola:
Good, will merge once I release the new `node-webid`

PR #127, opened 2015-08-18
fixing issues with errors and enabling custom error handlers

nicola:
- Introducing a custom error `HttpError` (`./lib/http-error.js`)
- Making every error an instance of `HttpError`
- Adding `ldp.errorHandler` in the options to enable custom error handlers
- Fixing little issues
- updated README
- added a custom error example
- now errors are not shown as `[object Object]` since instead of passing the error Object as a message, we pass the error String


deiu:
r+

PR #128, opened 2015-08-18
browseSkin is now --skin

nicola:
You can now set what skin to use with `--skin`

- `--skin` if not set, `https://linkeddata.github..` is used
- `--no-skin` to avoid using skins

### DIFF
--- a/bin/ldnode.js
+++ b/bin/ldnode.js
@@ -93,6 +93,9 @@ var argv = require('nomnom')
     full: 'error-pages',
     help: 'Folder from which to look for custom error pages files (files must be named <error-code>.html -- eg. 500.html)'
   })
+  .option('skin', {
+    help: 'Path to a skin to load (default: https://linkeddata.github.io/warp/#/list/)'
+  })
   .parse();
 
 // Print version and leave
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -75,6 +75,12 @@ function LDP(argv) {
   }
   ldp.errorHandler = argv.errorHandler;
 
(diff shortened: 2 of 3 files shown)


bin/ldnode.js
```diff
@@ -93,6 +93,9 @@ var argv = require('nomnom')
     full: 'error-pages',
     help: 'Folder from which to look for custom error pages files (files must be named <error-code>.html -- eg. 500.html)'
   })
+  .option('skin', {
+    help: 'Path to a skin to load (default: https://linkeddata.github.io/warp/#/list/)'
```
deiu: I think `path` is misleading. Maybe use `URI` instead?

deiu:
I think having just one param is enough (i.e. --skin), since the other one (--no-skin) is implied by default.

nicola:
Nomnom, the library that we are using for the command line has default --no- for every parameter, for example --no-webid.

However in this case I put the `linkeddata.github` a default skin, so in that case you won't do `--skin ...`, so to remove it you must do `--no-skin`

lib/ldp.js
```diff
@@ -75,6 +75,12 @@ function LDP(argv) {
   }
   ldp.errorHandler = argv.errorHandler;
 
+  // argv.skin is false when --no-skin is passed
+  if (argv.skin !== false) {
+    ldp.skin = argv.skin || 'https://linkeddata.github.io/warp/#/list/';
+  }
+  console.log(argv)
```
nicola: ops, I am gonna remove this now

PR #133, opened 2015-08-31
Fixed wrong CType returned for the root acl resource.

deiu:
This PR introduces the following fixes:
 * register acl and meta suffixes as Turtle extensions
 * adds missing test to check content-type header for a root acl resource
 * adds helper function to push items to an array if they don't exist
 * adds helper function to check if a given path contains an extension (out of a list of extensions)


nicola:
See https://github.com/linkeddata/ldnode/pull/134

PR #136, opened 2015-09-18
Create-server and create-app

nicola:
- I brought the code written in the index.js in create-server and creat-app
- I avoid passing key/cert/mount to createApp since it doesnt handle with that, while only createServer does
- I mount ldp on the routes as a middleware
- renamed debug file from logging to debug
- put proxy in its own file (./lib/handler)

tl;dr
no changes in code, just moved code around
this make it easier to change LdpMiddleware in case someone will develop a different one than ours that operates on files


test/params.js
```diff
@@ -89,44 +89,44 @@ describe('LDNODE params', function () {
     });
   });
```
deiu: Why have these been commented out?

nicola: since it can't be tested in that way.
In other words, now mount is a parameter of createServer, while this is a test for createApp, so mount can't be specified

PR #144, opened 2015-09-28
Adding web socket support

nicola:
It uses [solid-ws](http://github.com/nicola/solid-ws)

- [x] I cleaned some code
- [x] Removed the difference between GET/HEAD since HEAD fallback on GET by default
- [x] Support on websocket is self-contained in the library solid-ws
- [x] Added `next` to every res.send


nicola:
I can't replicate the rdflib bug I had a week ago.

nicola:
Blocked until we publish rdflib on npm, ready otherwise

PR #145, opened 2015-10-06
Dropped posix:Directory and posix:File in favor of ldp types

deiu:


nicola:
Both files and dirs should be resources

lib/ldp.js
```diff
@@ -292,7 +287,7 @@ LDP.prototype.listContainer = function (filename, uri, containerData, callback)
       ldp.stat(filename, function(err,  containerStats) {
         if (!err) {
           addStats(resourceGraph, '', containerStats);
-          resourceGraph.add(
+            resourceGraph.add(
```
nicola: why this space?

deiu: Hmm...weird. I think it looked out of place. My bad.

PR #147, opened 2015-10-08
New ACL

nicola:
Work in progress


nicola:
It finally passes all the tests, I am currently skipping tests for Groups (I have commented out Group implementation), for Owner-only (since we are not using it yet - at least not currently implemented) and for Glob.

I also commented the clean up since I changed the files I used for testing, so it became obsolete

lib/acl.js
```diff
@@ -14,451 +14,304 @@ var utils = require('./utils.js');
...
-    this.ldp = opts.ldp;
-    this.origin = opts.origin || '';
+  var self = this
+  opts = opts || {}
+  if (opts.store && opts.store.graph && !opts.fetch) {
+    self.fetch = opts.store.graph.bind(opts.store)
+  }
+  self.fetch = self.fetch || opts.fetch
+  self.match = opts.match || match
+  self.suffix = opts.suffix || '.acl'
```
deiu: Maybe we can replace the hardcoded value with the config/param one.

nicola: that is `opts.suffix`, you can pass the suffix

PR #164, opened 2015-12-24
Adding Identity Provider

nicola:
Fixes https://github.com/linkeddata/ldnode/issues/105

Ready to merge (only after https://github.com/linkeddata/node-webid/pull/9 is merged and published on NPM)

This has been a lot more fun (and a lot more work!) than we thought.

Things left to do:
- [x] certificate creation (@sible777 helped)
- [x] availability of subdomain storage (right now the "api" to create an account work, but one can't really access it yet - I have to enable ldp on subdomains!)
- [x] support for multiple hosts
- [x] make it work with solid-signup
- [x] creating an account and use the same session to request the certificate
- [x] fix the test that currently has a TLS error (maybe because the certificate is for one domain only)

Things left to do in new PRs:
-  creation of workspace, preference files, inbox.. (this should be just piping) 
-  certificate lost / stolen, how to get a new one? https://github.com/linkeddata/ldnode/issues/175


nicola:
This has been a lot mor fun (and a lot more work!) than we thought.
Things left to do:

- [ ] certificate creation (40%)
- [ ] availability of subdomain storage (right now the "api" to create an account work, but one can't really access it yet) (50%)
- [ ] creation of workspace, preference files, inbox.. (this should be just piping)

nicola:
@sible777 ^ :)

nicola:
There are a few things I need to understand better, but I will look at gold for now.

- [ ] creating an account and use the same session to request the certificate
- [ ] fix the test that currently has a TLS error (maybe because the certificate is for one domain only)

sible777:
>creating an account and use the same session to request the certificate @nicola 

This is something that I too would like to understand better! I tried implementing some stuff like this recently and the only way I could figure to do it was to have the certificate request come from a different domain/provider. 

Let me know what you figure out :+1:

nicola:
Ok, when tested with https://github.com/linkeddata/node-webid/pull/9 it works. However, we need to merge https://github.com/linkeddata/node-webid/pull/9 and update the version of `webid` in the `package.json` before merge

nicola:
Test do not pass because of VHOST, any idea on how to test that on travis @deiu ?

test/identity-provider.js
```diff
@@ -0,0 +1,152 @@
...
+        .end(function (err) {
+          if (err) return done(err)
+
+          var spkac = ''
+          subdomain.post('/accounts/cert')
+            .send('webid=https://nicola.' + host + '/profile/card#me&spkac=' + spkac)
+            .expect(500, done)
+        })
+    })
+  })
```
deiu: I wonder if we can add a test here to check if the user can login with the new cert -- i.e. get the same WebID from the User header, following an authenticated request.

PR #168, opened 2016-01-04
Updates the list of supported HTTP methods

csarven:
.. in the README


melvincarvalho:
Shouldnt HEAD be supported acc. to RFC 2616?

>   the methods GET and HEAD MUST be supported by all general-purpose servers. 
     All other methods are OPTIONAL

http://www.w3.org/Protocols/rfc2616/rfc2616.txt

melvincarvalho:
Ah, I see it's added now.  Great!

PR #170, opened 2016-01-05
refactoring get to fix #138

### LINKED ISSUE #138: Avoid translation when possible

nicola:
- Using streams when serving files
- Translate only when possible
- Fix mess with negotiation
- Added some tests

also, we got for free:
- Possibility to translate containers

--
It was a consistent amount of piping and cleaning, now it is very simple, the workflow is the following:

1. User asks for type x, y, z of FILE
2. We find the file and check if the mimetype matches x, y or z and uses the one which does not need any translation, otherwise translate
3. If translation is not possible 406 is given,

my key change is on line 55-97, the rest is to support streams

### DIFF
--- a/test/formats.js
+++ b/test/formats.js
@@ -25,15 +25,28 @@ describe('formats', function () {
         var isValidJSON = function(res) {
             var json = JSON.parse(res.text);
         };
-        it('Should return JSON-LD document if Accept is set to application/ld+json', function(done) {
+        it('Should return JSON-LD document if Accept is set to only application/ld+json', function(done) {
+            server.get('/patch-5-initial.ttl')
+                .set('accept', 'application/ld+json')
+                .expect(200)
+                .expect('content-type', /application\/ld\+json/)
+                .end(done)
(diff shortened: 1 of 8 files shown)


test/formats.js
```diff
@@ -25,15 +25,28 @@ describe('formats', function () {
...
+        });
+        it('Should return JSON-LD document if Accept is set to application/ld+json and other types', function(done) {
+            server.get('/patch-5-initial.ttl')
+                .set('accept', 'application/ld+json;q=0.9,application/rdf+xml;q=0.7')
                 .expect('content-type', /application\/ld\+json/)
                 .expect(200, done);
         });
         it('Should return valid JSON if Accept is set to JSON-LD', function(done) {
             server.get('/patch-5-initial.ttl')
-                .set('accept', 'application/ld+json;q=0.9,text/turtle;q=0.8,text/plain;q=0.7,*/*;q=0.5')
```
deiu: Why were the other values removed?

Issue #169, opened 2016-01-04
GET method needs refactoring


nicola:
The get method should be implemented in the same way as the other methods.

it should just contain `ldp.get(path, callback)`, and this should handle parsing or not, it is a bit of spaghetti right now

This effort should fix:
- https://github.com/linkeddata/ldnode/issues/140
- https://github.com/linkeddata/ldnode/issues/138

And help:
- https://github.com/linkeddata/ldnode/issues/105 to support subdomain in https://github.com/linkeddata/ldnode/pull/164

nicola:
done in https://github.com/linkeddata/ldnode/commit/3bab7844c741ff4f434b916e7cdf38f0bf0e8355

PR #171, opened 2016-01-05
secure cookies on https instances, fix #160

### LINKED ISSUE #160: Make sure cookies are set to be `Secure` when on https

nicola:


nicola:
Also fixes #169

PR #174, opened 2016-01-07
Updating documentation and simplifying the command-line example

nicola:
This may help towards https://github.com/linkeddata/ldnode/issues/173

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,19 +4,18 @@
 [![NPM Version](https://img.shields.io/npm/v/ldnode.svg?style=flat)](https://npm.im/ldnode)
 [![Gitter chat](https://img.shields.io/badge/gitter-join%20chat%20%E2%86%92-brightgreen.svg?style=flat)](http://gitter.im/linkeddata/ldnode)
 
-Linked Data Platform server based on [rdflib.js](https://github.com/linkeddata/rdflib.js) and [node.js](https://nodejs.org/). This is all you need to run distributed linked data apps on top of the file system.
+Ldnode implements the [Linked Data Platform](http://www.w3.org/TR/ldp/) and [Solid](https://github.com/solid) in [NodeJS](https://nodejs.org/). This is all you need to run distributed Linked Data apps on top of the file system.
 
-You can run ldnode as a [command-line tool](https://github.com/linkeddata/ldnode/blob/master/README.md#command-line-tool) or as a [library](https://github.com/linkeddata/ldnode/blob/master/README.md#library) for your Nodejs/Express app.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,30 +28,40 @@ The command line tool has the following options
...
+   -s, --secret                HTTP Session secret key (e.g. "your secret phrase")
+   -fU, --force-user           Force a WebID to always be logged in (usefull when offline)
+   -P, --proxy                 Use a proxy on example.tld/proxyPath
+   --no-live                   Disable live support through WebSockets
+   -sA, --suffix-acl           Suffix for acl files (default: '.acl')
+   -sM, --suffix-meta          Suffix for metadata files (default: '.meta')
+   -sE, --suffix-sse           Suffix for SSE files (default: '.events')
+   --no-error-pages            Disable custom error pages (use Node.js default pages instead)
+   --error-pages               Folder from which to look for custom error pages files (files must be named <error-code>.html -- eg. 500.html)
+   --skin                      URI to a skin to load (default: https://linkeddata.github.io/warp/#/list/)
```
dmitrizagidulin: What does skin refer to?

nicola: good point.. Tabulator or Warp are skins.. they are like default viewer. It is like a default app that you want to use to browse the data, you can also pass `--no-skin` to jave none

dmitrizagidulin: Makes sense. :) Would you be opposed to renaming them to apps, instead of skins?

sandhawke: Skin is a confusing word, but it's important to capture the idea that sometimes an app is invoked as a viewer for particular resource, as opposed to its usual mode.   "Viewer" isn't right, though, because it doesn't have to be read-only.    It's like how operating systems pick some app to run when you indicate you want to open a file (e.g.by double clicking).   What do you call that app?   The "default opener" or something.

csarven: re: "Default application" for file types. Tabulator/Warp are part of a class of "file browser" or "file manager" applications. If for instance, the default application to open is a "photo manager", it will act like a personal photo management software, displaying and offering special controls to only the file types in knows what to do with. I think renaming to something like `--default-app` seems intuitive enough and the terminology is closer to what's used in OSs, and with no special binding to a particular class of applications. `skin` on the other hand comes across like a theme, decoration, or appearance.

dmitrizagidulin:
Aside from renaming `skin` to `default-app` (which should be its own issue), +1 on this PR, looks good to me. :)

README.md
dmitrizagidulin: Moving the 'skin'/'default app' discussion to https://github.com/solid/solid-spec/issues/65

dmitrizagidulin:
Merging.

PR #183, opened 2016-01-17
Refactoring POST and code style

nicola:
- removing events SSE
- use `extend` on Ldp object instead of manually listing properties
- rewrite of `post` handler to be: (1) find a free valid path (2) put the request body either in the  `newPath.meta` or the in the `newPath` 
- removing all redundant code for file/container creation
- removed usedURI global variable
- implemented `error` instead of `HttpError` for nice and concise error handling
- forcing one code style

See https://github.com/nicola/ldnode/commit/eea1e436f0b7b0cf810de5203ec5f0079d319abc for changes, since the net changes are too noisy, with tabs and semicolons


lib/handlers/post.js
```diff
@@ -1,153 +1,52 @@
...
+      contentType === 'application/sparql-update') {
+    debug('POST -- Handling sparql query via PATCH')
+    return patch.handler(req, res, next)
+  }
+
+  var containerPath = req.path
+  debug('POST -- On parent: ' + containerPath)
+
+  // Not a container
+  if (containerPath[containerPath.length - 1] !== '/') {
```
deiu: Do we currently have any other behavior for POST, except for creating new resources? In this case, shouldn't we just "correct" the URI by appending the missing `/`?

nicola: Just added a patch for this with 2 unit tests

deiu:
r+

dmitrizagidulin:
What's the coding style? (To which it was standardized?)

nicola:
@dmitrizagidulin right now to github.com/feross/standard.

Although it takes a command in terminal to convert into any other (maybe https://github.com/Flet/semistandard).

At least we have it in all in one style for future changes :)

sandhawke:
Does all of ldnode now follow standard?

nicola:
@sandhawke, yes (all the code, tests and examples)

PR #185, opened 2016-01-18
Removing function isMetadataFile

nicola:
I am currently removing unused functions from ldp.
This function is only called once and it does some trivial string checking,
there is no need for a function


dmitrizagidulin:
@nicola - are you sure about this? The function was way more readable... :)

PR #187, opened 2016-01-18
WIP: Exporting handlers as module

nicola:
- exporting handlers as modules
- use one-liners for routing


nicola:
Do we really need to send the response time back? @deiu ?

deiu:
I don't think so. In fact, I don't know what that did.

nicola:
@deiu this PR was WIP (work in progress), we need one more merge

PR #196, opened 2016-01-20
avoid utf-8 encoding by default

nicola:


lib/ldp.js
```diff
@@ -76,7 +76,7 @@ LDP.prototype.stat = function (file, callback) {
 }
 
 LDP.prototype.createReadStream = function (filename) {
-  return fs.createReadStream(filename, { 'encoding': 'utf8' })
```
dmitrizagidulin: So, what encoding will actually be used? None? Or whatever the client requests?

deiu: No ecoding is needed for serving piping the stream.

PR #204, opened 2016-01-22
Ftu signup

nicola:


nicola:
I am now resolving the conflict

nicola:
@deiu use #205

Issue #201, opened 2016-01-21
Implement launching LDNode is Single-User mode


dmitrizagidulin:
(Currently being implemented on branch https://github.com/linkeddata/ldnode/tree/ftu-signup , originally PR #182)

LDNode currently works in a multi-user mode (such as being run by a Data Storage Provider like Databox.me), where multiple users can create accounts, and they each get their own subdomain.

The idea behind this issue is: LDNode also needs to support a Single-User use case. This includes:

* Server owner specifies `--single-user-mode` on the command line, when launching, which does the following:
* Serves up a one-time Install app, where the user can go to generate a new account / WebID / certificates, which will then be written in the config file of the server. (and the Install app is disabled)
* Disables new account creation and deletion API endpoints.
* Support running in a sub-directory (so that a user can run an LDNode Solid instance from `usershomepage.com/solid/` instead of root. (This sounds like it's already implemented).

nicola:
done in https://github.com/linkeddata/ldnode/issues/205

PR #205, opened 2016-01-22
Ftu signup clean

nicola:


bin/ldnode.js
```diff
@@ -97,6 +97,11 @@ var argv = require('nomnom')
     full: 'default-app',
     help: 'URI to use as a default app for resources (default: https://linkeddata.github.io/warp/#/list/)'
   })
+  .option('signup', {
```
dmitrizagidulin: Can we rename this option to `singleUserMode` (from our in-person discussion / #201?)

nicola: This is actually just to sign up the first user.
Single user mode right now is done by not setting the flag --idp

This is like --create-admin



On Fri, Jan 22, 2016 at 6:00 AM -0800, "Dmitri Zagidulin" <notifications@github.com> wrote:












In bin/ldnode.js:
> @@ -97,6 +97,11 @@ var argv = require('nomnom')
>      full: 'default-app',
>      help: 'URI to use as a default app for resources (default: https://linkeddata.github.io/warp/#/list/)'
>    })
> +  .option('signup', {



Can we rename this option to singleUserMode (from our in-person discussion / #201?)

dmitrizagidulin: Makes sense. (So this is more for https://github.com/solid/solid/issues/7). 
Can we rename it to `--create-admin` then?

nicola: done

PR #210, opened 2016-01-23
Use ./index.html when found                                                                                                                           

nicola:


deiu:
Is this ready for merge?

nicola:
we should do that if there is no sign of work in progress, things can be merged

deiu:
OK, I got confused by the presence of the `In progress` label.

PR #211, opened 2016-01-23
Fixed test for index.html

deiu:
The previous test was not complete, which is why it passed even though the actual implementation was broken.


deiu:
Ready for merge.

nicola:
well spot! thanks!

PR #212, opened 2016-01-23
WIP - fix turtle dir listing

deiu:
Currently, dir listing does not use the preferred media type.


deiu:
It seems that `negotiator.mediaType()` doesn't return the media type with the highest `q` value. ( cf. https://github.com/linkeddata/ldnode/blob/master/lib/handlers/get.js#L33)

deiu:
It turns out rdflib.js omits the `q` value for `application/rdf+xml` in the Accept header.

PR #215, opened 2016-01-25
Fix delete on containers

deiu:
Previously, the delete operation did not remove the corresponding meta and acl files. It also failed when trying to delete an empty container.


deiu:
Ready for merge.

sandhawke:
Considering opening an issue on whether ACLs should survive delete.   It's unclear.   There are use cases on either side of that.    On w3.org they do.

deiu:
This is the ACL document which lives within the container. If the container is deleted, it would be impossible to keep the ACL document.

dmitrizagidulin:
Fixing tests, stand by

dmitrizagidulin:
:+

PR #220, opened 2016-01-27
Fixes #219; serve requested resource if missing mime type

### LINKED ISSUE #219: Not redirecting to `defaultApp` when browsing with `Accept: text/html`
To reproduce:

1. (Optional, pretty sure this will work with existing LDNode resources) Start up LDNode, create a new account using `--create-admin`.
2. Try to view a container (like `/`) or resource (like `/profile/card`) in your browser (after authenticating).

**Expect:** A redirect to the `--default-app` ([Warp](https://linkeddata.github.io/warp/#/list/) by default).

**Actual:** Get a 404 Not found (even on a public resource like the profile card).

Note: Passing in `Accept: text/turtle` (say, via `curl`) works as expected.

deiu:


deiu:
Ready for merge.

dmitrizagidulin:
Tested it out, fixes the issue. :+1: from me.

Issue #223, opened 2016-01-28
Server crashes when requesting a container as `application/ld+json`


dmitrizagidulin:
If you do a GET request to a Solid container, with any of the following headers, it crashes `ldnode`.

* `Accept: application/ld+json`
* `Accept: application/nquads`
* `Accept: application/n-quads`

This is partly caused by the RDFs list [in lib/handlers/get.js#25](https://github.com/linkeddata/ldnode/blob/e908b51022b1450f190023b0ba5f6c78cf63022c/lib/handlers/get.js#L25) -- it tries to parse the container, which eventually ends up calling `lib/utils.js` / `stringToStream()` with an undefined value.
This results in an uncaught type error:

```
Uncaught TypeError: Cannot read property 'length' of undefined
```

To do:

1. Harden `stringToStream()` -- needs to return if the argument is not actually a string.
2. Fix the handling/parsing of `application/ld+json` on containers.
3. Fix or remove `application/nquads` and `application/n-quads` from the list in the get handler.

dmitrizagidulin:
See PR #223 for failing test.

dmitrizagidulin:
Fixed by https://github.com/linkeddata/ldnode/pull/225
Closing.

PR #224, opened 2016-01-28
Fix crash when requesting a container as `application/ld+json`

dmitrizagidulin:
Starting a PR:

- add failing tests that highlight issue #223 (doing a GET on a container with `Accept: application/ld+json`).

### DIFF
--- a/test/http.js
+++ b/test/http.js
@@ -150,6 +150,12 @@ describe('HTTP APIs', function () {
   })
 
   describe('GET API', function () {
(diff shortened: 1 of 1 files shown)


nicola:
Strange.
The function `translate` should handle this..

nicola:
So, I have been digging a couple of minutes, the issue is only with containers, since we have a passing test for resources in `test/formats`

dmitrizagidulin:
Closing this PR, superceded by https://github.com/linkeddata/ldnode/pull/225

PR #225, opened 2016-01-29
now containers return jsonld if requested so

nicola:
This solution is a bit of a hack.

`Jsonld` apparently does not like relative urls (it needs a base `@id:`). When we list the containment triple, we use relative paths; so, when we serialize to `jsonld`, the serializer complains for this reason.

This fix forces the containment triples to be formatted in text/turtle and later on converted to jsonld (in constrast on how it should be: directly serialize to jsonld). This _hack_ works because when we translate from `turtle` to `jsonld` we pass a base url that `jsonld` will be using (so it will drop relative paths).

Before we merge this, it would be great to have a better understanding on the jsonld library.

__Note__: (1) in the code I left two TODO showing when I force the translation to happen. (2) I readded the test used by dmitry

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -13,6 +13,7 @@ var utils = require('./utils')
 var ns = require('./vocab/ns').ns
 var error = require('./http-error')
 var stringToStream = require('./utils').stringToStream
+var serialize = require('./utils').serialize
 var extend = require('extend')
 var doWhilst = require('async').doWhilst
 var rimraf = require('rimraf')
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
I propose we merge this fix, and open an issue to refactor/understand the jsonld library.

deiu:
Is it ready for merge?

deiu:
:+1: from me

dmitrizagidulin:
Looks ready to me. It passes the tests, and Dokieli works now. :+1:

dmitrizagidulin:
@nicola - feel free to open a separate ticket to address whatever you feel needs to be refactored or added re `jsonld`

PR #229, opened 2016-02-02
WIP - URI to file system mapping

deiu:
Started some preliminary work on HTTP file mapping. This is based on a discussion between Tim and I  that resulted in a [DesignIssues post by Tim](https://www.w3.org/DesignIssues/HTTPFilenameMapping.html).


RubenVerborgh:
Addressed by #643, will be incorporated by #662.

PR #230, opened 2016-02-02
Minor fixes

deiu:


deiu:
Ready for merge.

lib/handlers/get.js
```diff
@@ -93,8 +93,8 @@ function handler (req, res, next) {
     // Translate from the contentType found to the possibleRDFType desired
     translate(stream, baseUri, contentType, possibleRDFType, function (err, data) {
       if (err) {
-        debug('error translating: ' + req.originalUrl + ' ' + contentType + ' -> ' + possibleRDFType + ' -- ' + 500 + err.message)
-        return next(error(500, err.message))
+        debug('error translating: ' + req.originalUrl + ' ' + contentType + ' -> ' + possibleRDFType + ' -- ' + 406 + ' ' + err.message)
+        return next(error(406, err.message))
```
nicola: this is not always true, since the error could be because of the translation issues, so this should be 500 and maybe before we check that if possibleRDFType is empty, then we return 406 (which I think we already do)

deiu: OK, I'll revert to 500 for now, at least until we add tests that prove otherwise.

lib/ldp.js
```diff
@@ -473,11 +479,12 @@ LDP.prototype.deleteResource = function (filename, callback) {
 
 LDP.prototype.getAvailablePath = function (host, containerURI, slug, callback) {
   var self = this
+  var exists
 
   if (!slug) {
     slug = uuid.v1()
   }
-  var exists
```
nicola: this is used later on in the code (487 | 494), it can't be deleted or otherwise the other would become a global variable (and we don't want that)

PR #248, opened 2016-02-12
Fix cors expose headers

deiu:


dmitrizagidulin:
Couple more problems with our Proxying code (tested on this PR's branch):

1. The proxy is not passing along the `Accept:` header to the original server
2. The proxy is not returning the response's `Content-Type:` and `Content-Length:` headers.
3. The proxy is injecting a `Access-Control-Allow-Credentials: true` header, even if original server did not return it.

**To test**

Melvin's server correctly returns 'text/turtle' if requested directly:

```bash
curl -s http://melvincarvalho.com/#me -H 'Accept: text/turtle' -v | grep "Content-Type"

< HTTP/1.1 200 OK
< Access-Control-Allow-Origin : *
< Content-Length: 7471
< Content-Type: text/turtle
<
@prefix dcterms: <http://purl.org/dc/terms/> .
```

However, using LDNode's proxy, both the 'Accept:' request header and 'Content-Type' response headers are dropped:

```bash
curl "https://localhost:8443/proxy?uri=http%3A%2F%2Fmelvincarvalho.com%2F%23me" \
    -H 'Accept: text/turtle' -vk

< HTTP/1.1 200 OK
< X-Powered-By: ldnode
< Vary: Origin
< Access-Control-Allow-Credentials: true
< Access-Control-Expose-Headers: User, Location, Link, Vary, Last-Modified, Content-Length
< Date: Wed, 17 Feb 2016 18:41:27 GMT
< Connection: keep-alive
< Transfer-Encoding: chunked
<
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN" "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-1.dtd">
```

Also note the injected `Access-Control-Allow-Credentials: true` header.

deiu:
These issues should be raised separately. I don't think they belong in this PR.

dmitrizagidulin:
@deiu - no prob. Opening them separately.

dmitrizagidulin:
(Opened issue #251 and #252, respectively. The `Content-Length:` issue already exists: #250

PR #256, opened 2016-02-18
Rewriting flags order and description

nicola:
Also converted idp to allow-signup

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,27 +28,29 @@ The command line tool has the following options
 Usage: ldnode [options]
 
 Options:
-  -v, --verbose               Print the logs to console
-  --version                   Print current ldnode version
-  -m, --mount                 Relative URL from which to serve the Linked Data Platform (default: '/')
-  -r, --root                  Root location on the filesystem to serve resources
-  -p, --port                  Port to use
-  -c, --cache                 Set cache time (in seconds), 0 for no cache
-  -K, --key                   Path to the ssl key file
-  -C, --cert                  Path to the ssl cert file
-  --webid                     Enable WebID+TLS authentication
-  -idp, --identity-provider   Allow registration of WebIDs
-  --create-admin              Allow a user to set up their initial identity in single-user mode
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
What order are the flags in now, as a result? :)

nicola:
```
Usage: ldnode [options]

Options:
   --version          Print current ldnode version
   -v, --verbose      Print the logs to console

   --root             Root folder to serve (defaut: './')
   --port             Port to use
   --webid            Enable WebID+TLS authentication
   --ssl-key          Path to the SSL private key in PEM format
   --ssl-cert         Path to the SSL certificate key in PEM format
   --allow-signup     Allow users to register their WebID on subdomains

   --create-admin     Allow a user to set up their initial identity in single-user mode
   --no-live          Disable live support through WebSockets
   --default-app      URI to use as a default app for resources (default: https://linkeddata.github.io/warp/#/list/)
   --proxy            Use a proxy on example.tld/proxyPath
   --suffix-acl       Suffix for acl files (default: '.acl')
   --suffix-meta      Suffix for metadata files (default: '.meta')
   --session-secret   Secret used to sign the session ID cookie (e.g. "your secret phrase")
   --no-error-pages   Disable custom error pages (use Node.js default pages instead)
   --error-pages      Folder from which to look for custom error pages files (files must be named <error-code>.html -- eg. 500.html)
   --mount            Serve on a specific URL path (default: '/')
   --force-user       Force a WebID to always be logged in (useful when offline)
```

dmitrizagidulin:
I meant more like - what's the logic behind the order? :)

nicola:
- I improved the description of the flags
- I moved the flags around so that you can have more frequent flags at the top (totally subjective)

dmitrizagidulin:
Ah, nice. +1 to this.

dmitrizagidulin:
Any reason why we haven't merged this?

nicola:
I am not sure, I think we didn't want to merge it since we need a much better cleaning - so merging commands or having new ones. I think @deiu didn't want to merge something that would have been a breaking change leading to another breaking change

PR #267, opened 2016-03-02
Add settings/ and accounts/ to .gitignore

dmitrizagidulin:


deiu:
Do we actually need this? You should not run ldnode (or set the root) in the git repository.

dmitrizagidulin:
@deiu - what do you mean? Developers will often run ldnode in the git repo (including myself). This is to prevent them accidentally committing the created dirs.

deiu:
In fact, what we need is to block running ldnode without specifying a root directory.

nicola:
@deiu why? it is incredibly handy to just type "ldnode" and everything works!

deiu:
Sure, it works, but you end up with a lot of garbage in your dev directory. In fact, I didn't mean running `ldnode` after installing from npm but running ldnode from the repo.

dmitrizagidulin:
Why don't we just merge this PR and let developers decide where they want to run it from :)

csarven:
There is room for both:

git cloners are going to do whatever is obvious and convenient ("hey, is this going to work right out of the box? Do I really have to read/configure.. "). Adding some .gitignores is a simple-enough measure so that it does not pose any conflicts for their next pull. Granted, the long-term solution is to use the config properly, but why force that right from the start.

IMO, removing all frictions to get something working out of the box should be a priority. Personally, I want to see what works with the defaults - I can then later decide what to do with the configs once I'm better familiar.

Having said all that, the out of the box is more fitting for the npm approach than it is to git repo. So, whatever... :)

nicola:
+1 for me

PR #271, opened 2016-03-04
Implementing Account Recovery via email

nicola:
The following does:

- Use memory to store tokens (maybe very bad)
- Expose API to request tokens and being logged in via a token
- send email when requesting for a token


nicola:
This is holding, since we may delegate email reset to external libraries

PR #273, opened 2016-03-04
Fix redirect

deiu:
Redirect `/foo` to `/foo/` if missing slash and the resource is a container.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -62,15 +62,12 @@ function handler (req, res, next) {
 
     // Redirect to container with / if missing
     if (container && path.lastIndexOf('/') !== path.length - 1) {
-      debug(contentType)
       return res.redirect(301, _path.join(path, '/'))
     }
 
     // Till here it must exist
     if (!includeBody) {
       debug('HEAD only')
-      // res.set('Content-Type', contentType)
-      // res.set()
       res.sendStatus(200)
       return next()
     }
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -2,7 +2,7 @@ module.exports = handler
 
 var fs = require('fs')
 var glob = require('glob')
-var path = require('path')
+var _path = require('path')
 var $rdf = require('rdflib')
 var S = require('string')
 var async = require('async')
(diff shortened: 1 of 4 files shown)


lib/handlers/get.js
```diff
@@ -60,9 +60,17 @@ function handler (req, res, next) {
       return next(err)
     }
 
+    // Redirect to container with / if missing
+    if (container && path.lastIndexOf('/') !== path.length - 1) {
+      debug(contentType)
```
nicola: can you add some details to this log?

lib/handlers/get.js
```diff
@@ -60,9 +60,17 @@ function handler (req, res, next) {
...
+      return res.redirect(301, _path.join(path, '/'))
```
nicola: do we really want to redirect every time? I am not sure I understand why this should _always_ happen

lib/handlers/get.js
deiu: oops, forgot to remove that debug

lib/handlers/get.js
deiu: Yes, because `/foo` and `/foo/` are not the same thing.

deiu:
Hold on, removing debugs.

nicola:
Sorry I just tried out the new feature from github that lets you merge the master into the PR so that the PR is up to date

lib/ldp.js
```diff
@@ -383,7 +383,7 @@ LDP.prototype.get = function (host, reqPath, baseUri, includeBody, contentType,
 
     // Just return, since resource exists
     if (!includeBody) {
-      return callback(null, stats)
+      return callback(null, stats, contentType, stats.isDirectory())
```
nicola: if we pass stats there is no need to pass `stats.isDirectory()`

nicola: also when is the `contentType` used?

deiu: The problem occurs later on, since `container` will be undefined. This is the relevant line: https://github.com/linkeddata/ldnode/blob/fix-redirect/lib/handlers/get.js#L64

PR #282, opened 2016-03-11
Fix ACL check for internal index.html redirect

deiu:
Given a private container `/foo/` and a public document `/foo/index.html`, lnode's ACL system currently checks if `/foo/` is readable before passing the request to the index handler, therefore blocking access to `/foo/index.html`.

This PR introduces a fix, where the index handler takes precedence over the ACL check for read operations (GET).

### DIFF
--- a/lib/handlers/index.js
+++ b/lib/handlers/index.js
@@ -16,26 +16,25 @@ function handler (req, res, next) {
     if (err) return next()
 
     res.locals.path = req.path
-    if (stats.isDirectory()) {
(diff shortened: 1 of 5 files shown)


lib/handlers/index.js
```diff
@@ -18,40 +16,27 @@ function handler (req, res, next) {
     if (err) return next()
 
     res.locals.path = req.path
```
nicola: no need for this line here

deiu:
OK, I've removed that obsolete line.

Issue #283, opened 2016-03-13
typo in docs: -idp should be --idp


melvincarvalho:
Occurs in README and ldnode --help

dmitrizagidulin:
Turns out that node or express doesn't allow multi-letter abbreviated command line params. (so, `-i` is ok but `-idp` is invalid). Fix pending.

nicola:
Fixed in latest.
This is now `--allow-signup`

PR #284, opened 2016-03-13
Fix cli params (-idp, -sA, -sM, -fU)

dmitrizagidulin:
Fixes issue #283.


nicola:
should we go with this instead -> https://github.com/linkeddata/ldnode/pull/256 ?
The one I am linking re-orders and removes abbraviations

dmitrizagidulin:
Sure, that one works too!

nicola:
let's merge it then!

nicola:
Will merge   #256

PR #285, opened 2016-03-14
make webid-tls auth by default

nicola:


nicola:
This fails since the tests need to be updated about this new change

PR #291, opened 2016-03-16
Adding renegotiation on ACL fail

nicola:
There are some issues with that, resources will show a 401 instead of a 403.


RubenVerborgh:
No longer applies since incorporating OIDC.

PR #293, opened 2016-03-16
adding --strict-origin flag

nicola:


dmitrizagidulin:
What's the context? Did I miss an issue somewhere explaining this? :)

nicola:
the strict flag enforces the same origin policy on the acl, in other words, if the ACL has no origin specified, no one is granted access except the host website.

PR #294, opened 2016-03-17
Cookie max age

nicola:
I set max age to be 24 hours, this time is reset every time you do an authenticated request.


deiu:
> this time is reset every time you do an authenticated request.

Including sending a previous cookie?

nicola:
@deiu see https://github.com/expressjs/session#rolling

deiu:
:+1:

PR #295, opened 2016-03-17
Avoid walking the path to https:/.acl

deiu:
Before, the ACL code was walked the path in reverse order, leading up to `https:/.acl` instead of `https://localhost/.acl`.

### DIFF
--- a/lib/acl.js
+++ b/lib/acl.js
@@ -256,6 +256,7 @@ function possibleACLs (uri, suffix) {
   }
 
   var times = parsedUri.pathname.split('/').length
+  // TODO: improve temporary solution to stop recursive path walking above root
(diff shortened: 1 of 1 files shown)


lib/acl.js
```diff
@@ -256,6 +256,10 @@ function possibleACLs (uri, suffix) {
   }
 
   var times = parsedUri.pathname.split('/').length
+  if (parsedUri.pathname.endsWith('/')) {
```
nicola: maybe add a comment with a // TODO ?

deiu: A TODO for what?

nicola: for cleaning the hack?

PR #298, opened 2016-03-24
Add fallthrough for failed auth

deiu:
The auth logic should not return 403 if a user fails to authenticate. The process should continue as if the user did not provide any credentials.

### DIFF
--- a/lib/handlers/authentication.js
+++ b/lib/handlers/authentication.js
@@ -2,7 +2,6 @@ module.exports = handler
 
 var webid = require('webid/tls')
 var debug = require('../debug').authentication
-var error = require('../http-error')
 
 function handler (req, res, next) {
   var ldp = req.app.locals.ldp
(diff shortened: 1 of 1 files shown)


timbl:
If the person has a cert problem they need to fix, will they get that error when eventually they fail to access the resource?  A bit worried about losing error info here.  Just clicking through with no cert or not having a cert is one thing, but using a cert which has bugs is another.

deiu:
@timbl - this fix does not address the use case you described. In fact, it fixes a general issue of returning `HTTP 403` if authentication fails. We need to keep authentication separate from access control.

nicola:
How do we notify that authentication has failed?
Maybe instead of a 403, a 500?

This case it is a tricky one since authentication come with each request it is not done via a different url, let's say `login`, when you can tell to the users that their authentication failed.

I am happy to merge this, since I think this will be accounted in future auth protocols

deiu:
Yes, we will have to handle it in a different fix. In any case, it should not return 403.

PR #322, opened 2016-04-19
Minor layout refactoring

dmitrizagidulin:
- Wrapped code to 80 chars
- Extracted nested inner functions to module-level scope in lib/ldp.js

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -53,7 +53,8 @@ function LDP (argv) {
   }
 
   if (this.fileBrowser !== false) {
-    this.fileBrowser = argv.fileBrowser || 'https://linkeddata.github.io/warp/#/list/'
+    this.fileBrowser = argv.fileBrowser ||
+      'https://linkeddata.github.io/warp/#/list/'
   }
 
   if (this.dataBrowser !== false) {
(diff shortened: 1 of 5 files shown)


lib/ldp.js
```diff
@@ -19,6 +19,160 @@ var doWhilst = require('async').doWhilst
...
+          return callback(error(err, 'Can\'t parse container metadata'))
+        }
+        return callback(null, metadataGraph)
+      })
+    } else {
+      return callback(null, $rdf.graph())
+    }
+  })
+}
+
```
nicola: I am not sure that taking this function out is a good idea.
They have one particular purpose bound to one particular function, putting them out is a bit confusing.
I think I like them being well encapsulated in the scope of the function. Also, these functions may even disappear or should be in a library (like [folder-to-rdf](https://github.com/nicola/folder-to-rdf) was initially)

dmitrizagidulin: Defining multiple nested functions inside a function, and then calling them, is a pretty serious anti-pattern tho. 
Feel free to move em to a separate file or module tho.

nicola: I agree, it doesn't feel right.
Putting out a function in the middle of many others, doesn't feel right either.

Moving them in a file sounds great. Recently I keep on seeing this (the anti-pattern that you mention) happen more and more in other javascript projects that I am following

dmitrizagidulin:
@nicola how's this? I moved the inner functions in question to their own module file, `ldp-container.js`. (We're gonna refactor these files anyway when we move to a modular backend API).

nicola:
Perfect @dmitrizagidulin!

PR #328, opened 2016-05-02
Drop create admin

nicola:
Fixes https://github.com/linkeddata/ldnode/pull/326
Helps https://github.com/linkeddata/ldnode/issues/327

If there is no root acl, the user is asked to log in (this could be avoided with a possible new flag `--owner` that will create the acl automatically)


nicola:
@rhiaro

PR #331, opened 2016-05-03
renaming from ldnode to solid-server

nicola:


dmitrizagidulin:
:+1: , seems to work fine after the rename.

nicola:
I will wait for @deiu to merge since this is a big change

PR #333, opened 2016-05-04
new command line

### LINKED ISSUE #332: Implement ability to specify an external config file to launch solid server
(as discussed with @nicola).
Implement the ability for solid-server to load settings from a `config.json` file (though the settings can be also overridden via commandline flags as currently done).
(Needed for the OIDC integration work).

nicola:
fix #332

#### THIS IS A BREAKING CHANGE

The new command line works in the following way:

```
$ solid init
// will ask questions and generate config.json accordingly (in the folder where you are running it)
$ solid init --advanced
// will ask extra question
```

to run solid-server

```
$ solid start
// this will load the config.json
$ solid start --port 8080
// this will overwrite the config.json
```

**Note**: there is no need to create the config.json with `solid init`, so you can just run `solid start` with all the flags you need (so previous `ldnode` command line still works)

Try it, it feels much better than before


deiu:
👍 from me.

PR #336, opened 2016-05-06
oidc + email support to IdentitProvider

nicola:


nicola:
will happen in a different PR

PR #338, opened 2016-05-09
Account Recovery build on top of `email-support` branch

nicola:


nicola:
A lot of progress, I am currently having an rdflib error, but it could be my bad implementation

PR #330, opened 2016-05-02
OIDC Integration

dmitrizagidulin:
Server-side implementation of the [WebID-OIDC (OpenID Connect) authentication spec](https://github.com/solid/webid-oidc-spec).

Major features:

1) Replaces WebID-TLS authentication with WebID-OIDC. WebID-TLS auth will continue to be available on a maintenance `v3.5` branch of solid-server.
2) Moves tests into one of two directories, `/test/unit` and `/test/integration`, corresponding.
3) Improves post-install "out of the box" user experience -- sets up server front page in multi-user mode, installs default signup & signin apps, etc.
4) Increases unit and integration test coverage for the server.

PR #339, opened 2016-05-10
Accounts API v1

nicola:
List of APIs we are planning on implementing

#### LDNODE
- [x] `/api/accounts/signin`: 302 to oidc if possible
- [x] `/api/accounts/signout`: removes the session
- [x] `/api/accounts/recover`: generates a token and contacts the user sending the token
- [x] `/api/accounts/validateToken`: user can claim a token, gets the cookies and gets redirected to /
- [x] `/api/accounts/new`: creates a new account on LDP and create it under OIDC https://github.com/solid/node-solid-server/pull/349

#### OIDC integration
- [x] `/api/accounts/signin`
- [x] `/api/accounts/new`: call to OIDC to create new user https://github.com/solid/node-solid-server/pull/349

for v2 see https://github.com/solid/node-solid-server/issues/350

## Other things to do:

- [x] move all the endpoints to `/api`
- [ ] get OIDC finalized https://github.com/solid/node-solid-server/pull/330


lib/api/accounts/signin.js
```diff
@@ -0,0 +1,32 @@
...
+
+function signin () {
+  return (req, res, next) => {
+    if (!validUrl.isUri(req.body.webid)) {
+      return res.status(406).send('This is not a valid URI')
+    }
+
+    request({ method: 'OPTIONS', uri: req.body.webid }, function (err, req) {
+      if (err) {
+        res.status(406).send('Did not find a valid endpoint')
```
dmitrizagidulin: Why 406 here? I thought that error only dealt with `Accept` headers & content types?

nicola: if you send a bad URI, you should know about it

dmitrizagidulin: I agree. But this should be a 400 error, not 406.

nicola: even for the 409 happening later?

Issue #309, opened 2016-03-31
[Proxy] do not proxy secure networks (LAN)


deiu:
The proxy system should not allow requests to LAN servers (IP ranges).

10.0.0.0/8 IP addresses: 10.0.0.0 -- 10.255.255.255
172.16.0.0/12 IP addresses: 172.16.0.0 -- 172.31.255.255
192.168.0.0/16 IP addresses: 192.168.0.0 – 192.168.255.255

nicola:
Fixed in github.com/solid/node-solid-server/pull/342

PR #342, opened 2016-05-11
Safer proxy

nicola:
Fixes https://github.com/solid/node-solid-server/issues/309

- skip local network
- avoid crashing when uri is not valid
- added tests


test/params.js
```diff
@@ -25,6 +25,22 @@ describe('LDNODE params', function () {
         .expect(200, done)
     })
 
+    it('should return local network requests', function (done) {
```
deiu: Isn't this description misleading? The test exists to make sure that the server does **not** serve anything that is on the local network.

nicola: sweet

deiu:
👍 Looks good.

Issue #366, opened 2016-05-16
Add a pointer to the WebID from the root meta


deiu:
The root meta file should have a pointer to the WebID, saying that this particular space (domain) is an account of said WebID.

nicola:
Fixed in https://github.com/solid/node-solid-server/pull/371

PR #371, opened 2016-05-17
Discover WebID from root account

deiu:
Add a triple to the root meta file that links an account (domain) to the WebID of the owner. Implements #366.


deiu:
Fixes #366

Issue #375, opened 2016-05-18
Empty POST to /api/accounts/new should fail in Multi User/--idp mode


dmitrizagidulin:
If you launch solid-server in `--idp` mode, an empty post (with no username) to `/api/accounts/new` should fail with a 400 (with the error message `Username required`)
Currently fails with `Account already exists`

nicola:
I rephrase the question in this way:

Should the user create a new account by doing a POST on:

- POST https://databox.com/api/accounts/new {username: 'nicola'}
- https://nicola.databox.com/api/accounts/new (no username needed to be passed)

?

The second is a sort of a claim of a url (also this is what we have done so far)

dmitrizagidulin:
I'm not sure I'm understanding that second option. What http verb is that, still a post?

nicola:
yes!

dmitrizagidulin:
Ohhhh I see!

dmitrizagidulin:
Yeah, that seems reasonable. We should differentiate between those two use cases (and not require usernames in the second case)

dmitrizagidulin:
After further discussion with @nicola, we agree that we currently don't need the concept of usernames. Closing this issue.

PR #376, opened 2016-05-18
Fail with a 400 Username required on /accounts/new in --idp mode

dmitrizagidulin:
Fixes issue #375.

### DIFF
--- a/lib/identity-provider.js
+++ b/lib/identity-provider.js
@@ -529,13 +529,13 @@ IdentityProvider.prototype.post = function (req, res, next) {
   const idpMode = req.app.locals.ldp.idp
   if (!req.body) {
     debug('Options missing')
-    var err = new Error('Settings to create the account have not been passed!')
+    let err = new Error('Settings to create the account have not been passed!')
     err.status = 400
     return next(err)
   }
   if (idpMode && !req.body.username) {
     debug('Username missing')
-    var err = new Error('Username required')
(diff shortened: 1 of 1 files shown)


nicola:
great!
can we add a test for this? (I can do)

dmitrizagidulin:
@nicola if you could, that would be great

lib/identity-provider.js
```diff
@@ -526,10 +526,17 @@ IdentityProvider.prototype.getGraph = function (uri, callback) {
 
 // Handle POST requests on account creation
 IdentityProvider.prototype.post = function (req, res, next) {
+  const idpMode = req.app.locals.ldp.idp
```
nicola: rules of 4 dots..

dmitrizagidulin: I'm going to push back on this. That line seems completely legit to me.  (I also don't think the IDP mode should be stored in `ldp` anyways. So if anything, it should be req.app.locals.idp)

nicola: this will be in req.app.locals.solid soon (so req.app.locals.ldp), we should not populate req.app.locals of too many variables since this name space could also be used by other variables.

Rules of 4 dots applies it anywhere!

dmitrizagidulin:
After further discussion with @nicola, we agree that we currently don't need the concept of usernames. Closing this PR.

PR #380, opened 2016-05-23
Using 'url' to get the base URI for a resource

antoniogarrote:
'baseUrl' was returning the empty string, leading to a url for the
resource always matching the domain. 
This was causing an error in the JSON-LD serialization:

```
curl -iv -X GET -H "Accept: application/ld+json" -k https://localhost:8443/20160512191736/

[{"@id":"https://localhost:8443", ...}]
```

```
curl -iv -X GET -H "Accept: application/ld+json" -k https://localhost:8443/20160512191736/Library/

[{"@id":"https://localhost:8443", ... }]
```

The change fixes the error and now it returns the right value for the '@id' property.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -51,7 +51,7 @@ function uriAbs (req) {
 }
 
 function uriBase (req) {
-  return uriAbs(req) + (req.baseUrl || '')
+  return uriAbs(req) + (req.url || '')
 }
 
 function pathBasename (fullpath) {
--- a/test/formats.js
+++ b/test/formats.js
@@ -1,6 +1,7 @@
 var supertest = require('supertest')
 var ldnode = require('../index')
 var path = require('path')
+var assert = require('chai').assert
 
 describe('formats', function () {
   var ldp = ldnode.createServer({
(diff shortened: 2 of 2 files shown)


dmitrizagidulin:
Thanks @antoniogarrote! We'll take a look shortly.

nicola:
Hey @antoniogarrote nice to see you here :)

I don't understand your example. Can you show me a valid and an invalid outcome? Paths are meant to be relative to the base (`https://127.0.0.1:8443` in your case), let me know

antoniogarrote:
Hi @dmitrizagidulin, @nicola , thanks for your work on Solid.

I have added a test to try to clarify the issue.

Basically, I was always getting always only protocol+domain as the URI of the resource I was trying to fetch. So if the resource is 'https://domain/path/to/document#me',  I was only getting 'https://domain#me' in the JSON-LD payload when I requested 'https://domain/path/to/document'.
It only happened if I requested the JSON-LD serialisation.

I tracked to the way the base uri was being generated in 'lib/utils.js' and the way it is passed to the JSON-LD library in the translation logic.

I don't know much about express, so I hope it is not because of some kind of misconfiguration on my side.

Thanks!

nicola:
Hey @antoniogarrote thanks a lot, I understand what is the problem now.
This is very interesting, apparently it works well with turtle but not with JSON-LD,

(super thanks for your tests!)

nicola:
update: Ok, this is an issue that happens when `translate` is called on resources not on containers

nicola:
I didn't have time to go through this this week, will fix this soon

dmitrizagidulin:
Possibly related to https://github.com/solid/node-solid-server/issues/476 ?

dmitrizagidulin:
Closing this PR - the fix is being handled instead in PR #496. Thanks again to @antoniogarrote for bringing this to our attention.

PR #383, opened 2016-05-23
Fix invalid relative ./ usage in .gitignore

dmitrizagidulin:


nicola:
I have added the `/` so that only the root ones are matched (it was hiding `/lib/api/accounts`)

Issue #387, opened 2016-05-26
[oidc] Do not create solid account dir if OIDC user creation fails


dmitrizagidulin:
Make sure the solid user directory is not created if there's an OIDC user creation error (like invalid password, or email already taken, etc).

dmitrizagidulin:
Implemented. closing.

PR #390, opened 2016-05-27
Do not create a Solid account folder if OIDC user creation fails

dmitrizagidulin:
Fixes issue #387.
Also adds some doc strings.

### DIFF
--- a/lib/identity-provider.js
+++ b/lib/identity-provider.js
@@ -532,55 +532,6 @@ IdentityProvider.prototype.getGraph = function (uri, callback) {
 }
 
 /**
- * Sends a request to the OIDC Provider's Users API endpoint, to create a new
- * user record with the provider.
- * @method createOIDCUser
- * @param oidcRpClient {OidcRpClient} Local/trusted OIDC client
- * @param webId {String} WebID URL of the new user to be created
- * @param options {Object} User options hashmap
- * @param options.password {String} User's signin password. NOTE: Must be 8+
- *   characters, mix of alpha and numeric
- * @param [options.email] {String} User's email address (for recovery etc)
- * @param [options.name] {String} User's name
- * @return {Promise}
- */
-IdentityProvider.prototype.createOIDCUser =
-  function createOIDCUser (oidcRpClient, webId, options = {}) {
-    if (!oidcRpClient) {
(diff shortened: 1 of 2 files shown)


lib/identity-provider.js
```diff
@@ -524,15 +531,77 @@ IdentityProvider.prototype.getGraph = function (uri, callback) {
...
-IdentityProvider.prototype.post = function (req, res, next) {
-  if (!req.body) {
+/**
+ * Sends a request to the OIDC Provider's Users API endpoint, to create a new
+ * user record with the provider.
+ * @method createOIDCUser
+ * @param oidcRpClient {OidcRpClient} Local/trusted OIDC client
+ * @param webId {String} WebID URL of the new user to be created
+ * @param options {Object} User options hashmap
+ * @param options.password {String} User's signin password. NOTE: Must be 8+
```
nicola: this should be optional too, we can of course pass a setting flag that is `--auth-password` or similar, we shouldn't enforce password as auth schema

lib/identity-provider.js
```diff
@@ -524,15 +531,77 @@ IdentityProvider.prototype.getGraph = function (uri, callback) {
...
+ *   characters, mix of alpha and numeric
+ * @param [options.email] {String} User's email address (for recovery etc)
+ * @param [options.name] {String} User's name
+ * @return {Promise}
+ */
+IdentityProvider.prototype.createOIDCUser =
```
nicola: createOIDCUser shouldn't be in the Identity provider package (maybe put that in the oidc one?)

dmitrizagidulin: > createOIDCUser shouldn't be in the Identity provider package

Yeah, I agree. Should we factor out the tls-related user creation files somewhere too?

lib/identity-provider.js
dmitrizagidulin: > this should be optional too

Not sure I undersrtand. It's not optional, currently. (in our oidc integration)

dmitrizagidulin:
@nicola - removed the oidc-related user code for now. Ready for review.

dmitrizagidulin:
(The fix to #387 has been merged into the `dz_oidc` branch, however)

PR #412, opened 2016-07-03
Add Accept-Post to Access-Control-Expose-Headers

csarven:
Required in order to expose a non-simple header for CORS.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -21,7 +21,7 @@ var corsSettings = cors({
   methods: [
     'OPTIONS', 'HEAD', 'GET', 'PATCH', 'POST', 'PUT', 'DELETE'
   ],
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
I thought about adding that header to `Access-Control-Expose-Headers`, in my original PR. And then decided against it -- the `Access-Control-...` header only applies to non-OPTIONS requests. Whereas the `Accept-Post` header is only returned with OPTIONS. Hence, we don't need it.

csarven:
Consider:

Accept-Post is a MUST in OPTIONS in LDP, but it is not strictly for OPTIONS. It can be present in other methods.

Browsers still throw a `Refused to get unsafe header` if Accept-Post is not present. There is no harm in adding this as far as I can tell.

dmitrizagidulin:
it's only implemented to return with OPTIONS, tho...

PR #419, opened 2016-07-12
Implement COPY verb support

dmitrizagidulin:
Implements issue https://github.com/solid/solid/issues/49, adds support for an
HTTP Copy verb that allows importing an external resource to a given Solid destination.


lib/handlers/copy.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Handles HTTP COPY requests to import a given resource (specified in the
+ * `Source:` header) to a destination (specified in request path).
+ * For the moment, you can copy from public resources only (no auth delegation
+ * is implemented), and is mainly intended for use with
+ * "Save an external resource to Solid" type apps.
+ * @method handler
+ */
+function handler (req, res, next) {
```
nicola: hey @dmitrizagidulin in order to make our life easy once we want to move to another type of store, we are abstracting all of the file system stuff into ldp.js. The way I see it is the following:

1) File system
2) LDP.copy
3) HTTP Server handling

this `copy.js` should call LDP.copy (that will be implemented with fs calls and a callback*) and in the callback* do the http handling (see get.js for example!)

dmitrizagidulin: Ah, I see what you're saying. Sure, I can refactor it to do that, np.

dmitrizagidulin:
@nicola - Refactored the code to call LDP copy, per your request.

dmitrizagidulin:
@nicola - I'll be merging this PR later this week. last chance to protest :)

PR #424, opened 2016-08-04
Refactor to use external solid-namespace library

dmitrizagidulin:
Refactor to use the `solid-namespace` lib, remove the local `lib/vocab/ns.js` code.


nicola:
LGTM

dmitrizagidulin:
@nicola thanks! merging.

PR #425, opened 2016-08-04
Move debrack() to utils.js, remove unused vocab/rdf.js functions

dmitrizagidulin:
The only function from vocab/rdf.js used anywhere was `debrack()`.
Moved it to `utils.js`, removed all the other unused functions in that
file, and the file itself.


lib/vocab/rdf.js
```diff
@@ -1,64 +0,0 @@
-/*
- * Copyright 2014 IBM Corporation.
```
nicola: I don't know what's up with this

nicola:
LGTM (except the copyright note)

lib/vocab/rdf.js
dmitrizagidulin: Lol. wasn't me, didn't do it :)

dmitrizagidulin: (It's fine, though, we're deleting that file anyway.)

PR #428, opened 2016-08-09
Refactor acl.js to prep for external solid-permissions lib

dmitrizagidulin:
- Move `lib/acl.js`/`allow()` handler to `lib/handlers/allow`
- Extract ACL class to `lib/acl-checker.js` (convert to es6)
- Remove S() usage (use native String functions instead)
- Pass in the debug() logger as a dependency injection for ACLChecker
- Use new rdflib graph.match() natively


package.json
```diff
@@ -57,7 +57,7 @@
     "node-uuid": "^1.4.3",
     "nodemailer": "^2.3.2",
     "nomnom": "^1.8.1",
-    "rdflib": "^0.8.0",
+    "rdflib": "git://github.com/linkeddata/rdflib.js#dz_match_alias",
```
nicola: all LGTM, but this

lib/acl-checker.js
```diff
@@ -0,0 +1,230 @@
...
+
+const async = require('async')
+const path = require('path')
+const url = require('url')
+
+const DEFAULT_ACL_SUFFIX = '.acl'
+
+class ACLChecker {
+  constructor (options = {}) {
+    this.debug = options.debug || console.log
```
nicola: console.log would not work, this should be console.log.bind(console)

lib/acl-checker.js
```diff
@@ -0,0 +1,230 @@
...
+    this.fetch = options.fetch
+    this.strictOrigin = options.strictOrigin
+    this.suffix = options.suffix || DEFAULT_ACL_SUFFIX
+  }
```
nicola: can we still keep an optional match?
in principle, I wanted this to be able to work with any match function, in fact, this piece of code was working with rdflib and rdf-ext

nicola:
it was hard from the diff to detect changes across different files, next time let's first change the files and then move them around (it was easy to do this by looking at individual commits)

lib/acl-checker.js
dmitrizagidulin: We could. But the whole point of going the RDFJS standard API route is so that there's only one match function, right? That's the one I've implemented in rdflib.js, and it's also being added to rdf-ext. Will that be sufficient, or do you still want to inject the match function?

dmitrizagidulin:
> next time let's first change the files and then move them around

@nicola - Heh, yeah, good call, I realized that as soon as I made the PR. :) I can probably go back & do it in several commits, let's see if that's any more intelligible.

lib/acl-checker.js
dmitrizagidulin: Oh, good catch. Thanks!

package.json
dmitrizagidulin: So that bit is just to give us extra time to review this PR, while the rdflib PR https://github.com/linkeddata/rdflib.js/pull/141 is being reviewed/merged.
I'll update it with the actual published rdflib version before merging this one.

nicola: OK

lib/acl-checker.js
nicola: this would work as soon as we have rdflib working with rdfjs (with the same data model) that I don't think we have right now, right?

nicola:
@dmitrizagidulin no no worries! next time!

lib/acl-checker.js
dmitrizagidulin: We're almost there. There's two pending PRs for rdflib that bring it into conformance to the RDFJS data model. And there's similar progress to rdf-ext on the v1 branch that does the same. and the signature of the `match()` function is at least standardized, and matches.

(There's still work to be done on standardizing the *results* of `match()` - Bergos wants to standardize a DataSet model, whereas we're just returning an array for now. But in any case, I think it's safe to just call `graph.match()` and be reasonably future-proof.)

nicola: perfect then

dmitrizagidulin: If you're curious about the progress (or rdfjs compatibility for rdflib), here's the tracking issue: https://github.com/linkeddata/rdflib.js/issues/137

dmitrizagidulin:
package.json updated with proper non-github link version for rdflib. Merging.

Issue #413, opened 2016-07-05
Return correct Content-Type for JSON-LD resources (from file extension)


csarven:
If Content-Type is derived from the file extension, return the correct mime-type in the response headers. Appears to default to text/turtle.

e.g.,
- `foo.jsonld` returns text/turtle Content-Type
- `foo.json` returns application/json with HTTP 501 (this is probably a separate issue)

Possibly related:
- `foo` (in JSON-LD) returns text/turtle Content-Type
- `"Accept: application/ld+json"` returns `Error translating between RDF formats` (trying to parse it as Turtle before trying serialise as JSON-LD - see also https://github.com/solid/node-solid-server/issues/226)

dmitrizagidulin:
I suspect that whatever library we're using to derive content type from file extensions doesn't recognize `.jsonld`.. We should look into it (and if there's nothing to be done there, handle it separately in the code).

dmitrizagidulin:
Let's tackle these issues one at a time.

> `foo.jsonld` should return `application/ld+json` as its Content-Type

The main culprit is that the [mime](https://www.npmjs.com/package/mime) npm package that `node-solid-server` is using to translate from a file extension to a MIME Type just doesn't recognize `application/ld+json` as a valid type. Try this out in your terminal:

```bash
npm install mime
node
> var mime = require('mime')
> mime.extension('application/json')
'json'
> mime.extension('application/ld+json')
undefined
```

Same thing in the other direction:

```bash
> mime.lookup('.json')
'application/json'
> mime.lookup('.jsonld')
'application/octet-stream'
```

I suspect this is because the JSON-LD MIME Type is still undergoing the IANA standardization process? I think a temporary fix for this might be adding in a hardcoded custom check to `node-solid-server`, to test specifically for the `.jsonld` extension.

csarven:
https://www.iana.org/assignments/media-types/application/ld+json is standardised as it gets?

dmitrizagidulin:
@csarven aha! Excellent. Let's file an issue on the `mime` npm package, to add it?

melvincarvalho:
FYI: A few sources:

http://www.iana.org/assignments/media-types/media-types.xhtml
http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types
http://hg.nginx.org/nginx/raw-file/default/conf/mime.types

These are aggregated here:

https://github.com/jshttp/mime-db

Github uses this, for example.

[...]

PR #430, opened 2016-08-10
Switch from node-mime to mime-types lib

dmitrizagidulin:
Researching a fix for issue #413 (fixing support for the .jsonld file extension) pointed to the limitation of the `node-mime` lib we're using. This PR proposes to switch to the [`mime-types`](https://github.com/jshttp/mime-types) lib instead (which, incidentally, already has support for `jsonld`). Benefits of switching:

- It's based on https://github.com/jshttp/mime-db, which draws from IANA, Apache and Nginx registries
- It looks better maintained (take a look at the discussion in issue 108 at https://github.com/broofa/node-mime/issues/ for example)
- The API is the same (aside from not having a `default_type`, which is addressed in the PR)


nicola:
LGTM

csarven:
```
  solid:get error translating: /dokieli/inbox/abcdef false -> text/turtle -- 500 Error trying to parse <https://localhost:8443> as false:
Error: Don't know how to parse false yet:
Error: Don't know how to parse false yet
    at Object.parse (/var/www/node-solid-server.original/node_modules/rdflib/src/parse.js:58:13)
    at parse (/var/www/node-solid-server.original/lib/utils.js:104:10)
    at ReadStream.<anonymous> (/var/www/node-solid-server.original/lib/utils.js:144:7)
    at emitNone (events.js:91:20)
    at ReadStream.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:926:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9) +0ms
  solid:server Error page because of Error translating between RDF formats +0ms
```

abcdef:
```
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix as: <http://www.w3.org/ns/activitystreams#> .
@prefix oa: <http://www.w3.org/ns/oa#> . 
@prefix schema: <https://schema.org/> .
<> a as:Dislike
    ; as:object <https://localhost:8443/dokieli/index.local.html#researchers>
    ; as:updated "2016-08-08T17:26:43.768Z"^^xsd:dateTime
    ; as:actor <http://csarven.ca/#i>
    ; schema:license <https://creativecommons.org/licenses/by/4.0/>
    .
```

csarven:
Probably due this:
`Accept:text/turtle, application/ld+json, application/xhtml+xml, text/html`
`Content-Type:text/html; charset=utf-8`

dmitrizagidulin:
@csarven should be fixed in PR 431

csarven:
Note that `.nt` gets used as `text/turtle` because N-Triples is not listed at http://www.iana.org/assignments/media-types/media-types.xhtml - and so mime-types/mimes-db lib are not aware of it. N-Triple is intended to have `application/n-triples` mime type. The same parser (N3 probably) that's used for Turtle should be fine. You could manually at that in or ping upstream.

:santa:

PR #443, opened 2016-09-15
Refactor acl checker to use solid-permissions lib

dmitrizagidulin:


package.json
```diff
@@ -62,6 +62,7 @@
     "rimraf": "^2.5.0",
     "run-waterfall": "^1.1.3",
     "solid-namespace": "^0.1.0",
+    "solid-permissions": "git://github.com/solid/solid-permissions.git#dz_allows_helpers",
```
dmitrizagidulin: Temporary (for this pr review only), until PR https://github.com/solid/solid-permissions/pull/3 gets merged.

dmitrizagidulin:
@dan-f @nicola - ready for review.

lib/acl-checker.js
```diff
@@ -20,7 +22,7 @@ class ACLChecker {
     var accessType = 'accessTo'
     var possibleACLs = ACLChecker.possibleACLs(resource, this.suffix)
     // If this is an ACL, Control mode must be present for any operations
-    if (this.isAcl(resource)) {
+    if (this.isAcl(resource, this.suffix)) {
```
nicola: Can this.suffix not be used directly in `this.isAcl`?
In the rest of the code `this.isAcl` doesn't take any second parameter

lib/acl-checker.js
```diff
@@ -81,127 +83,66 @@ class ACLChecker {
...
+    let isContainer = accessType.startsWith('default')
+    let aclOptions = {
+      aclSuffix: this.suffix,
+      graph: graph,
+      host: options.host,
+      origin: options.origin,
+      rdf: rdf,
+      strictOrigin: this.strictOrigin,
+      isAcl: (uri) => { return this.isAcl(uri) },
+      aclUrlFor: (uri) => { return this.aclUrlFor(uri) }
```
nicola: at this point, can we not pass the entire ACL instance?
(maybe not)

lib/acl-checker.js
```diff
@@ -81,127 +83,66 @@ class ACLChecker {
...
-    debug('Found policies in ' + acl)
-    // Check for mode
-    var statements = this.getMode(graph, mode)
-    if (mode === 'Append') {
-      statements = statements.concat(this.getMode(graph, 'Write'))
+    let isContainer = accessType.startsWith('default')
+    let aclOptions = {
+      aclSuffix: this.suffix,
+      graph: graph,
+      host: options.host,
```
nicola: how is this used?
where is `host` set and what is this set to?

lib/acl-checker.js
```diff
@@ -81,127 +83,66 @@ class ACLChecker {
...
+        if (hasAccess) {
+          debug(`${mode} access permitted to ${user}`)
+          return callback()
+        } else {
+          debug(`${mode} access not permitted to ${user}`)
           return callback(new Error('Acl found but policy not found'))
         }
-        return callback(null)
+      })
+      .catch(err => {
```
dan-f: Could you clarify the distinction between a rejected promise and a resolved promise with a falsy value in `checkAccess`?

lib/acl-checker.js
```diff
@@ -81,127 +83,66 @@ class ACLChecker {
...
-          debug(mode + ' allowed access as FOAF agent')
-          return found(true)
-        }
-        return found(false)
-      },
-      callback)
-  }
-
+  /**
+   * @method findRule
```
dan-f: Awesome JSDoc, but what does this method do?

lib/acl-checker.js
```diff
@@ -81,127 +83,66 @@ class ACLChecker {
...
-          done)
-      },
-      function (found) {
-        if (!found) {
+    let acls = new PermissionSet(resource, acl, isContainer, aclOptions)
+    acls.checkAccess(resource, user, mode)
+      .then(hasAccess => {
+        if (hasAccess) {
+          debug(`${mode} access permitted to ${user}`)
+          return callback()
```
dan-f: Since this method is called `findRule`, I think I assumed that the callback function would get passed the access rule/policy for the given user.

lib/acl-checker.js
dmitrizagidulin: Heh, yeah, I s'pose it wouldn't hurt to add an actual description.

lib/acl-checker.js
dmitrizagidulin: `host` gets set in the `allow` handler, [here](https://github.com/solid/node-solid-server/blob/80e0675459a69a2fd25f2a6eb804a5dee05a7e84/lib/handlers/allow.js#L42).
It's used by the 'enforce strict origin' code in the permissions lib.

lib/acl-checker.js
dmitrizagidulin: We could? I'd rather be explicit, though..

dmitrizagidulin:
> It looks like you introduced some error handling in findRule. Could you add tests to cover that error handling?

I don't think there's any additional error handling introduced. The existing test suite should cover the various cases.

dan-f:
> I don't think there's any additional error handling introduced. The existing test suite should cover the various cases.

I don't see tests for the "parsing error, argument error, network errors" cases.

dmitrizagidulin:
> I don't see tests for the "parsing error, argument error, network errors" cases. 

Those belong in the permissions library, though, not here.

dan-f:
> Those belong in the permissions library, though, not here.

... which I just checked and does not have tests for such cases.

dmitrizagidulin:
I'm not sure I understand your point. This PR is a net increase in test coverage, in this acl checking code.

dan-f:
> I'm not sure I understand your point. This PR is a net increase in test coverage, in this acl checking code.

My point is that `checkAccess` is supposed to catch network (and other?) errors, and return false, that the user does not have access in such cases. But those cases are not covered either in node-solid-server nor in the permissions library. We should have a test case that demonstrates that the error handling actually works as expected, since we don't want to accidentally give permissions when the network fails, for example.

lib/acl-checker.js
nicola: perfect! works with me!

dmitrizagidulin:
@dan-f so I've been looking at this issue (testing the `.catch()` clause of checkAccess) for a bit.
I'm still not really seeing what I can do here. 
Am I testing the Promise error behavior? (That if there's an error inside a promise, it'll go into the `.catch(err)` clause?) That doesn't seem right - that's up to the Node foundation etc.
Am I testing that if `checkAccess()` calls the `callback()` with an error, the user is denied access? The rest of the existing suite checks that.
I don't really see what else I can do here.

dan-f:
@dmitrizagidulin the code path I'm thinking of is where the call to `acls.checkAccess(resource, user, mode)` returns a Promise which rejects. You told me above that your library rejects a promise in the case of "parsing error, argument error, network errors, etc," so shouldn't we verify that this consumer code handles permissions correctly when one of those errors arise?

dmitrizagidulin:
@dan-f Added `proxyquire`-based unit tests for `acl-checker.checkAccess()`.

dan-f:
Looks awesome, @dmitrizagidulin. Thanks! :+1:

PR #450, opened 2016-10-14
Convert LDP class to es6 class syntax

dmitrizagidulin:
Convert the LDP class (`lib/ldp.js`) to ES6 class syntax.
Also remove external `string` lib dependency in LDP class.

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -1,7 +1,5 @@
-module.exports = LDP
 var mime = require('mime-types')
 var path = require('path')
-var S = require('string')
 var fs = require('fs')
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
@nicola @dan-f @deiu Ready for review. (This was a syntactic change only, conversion to ES6 notation.)

dan-f:
Protip: add `?w=1' to the Github URL to ignore whitespace in the diff.

dan-f:
Fine by me, though there's more es6-ing that could take place, like replacing `var`s with `let`s or `const`s. Wondering why we're doing this now.

PR #451, opened 2016-10-14
dataBrowserOption option to specify path of db file

timbl:
When re-installing a new version of the server, it overwrites the static/databrowser.html file each time if the usr has customized it.   It is preferable to make the location of the file variable in an option.  The option is --dataBrowserPath

A flag option is called  --suppress-data-browser because calling it no-data-browser has some magic effect of interfering with a data-browser flag in the options package.

### DIFF
--- a/bin/lib/options.js
+++ b/bin/lib/options.js
@@ -115,6 +115,7 @@ module.exports = [
     default: '/proxy',
     prompt: true
   },
+
   {
     name: 'file-browser',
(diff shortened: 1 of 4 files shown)


dmitrizagidulin:
Tagging @dan-f or @nicola or @deiu for review.

dan-f:
Sorry! I forgot to "submit" my review comments.

PR #453, opened 2016-10-21
Remove unused n3 dependency

dmitrizagidulin:
Removing unused n3 dependency -- N3 parsing is taken care of by rdflib etc.


dmitrizagidulin:
Tagging @nicola / @deiu / @dan-f for review.

deiu:
We *could* drop it for now, but I have a feeling Tim wants to stream N3 patches through websockets soon.

dan-f:
:+1: for deleting a commented out file and unused dependency. If we want to add that n3 library in the future and use it, I don't see how this would at all get in the way of that.

dmitrizagidulin:
These updates are now part of PR https://github.com/solid/node-solid-server/pull/456. (Closing this one).

Issue #446, opened 2016-10-06
Return the Content-Type of resources in container list data


timbl:
It would be great if, along with the normal LDP container listing data, a solid server returned for each file (or directory) in the container, its MIME type as an RDF type.  
- It allows me to spot which folders should be thought of as slide shows without cheating and looking at the file extensions on the URIs
- It allows a file listing to use appropriate icons, colors, etc
- It allows code which runs recursively over a directory tree can be much faster without having to do a HEAD on each thing

(I suggest that Internet Content Types be mapped onto https: space, possibly https://www.w3.org/ns/mime/image/png and some support for it be made by adding data about the mime types derived from the MIME type XML file at IANA.  An alternative is to use a base URI in iana space which will not be supported with RDF)

dmitrizagidulin:
This is a great idea! Couple of questions:

1. What's the MIME type of a directory?
2. What's the predicate that should be used?
3. Is it important that the object be a URI? (What about returning MIME types as literals?)

timbl:
1. Just return the RDF type of an LDP container
2. rdf:type
3. Yes  I think so ...

The reason that we use an actual RDF class is because we can then use the RDFs tools.   We can add for example the in the ontology that image/png is a rdfs:subClassOf dc:Image 
The type arc is how we say that something is in a category in RDF traditionally.
The UI stack and the solid core already have a type index, so once could immediately use that to put a particular type of file in, like a spreadsheet or a PDF file, and we can use the app/type registry for a user to select how she will deal with a given mime type.

dmitrizagidulin:
Got it, makes sense.

dmitrizagidulin:
Wait, re number 1, isn't that mixing RDF Types and MIME Types? To put it another way, 'LDPResource' isn't a mime type..

timbl:
Because we use rdf:type, and mapping the mime types into RDF types, then one benefit we have is we mix them.

dmitrizagidulin:
Ah, makes sense. :)

timbl:
Ok, so went out on a limb in https://github.com/solid/node-solid-server/pull/456 and decided on an arbitrary mapping for that to map Media Type

image/jpeg

to the class

http://www.w3.org/ns/iana/media-types/image/jpeg#Resource

```
@prefix terms: <http://purl.org/dc/terms/>.
@prefix XML: <http://www.w3.org/2001/XMLSchema#>.
@prefix ldp: <http://www.w3.org/ns/ldp#>.
[...]

PR #456, opened 2016-11-27
In a directory listing, adds the media type class when we know it 

timbl:
In a directory listing, adds the media type class when we know it  for a directory member.   Add sdct:modified with an RDF date, which IMHO should deprocate the mtime returned as a unixinteger. (The RDF way is when know the data type and RDF has it, you use it. Note the code is actually shorter, internally node uses s Date() object)

Addresses Issue https://github.com/solid/node-solid-server/issues/446


dmitrizagidulin:
@timbl I fixed the jsStandard formatting errors that were failing the test and pushed to this PR. Now there's some tests failing about container serialization; investigating.

timbl:
@dmitrizagidulin it is exactly th container list test  which I would expect to fail, as that is what has changed.  I need to understand the proper way of running the tests ....

dmitrizagidulin:
@timbl the easiest way to run the tests is to just invoke `npm test` on the command line. (That also runs the Standard linter).

test/ldp.js
```diff
@@ -116,7 +116,7 @@ describe('LDP', function () {
         '   dcterms:title "This is a magic type" ;' +
         '   o:limit 500000.00 .', 'sampleContainer/magicType.ttl')
 
-      ldp.listContainer(path.join(__dirname, '/resources/sampleContainer/'), 'https://server.tld', '', 'text/turtle', function (err, data) {
+      ldp.listContainer(path.join(__dirname, '/resources/sampleContainer/'), 'https://server.tld/resources/sampleContainer/', 'https://server.tld', '', 'text/turtle', function (err, data) {
```
nicola: Isn't this the equivalent of just passing "/resources/sampleContainer/" ?

bshambaugh:
I just figured out you're using Jasmine in ldp.js :) https://jasmine.github.io/2.0/introduction.html

PR #466, opened 2016-12-05
Allow video skip with Accept-Ranges closes #457 #462

### LINKED ISSUE #457: Allow video skip with Accept-Ranges: bytes ?
I notice that an HTML5 video clip hosted from this server does not allow skipping forwards.    (Whereas Youtub.com for example does).   Is the problem just that the server does not do the
```
Accept-Ranges: bytes
```
protocol?  

Could that be added from an existing express module?

It could be done just for video, to avoid interaction with things happen with turtle, etc, where byte ranges are not needed.

Of course solid' main goal is serving data, but serving everything in a person's life too is important

melvincarvalho:
This PR allows video to be skipped using byte ranges.  

It also refactors ldp.get to use options rather than positional variables in order to provide more flexibility.

Separation of concerns between ldp.js and handers/get.js is maintained

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -31,7 +31,10 @@ function allow (mode) {
       var reqPath = res && res.locals && res.locals.path
         ? res.locals.path
         : req.path
-      ldp.exists(req.hostname, reqPath, (err, stat) => {
+      ldp.exists(req.hostname, reqPath, (err, ret) => {
+        if (ret) {
+          var stat = ret.stream
+        }
         if (!reqPath.endsWith('/') && !err && stat.isDirectory()) {
           reqPath += '/'
         }
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -47,7 +47,15 @@ function handler (req, res, next) {
 
   debug(req.originalUrl + ' on ' + req.hostname)
 
-  ldp.get(req.hostname, path, baseUri, includeBody, possibleRDFType, function (err, stream, contentType, container) {
+  var options = {
+    'hostname': req.hostname,
+    'path': path,
+    'baseUri': baseUri,
+    'includeBody': includeBody,
+    'possibleRDFType': possibleRDFType,
+    'range': req.headers.range
+  }
+  ldp.get(options, function (err, ret) {
     // use globHandler if magic is detected
     if (err && err.status === 404 && glob.hasMagic(path)) {
       debug('forwarding to glob request')
@@ -58,6 +66,14 @@ function handler (req, res, next) {
     if (err) {
       debug(req.method + ' -- Error: ' + err.status + ' ' + err.message)
       return next(err)
+    }
+
+    if (ret) {
+      var stream = ret.stream
+      var contentType = ret.contentType
+      var container = ret.container
+      var contentRange = ret.contentRange
+      var chunksize = ret.chunksize
     }
 
     // Till here it must exist
@@ -91,7 +107,13 @@ function handler (req, res, next) {
     if (negotiator.mediaType([contentType])) {
       debug('no translation ' + contentType)
       res.setHeader('Content-Type', contentType)
-      return stream.pipe(res)
+      if (contentRange) {
+        var headers = { 'Content-Range': contentRange, 'Accept-Ranges': 'bytes', 'Content-Length': chunksize }
+        res.writeHead(206, headers)
+        return stream.pipe(res)
+      } else {
+        return stream.pipe(res)
+      }
     }
 
     // If it is not in our RDFs we can't even translate,
--- a/lib/handlers/post.js
+++ b/lib/handlers/post.js
@@ -25,9 +25,13 @@ function handler (req, res, next) {
   }
 
   // Check if container exists
-  ldp.exists(req.hostname, containerPath, function (err, stats) {
+  ldp.exists(req.hostname, containerPath, function (err, ret) {
     if (err) {
       return next(error(err, 'Container not valid'))
+    }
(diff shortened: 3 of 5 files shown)


lib/handlers/allow.js
```diff
@@ -31,7 +31,10 @@ function allow (mode) {
       var reqPath = res && res.locals && res.locals.path
         ? res.locals.path
         : req.path
-      ldp.exists(req.hostname, reqPath, (err, stat) => {
+      ldp.exists(req.hostname, reqPath, (err, ret) => {
```
dmitrizagidulin: What does `ret` stand for?

melvincarvalho:
@dmitrizagidulin  ret == return value as an object

Reason is that get has to pass back a number of items, such as the range, the stream, etc.

Changing it into an object gives more flexibility.

timbl:
Looks sensible to me

timbl:
Ok, I have run it on a test machine , seems to work nicely!     Video can be skipped backward amd forward and started at any point.  
Merged in.
Thank you @melvincarvalho !

lib/handlers/allow.js
```diff
@@ -31,7 +31,10 @@ function allow (mode) {
...
+        if (ret) {
+          var stat = ret.stream
```
nicola: Wait this doesn't sound right

lib/handlers/get.js
```diff
@@ -60,6 +68,14 @@ function handler (req, res, next) {
       return next(err)
     }
 
+    if (ret) {
+      var stream = ret.stream
+      var contentType = ret.contentType
+      var container = ret.container
+      var contentRange = ret.contentRange
+      var chunksize = ret.chunksize
```
nicola: Why do you create variables inside a block scope?

lib/handlers/post.js
```diff
@@ -25,11 +25,15 @@ function handler (req, res, next) {
...
 
   // Check if container exists
-  ldp.exists(req.hostname, containerPath, function (err, stats) {
+  ldp.exists(req.hostname, containerPath, function (err, ret) {
     if (err) {
       return next(error(err, 'Container not valid'))
     }
 
+    if (ret) {
+      var stats = ret.stream
```
nicola: Same here

lib/identity-provider.js
```diff
@@ -503,14 +503,24 @@ IdentityProvider.prototype.getGraph = function (uri, callback) {
...
+  self.store.get(options,
+    function (err, ret) {
       if (err) {
         debug('Cannot find WebID card')
         var notexists = new Error('Cannot find WebID card')
         notexists.status = 500
         return callback(notexists)
       }
+      if (ret) {
+        var stream = ret.stream
```
nicola: Same here

lib/ldp.js
```diff
@@ -301,11 +320,24 @@ class LDP {
...
+        var stream
+        if (range) {
+          var total = fs.statSync(filename).size
+          var parts = range.replace(/bytes=/, '').split('-')
+          var partialstart = parts[0]
+          var partialend = parts[1]
+          var start = parseInt(partialstart, 10)
+          var end = partialend ? parseInt(partialend, 10) : total - 1
+          var chunksize = (end - start) + 1
+          var contentRange = 'bytes ' + start + '-' + end + '/' + total
```
nicola: Same here

lib/ldp.js
```diff
@@ -269,7 +280,15 @@ class LDP {
     ], callback)
   }
 
-  get (host, reqPath, baseUri, includeBody, contentType, callback) {
+  get (options, callback) {
+    if (options) {
```
nicola: What if no options are passed?

melvincarvalho:
Hi @nicola thanks for the review!

I asked about how to do this on gitter and in the issues since September.  Since I didnt get any feedback I tried to make a patch based on my understanding of the code and what I have learnt in nodejs.

I'd be happy to refactor based on best practice recommendations.

Regarding variables in block scope, is that against our coding standards?  My normal coding principle is to declare variables close to where they are used.  As I understand it they are hoisted to the top of the function in JS.  I could do it another way.

If no options are passed it is the same as sending in undefined in the positional parameters, I think?

Re: `var stat = ret.stream` not sounding right, I agree!  But that's how the code was, overloading a positional parameter for multiple purpose.  With my change it is now easier to refactor.

Advice welcome!

nicola:
Hey @melvincarvalho of course your PR is super welcome, it really great to make a PR instead of just discussing changes! (and you even gave a test passing working PR, which is amazing!)

However, I am not sure it was ready to merge yet. The only parts that I don't find ready are the declarations of variables in a scope, they should be declared in the same scope as the others, which are mainly all the changes that I am pointing out.

Let me know if that makes sense or we can find some time to do it together

melvincarvalho:
@nicola thanks again for the kind feedback! :) 

What I understand of JavaScript is that there is not block scoped variables with the keyword `var`.  New block scoped variables will come in with the keyword `let`.  So as far as the languages cares, whether the var is in the block, or at the top of the closures makes no difference.  If you prefer it at the top of the closure, I'll be happy to make that change.  

When returning from `get` with byte ranges, there are more parameters that must be passed back to the calling function such as the chunksize and the accept range.  This was more difficult to return as positional parameters in a callback, so I turned it into an object.  

There is some inconsistency as sometimes the callback returns the stat, and sometimes the stream, and different calling functions use it differently.  That's why the names look a bit strange, but that could be improved with a refactor.  However, this was the original logic of the current code.

We can discuss this further here or in gitter, if that suits you.

PR #467, opened 2016-12-06
Remove rsvp lib, use native promises only

dmitrizagidulin:


dmitrizagidulin:
Tagging @dan-f / @nicola for review.

dan-f:
:+1:

Issue #422, opened 2016-07-29
Create directory on system without decoding encoded values in Slug 


csarven:
Perhaps the title can be better with less assumptions as to what may be happening. Here is the code that I've used:

``` js
var options = {};
var url = 'https://example.org/';
var contentType = 'text/turtle';
var links = '<http://www.w3.org/ns/ldp#Container>; rel="type"';
var slug = encodeURIComponent('http://csarven.ca/foo');
var data = '';

var http = new XMLHttpRequest();
http.open('POST', url);
http.setRequestHeader('Content-Type', contentType);
http.setRequestHeader('Link', links);
if (slug && slug.length > 0) {
    http.setRequestHeader('Slug', slug);
}
if (!options.noCredentials) {
    http.withCredentials = true;
}
http.onreadystatechange = function() {
    if (this.readyState == this.DONE) {
        console.log(this);
    }
};
http.send(data);
```

Response has `Location: /http%3A%2F%2Fcsarven.ca%2Ffoo`.

What's physically created on the system is directory `http:` with subdirectory `csarven.ca` with file `foo`

Expected directory: `http%3A%2F%2Fcsarven.ca%2Ffoo`

This should be analogous to `mkdir "http%3A%2F%2Fcsarven.ca%2Ffoo"`

csarven:
Oh there is more:

If I create directory `http%3A%2F%2Fcsarven.ca%2Ffoo` manually on the server, and then later POST to that directory, the POST goes through (i.e., it creates say resource `https://example.org/http%3A%2F%2Fcsarven.ca%2Ffoo/bar`. However, when I GET that container (`http://example.org/http%3A%2F%2Fcsarven.ca%2Ffoo`), it has `ldp:contains <http://example.org/bar>`. Expected: `ldp:contains <http://example.org/http%3A%2F%2Fcsarven.ca%2Ffoo/bar>`

dmitrizagidulin:
Thanks, Sarven, that's helpful.

dmitrizagidulin:
Fixed in PR #469, closing.

PR #469, opened 2016-12-07
Fix handling/url-encoding of container names

dmitrizagidulin:
Fixes issue #422.

Currently, if you `POST` to `/` with a slug of `http://csarven.ca/foo` (to create a directory with that name), it succeeds with a response `Location` header of `/http%3A%2F%2Fcsarven.ca%2Ffoo`.
But what is physically created on the system is directory `http:` with subdirectory `csarven.ca` with file `foo`. Even if you uri-encode the slug before `POST`ing, the same thing happens, because the current behavior is to url-decode the slug for some reason, on the server.

The fixed behavior is now:

```js
      let containerName = 'https://example.com/page'
      let expectedDirName = '/post-tests/https%3A%2F%2Fexample.com%2Fpage/'
      server.post('/post-tests/')
        .set('slug', containerName)
        .set('content-type', 'text/turtle')
        .set('link', '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"')
        .expect(201)
      // Now response.headers.location === expectedDirName, and that directory is created.
```
Similarly, retrieval of the created container works:

```js
      let containerUrl = '/post-tests/https%3A%2F%2Fexample.com%2Fpage/'
      server.get(containerUrl)
        .expect('content-type', /text\/turtle/)
        .expect(200, done)
```

### DIFF
--- a/test/http.js
+++ b/test/http.js
@@ -13,6 +13,7 @@ var ldpServer = ldnode.createServer({
   root: path.join(__dirname, '/resources')
 })
 var server = supertest(ldpServer)
+var assert = require('chai').assert
 
 /**
  * Creates a new test basic container via an LDP POST
@@ -534,7 +535,7 @@ describe('HTTP APIs', function () {
           if (err) return done(err)
           var stats = fs.statSync(path.join(__dirname, '/resources/post-tests/loans/'))
           if (!stats.isDirectory()) {
-            return done(new Error('Cannot read file just created'))
+            return done(new Error('Cannot read container just created'))
           }
           done()
         })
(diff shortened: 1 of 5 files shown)


dmitrizagidulin:
Tagging @dan-f / @nicola for review.

test/http.js
```diff
@@ -544,6 +545,36 @@ describe('HTTP APIs', function () {
...
+            let createdDir = fs.statSync(path.join(__dirname, 'resources', expectedDirName))
+            assert(createdDir.isDirectory(), 'Container should have been created')
+          } catch (err) {
+            return done(err)
+          }
+          done()
+        })
+    })
+
+    it('should be able to access new url-named container', (done) => {
```
dan-f: Do we not clean up after every single test? I would have expected these two tests to be the same test, and for the resource `/post-tests/https%3A%2F%2Fexample.com%2Fpage/` *not* to exist in the scope of other tests in this describe block.

dmitrizagidulin:
Heads up, @csarven - PR https://github.com/solid/node-solid-server/pull/471 is going to change the policy of URLs as container names, when it gets merged. New procedure: uri-encode the Slug header before creating your inbox container.

Issue #470, opened 2016-12-10
URL en/decoding of characters in file name missing


timbl:
When the LDP container lists files, it doesn't properly encode non-URL characters (Like @,  space, etc) using % encoding.

When it received a requests with a URL which HAS got that encoding, then it doens't decode it and typically 404s

dmitrizagidulin:
Should be fixed by PR #471, closing.

PR #471, opened 2016-12-10
Fix URI encoding in file listing and decoding to get file names

timbl:
Also improve debug a little: truncate massive directory listings in the debug, add filename not found when 404 not found.

Designed to address https://github.com/solid/node-solid-server/issues/470

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -33,7 +33,8 @@ function debrack (s) {
 }
 
 function uriToFilename (uri, base) {
-  var filename = path.join(base, uri)
+  var decoded = uri.split('/').map(decodeURIComponent).join('/')
+  var filename = path.join(base, decoded)
   // Make sure filename ends with '/'  if filename exists and is a directory.
   // TODO this sync operation can be avoided and can be left
   // to do, to other components, see `ldp.get`
--- a/test/http.js
+++ b/test/http.js
@@ -546,6 +546,29 @@ describe('HTTP APIs', function () {
         .expect(200, done)
     })
 
+    it('should create a container with a name hex decoded from the slug', (done) => {
+      let containerName = 'Film%4011'
+      let expectedDirName = '/post-tests/Film@11/'
+      server.post('/post-tests/')
+        .set('slug', containerName)
+        .set('content-type', 'text/turtle')
+        .set('link', '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"')
+        .expect(201)
+        .end((err, res) => {
+          if (err) return done(err)
+          try {
+            assert.equal(res.headers.location, expectedDirName,
+              'Uri container names should be encoded')
+            let createdDir = fs.statSync(path.join(__dirname, 'resources', expectedDirName))
+            assert(createdDir.isDirectory(), 'Container should have been created')
+          } catch (err) {
+            return done(err)
+          }
+          done()
+        })
+    })
+
+    /* No, URLs are NOT ex-encoded to make filenames -- the other way around.
     it('should create a container with a url name', (done) => {
       let containerName = 'https://example.com/page'
       let expectedDirName = '/post-tests/https%3A%2F%2Fexample.com%2Fpage/'
@@ -574,6 +597,7 @@ describe('HTTP APIs', function () {
         .expect('content-type', /text\/turtle/)
         .expect(200, done)
     })
+    */
 
     after(function () {
       // Clean up after POST API tests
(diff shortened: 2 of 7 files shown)


lib/utils.js
```diff
@@ -33,7 +33,8 @@ function debrack (s) {
 }
 
 function uriToFilename (uri, base) {
-  var filename = path.join(base, uri)
+  var decoded = uri.split('/').map(decodeURIComponent).join('/')
```
dmitrizagidulin: I think we're going to run into problems with this approach. What if the file (or container) was uploaded as already uri-encoded? Say, for example, it was uploaded as `hello%20world.ttl`? And exists on disk as `hello%20world.ttl`. If you then issue a GET on it and this method `uriToFilename()` is invoked, it will try and look for the filename `hello world.ttl` and not find it.

dmitrizagidulin: This is especially relevant given the recent PR #469, which url-encodes container names.

timbl: When a file is uploaded, any filename is URI-encoded to form the URI.   
If the file  "hello world.ttl" exists then the resource "hello%20world.ttl" exists already too.
If there is a file on disk called  "hello%20world.ttl" then there is corresponding LDP resource "hello%2520world.ttl"

timbl: Reversible mapping.   The file world is self-consistent.  The URL world is self-consistent.

dmitrizagidulin: In thinking about this, I realize we still have an open issue where we don't URI-encode file names on PUT. (Only container names). So it's still possible for people to try and create files with names that are illegal as far as the OS/filesystem is concerned.
If we fix this issue (and enforce URI-encoding of filenames on PUT and POST), then there is no need to uri-encode them on listing.

timbl: A goal of this project is to be able to export existing files system users have. This means hex encoding things like space which are not allowed in URLs.   A Windows or Mac Documents or Music folder should be be able to export into Solid in a useful way.

It is NOT a goal to be able to allow clients to put files with characters which are typically illegal in a file system, like say colon, slash, or vertical bar.   I think we could return an error when a PUT or  PATCH tries to do that (by  hex encoding them in a URL).  I don't see any value in allowing this characters.

dmitrizagidulin: > A goal of this project is to be able to export existing files system users have.

How would this export process work? The existing files would arrive on a Solid server via PUTs and POSTs, right? Which means their filenames would be safely encoded.

> It is NOT a goal to be able to allow clients to put files with characters which are typically illegal in a file system

What would you like to see as the behavior when users/developers inevitably try to PUT or POST those kind of file names (illegal in file system)?

timbl:
In Mac, not much is illegal  [] {}  | all turn up as Chrome when you print off a page as PDF uses the windows title as a filename with very little adaptation!

http://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

Should we be conservative to prevent badness down the line in file systems apps (like git, cross system backup, etc) or should web liberal so that in soemone'd complicated machup of different systems it does not mean that some in fact unnecessary constraint in node-solid-server end up breaking things which would otherwise work?

lib/utils.js
timbl: Maybe add   *  ?  \    
?

Or wait until the file system objects?

Example use case: I want to copy my entire Mac home folder into a solid server to sync it with other macs

dmitrizagidulin:
I strongly believe we should be conservative, as far as accepting file name inputs via PUT, POST and PATCH, and URI-encode them. (to prevent all sorts of downstream problems, both with filesystem and with URIs).

test/ldp.js
```diff
@@ -109,6 +109,7 @@ describe('LDP', function () {
     })
   })
   describe('listContainer', function () {
+    /*
```
dmitrizagidulin: Do you mind deleting the test, instead of commenting out?

test/http.js
```diff
@@ -546,6 +546,29 @@ describe('HTTP APIs', function () {
...
+            let createdDir = fs.statSync(path.join(__dirname, 'resources', expectedDirName))
+            assert(createdDir.isDirectory(), 'Container should have been created')
+          } catch (err) {
+            return done(err)
+          }
+          done()
+        })
+    })
+
+    /* No, URLs are NOT ex-encoded to make filenames -- the other way around.
```
dmitrizagidulin: This test should be changed to test for a 400 error response, per our conversation.

PR #474, opened 2016-12-22
fixes issue where requesting a different format of a jsonld resource crashed the server

ryankeener:
When setting the accept header to something other than `application/ld+json` for `.jsonld` resources the server would crash with the following error:

```
     Uncaught TypeError: callback is not a function
      at Object.tripleCallback [as _callback] (node_modules/rdflib/lib/parse.js:112:7)
      at Object.N3Parser._readInTopContext (node_modules/n3/lib/N3Parser.js:81:19)
      at node_modules/n3/lib/N3Parser.js:683:35
      at Object.N3Lexer._tokenizeToEnd (node_modules/n3/lib/N3Lexer.js:80:11)
      at Immediate._onImmediate (node_modules/n3/lib/N3Lexer.js:318:38)
```

Based on this comment it looks like the function signature may have changed somewhere along the way to be async?
https://github.com/linkeddata/rdflib.js/blame/master/src/parse.js#L17


dmitrizagidulin:
Thanks @ryankeener! Good catch on the callback.

Issue #473, opened 2016-12-18
on-behalf-of header crashes server


melvincarvalho:
When using WebID delegation, the on-behalf-of header is sent by the user agent.  Currently this will crash node solid server.

It would be better to throw an error message, and keep the server running, or fix the delegation code in allow.js 

https://github.com/solid/node-solid-server/blob/master/lib/handlers/allow.js

The root problem lies in the function : 

    function verifyDelegator (host, ldp, baseUri, delegator, delegatee, callback) {

Which takes 6 parameters.  It is being called 

    verifyDelegator(req.hostname, delegator, req.session.userId, function

With only 4 parameters.  Meaning that the callback is sent instead of the delegate URI.  And the string function on the callback will throw an exception.

@dmitrizagidulin any ideas on what the logic *should* be here?

dmitrizagidulin:
@melvincarvalho I'm not sure what the correct logic should be. Maybe @deiu can take a look?

dmitrizagidulin:
Taking a look at this.

dmitrizagidulin:
Fixed by PR #487. (WebID Delegation is currently not fully implemented in the server, but will be implemented by PR #475.) Closing.

PR #475, opened 2016-12-22
Fix verifyDelegator() logic in the allow handler

dmitrizagidulin:
Fixes issue #473.

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -30,26 +30,30 @@ function allow (mode) {
       strictOrigin: ldp.strictOrigin
     })
 
-    getUserId(req, fetchDocument, (err, userId) => {
-      if (err) return next(err)
-
-      var reqPath = res && res.locals && res.locals.path
-        ? res.locals.path
-        : req.path
-      ldp.exists(req.hostname, reqPath, (err, ret) => {
-        if (ret) {
-          var stat = ret.stream
-        }
-        if (!reqPath.endsWith('/') && !err && stat.isDirectory()) {
-          reqPath += '/'
-        }
-        var options = {
-          origin: req.get('origin'),
-          host: req.protocol + '://' + req.get('host')
-        }
-        return acl.can(userId, mode, baseUri + reqPath, next, options)
+    let webId = req.session.userId
+    let onBehalfOf = utils.debrack(req.get('On-Behalf-Of'))
+    getUserId(webId, onBehalfOf, fetchDocument)
+      .then(userId => {
+        let reqPath = res && res.locals && res.locals.path
+          ? res.locals.path
+          : req.path
+        ldp.exists(req.hostname, reqPath, (err, ret) => {
+          if (ret) {
+            var stat = ret.stream
+          }
+          if (!reqPath.endsWith('/') && !err && stat.isDirectory()) {
+            reqPath += '/'
+          }
+          var options = {
+            origin: req.get('origin'),
+            host: req.protocol + '://' + req.get('host')
+          }
+          return acl.can(userId, mode, baseUri + reqPath, next, options)
+        })
       })
-    })
+      .catch(err => {
+        next(err)
+      })
   }
 }
 
@@ -105,33 +109,32 @@ function documentFetcher (host, ldp, baseUri) {
  * - From `session.userId` (if no delegation header, or secretary unauthorized)
  *
  * @method getUserId
- * @param req
+ * @param webId {string} Current user, from session.userId
+ * @param onBehalfOf {string} Contents of the `On-Behalf-Of` delegation header
  * @param fetchDocument {Function} Result of `documentFetcher()` call
- * @param callback {Function}
- * @return {string} Callback with the user id (either the current user, or
+ * @return {Promise<string>} The user id (either the current user, or
  *   the user being delegated for).
  */
-function getUserId (req, fetchDocument, callback) {
-  let webId = req.session.userId
-  let principal = utils.debrack(req.get('On-Behalf-Of'))
+function getUserId (webId, onBehalfOf, fetchDocument) {
+  let principal = onBehalfOf
   if (!webId || !principal) {
-    return callback(null, webId)
+    return Promise.resolve(webId)
   }
   // WebID Delegation header present, verify secretary and return the principal
   let secretary = webId
   debug('Verifying delegator (secretary agent) ', secretary,
     ' representing ', principal)
 
-  verifyDelegator(fetchDocument, secretary, principal, (err, verified) => {
-    if (err) { return callback(err) }
-    if (verified) {
-      debug('Request User ID (via delegation): ' + principal)
-      return callback(null, principal)
-    } else {
-      debug('Secretary agent not authorized: ' + secretary)
-      return callback(null, secretary)
-    }
-  })
+  return verifyDelegator(fetchDocument, secretary, principal)
+    .then(verified => {
+      if (verified) {
+        debug('Request User ID (via delegation): ' + principal)
+        return principal
+      } else {
+        debug('Secretary agent not authorized: ' + secretary)
+        return secretary
+      }
+    })
 }
 
 /**
@@ -142,24 +145,26 @@ function getUserId (req, fetchDocument, callback) {
  *   principal.
  * @param principal {string} Web ID of the secretary's "owner". Authorizes
  *   the secretary to act on the principal's behalf.
- * @param callback {Function}
- * @return {Boolean} Callback with whether or not delegator is verified
+ * @throws {Error}
+ * @return {Promise<Boolean>} Whether or not delegator is verified
  */
 function verifyDelegator (fetchDocument, secretary, principal, callback) {
-  // Fetch the principal's WebID Profile document
-  fetchDocument(principal, function (err, profileGraph) {
-    if (err) {
-      debug("Error fetching the principal's profile: ", err)
-      err.status = 500
-      return callback(err)
-    }
-    let authorizedMatches = profileGraph
-      .match(
-        $rdf.namedNode(principal),
-        ns.acl('delegates'),
-        $rdf.namedNode(secretary)
-      )
-    let isAuthorized = authorizedMatches.length > 0
-    return callback(null, isAuthorized)
+  return new Promise((resolve, reject) => {
+    // Fetch the principal's WebID Profile document
+    fetchDocument(principal, (err, profileGraph) => {
+      if (err) {
+        debug("Error fetching the principal's profile: ", err)
+        err.status = 500
+        return reject(err)
+      }
+      let authorizedMatches = profileGraph
+        .match(
+          $rdf.namedNode(principal),
+          ns.acl('delegates'),
+          $rdf.namedNode(secretary)
+        )
+      let isAuthorized = authorizedMatches.length > 0
+      resolve(isAuthorized)
+    })
   })
 }
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -166,6 +166,9 @@ function getRequestingWebId (webId, onBehalfOf, fetchDocument) {
 }
 
 /**
+ * Verifies that a given principal has authorized the secretary to act on their
+ * behalf, in their WebID Profile.
+ *
  * @see https://www.w3.org/wiki/WebID/Delegation
  * @see https://github.com/seebi/WebID-Delegation-paper
  * @param secretary {string} Web ID of the delegator agent. Acts on behalf of the
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -1,4 +1,6 @@
 module.exports.allow = allow
+module.exports.getUserId = getUserId
+module.exports.verifyDelegator = verifyDelegator
 
 var ACL = require('../acl-checker')
 var $rdf = require('rdflib')
@@ -13,20 +15,22 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
 
 function allow (mode) {
   return function allowHandler (req, res, next) {
-    var ldp = req.app.locals.ldp
+    let ldp = req.app.locals.ldp
     if (!ldp.webid) {
       return next()
     }
-    var baseUri = utils.uriBase(req)
+    let baseUri = utils.uriBase(req)
+
+    let fetchDocument = documentFetcher(req.hostname, ldp, baseUri)
 
     var acl = new ACL({
       debug: debug,
-      fetch: documentFetcher(req.hostname, ldp, baseUri),
+      fetch: fetchDocument,
       suffix: ldp.suffixAcl,
       strictOrigin: ldp.strictOrigin
     })
 
-    getUserId(req, function (err, userId) {
+    getUserId(req, fetchDocument, (err, userId) => {
       if (err) return next(err)
 
       var reqPath = res && res.locals && res.locals.path
@@ -52,7 +56,7 @@ function allow (mode) {
 /**
  * Returns a fetch document handler used by the ACLChecker to fetch .acl
  * resources up the inheritance chain.
- * The `fetch(uri, callback)` results in the callback, with either:
+ * The `fetchDocument(uri, callback)` results in the callback, with either:
  *   - `callback(err, graph)` if any error is encountered, or
  *   - `callback(null, graph)` with the parsed RDF graph of the fetched resource
  * @method documentFetcher
@@ -102,11 +106,12 @@ function documentFetcher (host, ldp, baseUri) {
  *
  * @method getUserId
  * @param req
+ * @param fetchDocument {Function} Result of `documentFetcher()` call
  * @param callback {Function}
  * @return {string} Callback with the user id (either the current user, or
  *   the user being delegated for).
  */
-function getUserId (req, callback) {
+function getUserId (req, fetchDocument, callback) {
   let webId = req.session.userId
   let principal = utils.debrack(req.get('On-Behalf-Of'))
   if (!webId || !principal) {
@@ -114,9 +119,6 @@ function getUserId (req, callback) {
   }
   // WebID Delegation header present, verify secretary and return the principal
   let secretary = webId
-  let ldp = req.app.locals.ldp
-  let baseUri = utils.uriBase(req)
-  let fetchDocument = documentFetcher(req.hostname, ldp, baseUri)
   debug('Verifying delegator (secretary agent) ', secretary,
     ' representing ', principal)
 
@@ -135,7 +137,6 @@ function getUserId (req, callback) {
 /**
  * @see https://www.w3.org/wiki/WebID/Delegation
  * @see https://github.com/seebi/WebID-Delegation-paper
- *
  * @param fetchDocument {Function} Result of `documentFetcher()` call
  * @param secretary {string} Web ID of the delegator agent. Acts on behalf of the
  *   principal.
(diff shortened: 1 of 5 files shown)


melvincarvalho:
I can confirm that sending the on-behalf-of header no longer crashes the server

dmitrizagidulin:
@melvincarvalho Thanks for confirming!

dmitrizagidulin:
@dan-f / @nicola / @deiu  - ready for review.

csarven:
Confirmed.

lib/handlers/allow.js
```diff
@@ -12,59 +15,65 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
 
 function allow (mode) {
```
dan-f: Could you add a `JSDoc` comment for this function?

lib/handlers/allow.js
```diff
@@ -93,56 +102,69 @@ function fetchDocument (host, ldp, baseUri) {
...
-      .each(delegatorGraph.sym(delegator),
-        delegatorGraph.sym('http://www.w3.org/ns/auth/acl#delegates'),
-        undefined)
-
-    for (var delegateeIndex in delegatesStatements) {
-      var delegateeValue = delegatesStatements[delegateeIndex]
-      if (utils.debrack(delegateeValue.toString()) === delegatee) {
-        callback(null, true)
+/**
+ * @see https://www.w3.org/wiki/WebID/Delegation
```
dan-f: Awesome `JSDoc` - the `@see` block tag is new to me. I think this could still use a brief one-liner to the top of the comment to introduce the function.

lib/handlers/allow.js
```diff
@@ -93,56 +102,69 @@ function fetchDocument (host, ldp, baseUri) {
...
-    }
-    // TODO check if this should be false
-    return callback(null, false)
+      let authorizedMatches = profileGraph
+        .match(
+          rdf.namedNode(principal),
+          ns.acl('delegates'),
+          rdf.namedNode(secretary)
+        )
+      let isAuthorized = authorizedMatches.length > 0
```
dan-f: Great. I think you could squish these two statements into one using the `.any()`.

lib/handlers/allow.js
```diff
@@ -93,56 +102,69 @@ function fetchDocument (host, ldp, baseUri) {
...
+ * - From `session.userId` (if no delegation header, or secretary unauthorized)
+ *
+ * @method getUserId
+ * @param webId {string} Current user, from session.userId
+ * @param onBehalfOf {string} Contents of the `On-Behalf-Of` delegation header
+ * @param fetchDocument {Function} Result of `documentFetcher()` call
+ * @return {Promise<string>} The user id (either the current user, or
+ *   the user being delegated for).
+ */
+function getUserId (webId, onBehalfOf, fetchDocument) {
```
dan-f: I think this function could use a slightly more specific name. Something like `getRequestingWebId`?

Another thing is that I think the `webId` and `onBehalfOf` parameters are an implementation concern of this function rather than the caller. Could we just pass the request object instead?

lib/handlers/allow.js
```diff
@@ -93,56 +102,69 @@ function fetchDocument (host, ldp, baseUri) {
...
+  let principal = onBehalfOf
+  if (!webId || !principal) {
```
dan-f: I don't understand this logic. If there's no webId in the request, resolve to the webId? That seems like an error condition to me.

lib/utils.js
```diff
@@ -19,8 +19,15 @@ var S = require('string')
...
+/**
+ * Removes the `<` and `>` brackets around a string and returns it.
+ * Used by the `allow` handler in `verifyDelegator()` logic.
+ * @method debrack
+ * @param s {string}
+ * @return {string}
+ */
 function debrack (s) {
-  if (s.length < 2) {
+  if (!s || s.length < 2) {
```
dan-f: Not to be annoying, but this should have a test case :)

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+    rdf.namedNode(alice),
+    ns.acl('delegates'),
+    rdf.namedNode(agentWebId)
+  )
+const notVerifiedSecretaryGraph = rdf.graph()
+
+describe('allow() handler', () => {
+  describe('getUserId()', () => {
+    const emptyFetchDocument = () => {}
+    it('no webId present', done => {
```
dan-f: I have a strong preference for test cases to be named after the behavior that they're looking for rather than the data they accept. For example, rather than "no webId present" something like "should <do expected behavior> when no webId is present". This clarifies the intended logic for contributors, future developers, and us months down the line when we forget what the intended behavior was.

lib/handlers/allow.js
dan-f: Or does that to handle unauthenticated users? If we're trying to return a null value for when the request is unauthenticated I'd prefer to explicitly return `null`.

lib/handlers/allow.js
```diff
@@ -12,59 +15,65 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
...
+          }
+          var options = {
+            origin: req.get('origin'),
+            host: req.protocol + '://' + req.get('host')
+          }
+          return acl.can(userId, mode, baseUri + reqPath, next, options)
+        })
+      })
+      .catch(err => {
+        next(err)
```
dan-f: I'm not too familiar with express. Does this pass the `err` object to the next middleware?

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+const verifiedSecretaryGraph = rdf.graph()
+verifiedSecretaryGraph.add(
+    rdf.namedNode(alice),
+    ns.acl('delegates'),
+    rdf.namedNode(agentWebId)
+  )
+const notVerifiedSecretaryGraph = rdf.graph()
+
+describe('allow() handler', () => {
+  describe('getUserId()', () => {
```
dan-f: I expected there to be two separate (not nested) describe blocks - one for the `allow` function and one for the `getUserId()` function - since they are separate exported functions from the `allow` module.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+    const emptyFetchDocument = () => {}
+    it('no webId present', done => {
+      let webId  // no secretary
+      let onBehalfOf = alice  // principal
```
dan-f: Additionally, I expect the case in which the request has no webId but an on-behalf-of header to be a rejected promise, not for `getUserId` to resolve to `undefined`.

As an aside I'd generalize that I'd almost never expect a promise to resolve to `undefined`. Null or falsy values should be explicit.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      getUserId(webId, onBehalfOf, emptyFetchDocument)
+        .then(userId => {
+          assert.equal(userId, undefined,
+            'Should return undefined when no webId is present')
+          done()
```
dan-f: Rather than accept the `done` callback and imperatively call it, mocha lets you return a promise from your test function.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+        })
+    })
+    it('no onBehalfOf present', done => {
```
dan-f: Same comment regarding test case naming.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      let webId = alice  // principal
+      let onBehalfOf  // no delegation header
+      getUserId(webId, onBehalfOf, emptyFetchDocument)
+        .then(userId => {
+          assert.equal(webId, userId,
+            'Should return webId when no onBehalfOf is present')
+          done()
+        })
+    })
+    it('throws 500 error if fetchDocument errors', done => {
```
dan-f: I think its an anti-pattern to knowingly allow and test for 500 errors in a web server. If there's a case in which you know you cannot complete a request you've got a slew of semantic error codes to choose from.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      getUserId(webId, onBehalfOf, fetchDocument)
+        .then(userId => {
+          assert.equal(userId, agentWebId,
+            'Should return the webId (secretary id) if secretary is NOT verified')
+          done()
+        })
+        .catch(err => {
+          console.error(err)
+        })
+    })
```
dan-f: There needs to be a test case for an entirely unauthenticated request - no webId and no on-behalf-of header.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      let fetchDocument = (url, callback) => {
+        callback(null, verifiedSecretaryGraph)
+      }
+      getUserId(webId, onBehalfOf, fetchDocument)
+        .then(userId => {
+          assert.equal(userId, alice,
+            'Should return principal (onBehalfOf) if secretary is verified')
+          done()
+        })
+        .catch(err => {
```
dan-f: You should remove this `catch` block which will suppress errors from making it back to mocha in your test.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      let fetchDocument = (url, callback) => {
+        callback(null, notVerifiedSecretaryGraph)
+      }
+      getUserId(webId, onBehalfOf, fetchDocument)
+        .then(userId => {
+          assert.equal(userId, agentWebId,
+            'Should return the webId (secretary id) if secretary is NOT verified')
+          done()
+        })
+        .catch(err => {
```
dan-f: Remove this catch block.

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+          console.error(err)
+        })
+    })
+  })
+
+  describe('verifyDelegator()', () => {
+    it('should throw 500 error if fetchDocument errors', done => {
```
dan-f: Again, can we come up with a more semantic error code?

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      let fetchDocument = (url, callback) => {
+        callback(null, verifiedSecretaryGraph)
+      }
+      verifyDelegator(secretary, principal, fetchDocument)
+        .then(verified => {
+          assert.equal(verified, true,
+            'Should be true if profile authorizes the secretary')
+          done()
+        })
+        .catch(err => {
```
dan-f: remove this `catch` block

test/allow.js
```diff
@@ -0,0 +1,131 @@
...
+      let fetchDocument = (url, callback) => {
+        callback(null, notVerifiedSecretaryGraph)
+      }
+      verifyDelegator(secretary, principal, fetchDocument)
+        .then(verified => {
+          assert.equal(verified, false,
+            'Should be false if profile does not authorize the secretary')
+          done()
+        })
+        .catch(err => {
```
dan-f: remove this `catch` block

lib/handlers/allow.js
dmitrizagidulin: Yeah, `next()` is Express's generic error and flow control handler. If not handled by the next middleware layer, it ends up being displayed to the user in `error-pages.js` (which is usually the desired behavior).

lib/handlers/allow.js
dmitrizagidulin: Yeah, this is meant to handle un-authenticated users. Sure, we can return `null` explicitly, no prob.

dmitrizagidulin:
> The documentFetcher function isn't tested. Since it's basicly just a web client, rather than write a bunch of tests for it, couldn't we use the solid web client directly?

As much as I would *love* to refactor `documentFetcher()` to use solid-web-client (or something similarly reusable), that's a fairly involved task and out of scope for this fix/PR. It is a pain point for me, though, and I will refactor it soon (probably when integrating the ad hoc Groups permission support).

test/allow.js
dmitrizagidulin: I've heard that as well. But in testing, returning the promise and not calling `done()` only results in `Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.` :(

dan-f: You have to 1) remove the `done` parameter from the test function and 2) return the promise from the test function.

dmitrizagidulin: Ah I see, yeah, I was leaving in the `done` parameter to the test.

dmitrizagidulin:
PR feedback addressed in the latest commit.

- Added docstring to `allow(mode)` handler.
- Renamed `getUserId()` to `getRequestingWebId()`
- Returning `null` explicitly in `getRequestingWebId()`.
- Updated the test description strings
- Removed explicit testing for 500 errors
- Removed `.catch()` blocks in the tests.
- Added tests for utils/debrack()

> The `allow` handler itself isn't tested but should be. It would be a good use case for a http mocking library such as nock. Perhaps this is more of an integration level test?

It's being tested on an integration leve in `test/acl.js`.

> Are we running standard on the tests? I'd argue we should.

Yes, our `npm test` also invokes `standard` on the tests directory.

> The lack of newlines between test cases makes them a bit hard to read for me.

No prob, added more newlines.

> The `documentFetcher` function isn't tested. Since it's basicly just a web client, rather than write a bunch of tests for it, couldn't we use the solid web client directly?

Yeah, refactoring the `documentFetcher()` function is on my to-do list (it's used extensively in `acl-checker.js`, and tested indirectly in `test/acl.js`). Writing tests for it is out of scope for this PR.

lib/handlers/allow.js
```diff
@@ -10,61 +13,84 @@ var utils = require('../utils')
...
+    let webId = req.session.userId
+    let onBehalfOf = utils.debrack(req.get('On-Behalf-Of'))
+    getRequestingWebId(webId, onBehalfOf, fetchDocument)
+      .then(userId => {
+        let reqPath = res && res.locals && res.locals.path
+          ? res.locals.path
+          : req.path
+        ldp.exists(req.hostname, reqPath, (err, ret) => {
+          if (ret) {
+            var stat = ret.stream
```
nicola: I know it was there before, but, where is this used?

dmitrizagidulin: @nicola No idea.

test/allow.js
```diff
@@ -14,117 +14,94 @@ verifiedSecretaryGraph.add(
...
-      getUserId(webId, onBehalfOf, emptyFetchDocument)
+
+      return getRequestingWebId(webId, onBehalfOf, emptyFetchDocument)
         .then(userId => {
           assert.equal(webId, userId,
             'Should return webId when no onBehalfOf is present')
-          done()
-        })
-    })
-    it('throws 500 error if fetchDocument errors', done => {
```
dan-f: What happened to this test?

test/allow.js
```diff
@@ -14,117 +14,94 @@ verifiedSecretaryGraph.add(
...
-          done()
-        })
-        .catch(err => {
-          console.error(err)
         })
     })
   })
 
   describe('verifyDelegator()', () => {
-    it('should throw 500 error if fetchDocument errors', done => {
```
dan-f: Why remove these tests? My comment was about testing for a more semantic error status code, not removing the tests.

test/utils.js
```diff
@@ -31,4 +31,27 @@ describe('Utility functions', function () {
       assert.equal(utils.uriToFilename('quote%27', 'base/'), 'base/quote\'')
     })
   })
+
+  describe('debrack()', () => {
+    it('should return null if no string is passed', done => {
+      assert.equal(utils.debrack(), null)
+      done()
```
dan-f: You don't need to accept/call the `done` parameter in synchronous tests.

lib/handlers/allow.js
```diff
@@ -108,20 +125,31 @@ function documentFetcher (host, ldp, baseUri) {
...
  * @param webId {string} Current user, from session.userId
  * @param onBehalfOf {string} Contents of the `On-Behalf-Of` delegation header
  * @param fetchDocument {Function} Result of `documentFetcher()` call
  * @return {Promise<string>} The user id (either the current user, or
  *   the user being delegated for).
  */
-function getUserId (webId, onBehalfOf, fetchDocument) {
-  let principal = onBehalfOf
-  if (!webId || !principal) {
+function getRequestingWebId (webId, onBehalfOf, fetchDocument) {
```
dan-f: My comment still stands that I think this function would be cleaner if it accepted the request object rather than `webId` and `onBehalfOf`.

dan-f:
Looks like my last few comments got squished as "outdated". Not sure why, but I'd still like you to see them.

test/allow.js
dmitrizagidulin: I removed that test because the fetch document function callback is no longer throwing a 500. I realized that the callback had no way of knowing the semantics of why it failed, whether for a 4xx-type reason or a 5xx-type. 
It'll be the responsibility of `fetchDocument()`, when we refactor it, to provide a more semantic error code.

lib/handlers/allow.js
dmitrizagidulin: I disagree. That function has no need of the rest of the `req` object. And in fact, does not care which headers the Web ID or the secretary id live.

test/allow.js
dan-f: As noted below, I think that if the logic in this PR depends on an untested function, it is necessarily in scope for that function to be tested.

dmitrizagidulin:
> I disagree that testing documentFetcher() is "out of scope" for this PR. If fetching the resource fails, then the delegation verification doesn't actually accomplish anything at all.

I think you're losing sight of the big picture of this PR. This is a fairly minor fix for a *crashing bug* to solid-server. I did not create `documentFetcher()`. And I do not plan to test it in its current form, in this PR, since it (and several of the functions inside it) are going to be refactored shortly, in a separate branch.

lib/handlers/allow.js
dan-f: Some procedure has to know where that data comes from, and rather than have every single caller of this function reach into the request object and pull out those values, it seems more maintainable to keep that logic here.

test/allow.js
dmitrizagidulin: *Many* functions inside solid-server, on which this PR depends, are untested. Blocking a fix until the developer adds tests to all of them is unrealistic.

lib/handlers/allow.js
dmitrizagidulin: Again, I disagree. And there's only one caller of this function.

test/utils.js
dmitrizagidulin: Thanks. Removed.

dan-f:
> I think you're losing sight of the big picture of this PR. This is a fairly minor fix for a crashing bug to solid-server. I did not create documentFetcher(). And I do not plan to test it in its current form, in this PR, since it (and several of the functions inside it) are going to be refactored shortly, in a separate branch.

Again, I disagree. A patch is not a "fix" unless the patched system is fully tested for the the thing which it claims to "fix".

As one of the maintainers of this Solid implementation, you're free to merge whatever code you want. I'm not going to give the `:+1:` to an implementation that isn't fully tested. I think it's worth pointing out that I spent much of my last week fixing bugs in various solid libraries that did not have test coverage.

dmitrizagidulin:
> I think it's worth pointing out that I spent much of my last week fixing bugs in various solid libraries that did not have test coverage.

You say that as if this is some sort of exceptional circumstance, instead of our daily reality as developers. This is a rapidly moving, pre-alpha stage, experimental project. Most the the Solid libraries we use don't have test coverage. And while that should be addressed, as much as possible, as we go along, requiring a full stop until everything gets tested is not reasonable.

dmitrizagidulin:
I will refactor `documentFetcher()`, however, and add tests in this PR.

dmitrizagidulin:
After further investigation, refactoring and testing `documentFetcher()` requires nontrivial changes to the rest of the server. While these are being addressed, I pushed PR #487 to temporarily address the crashing issue.

RubenVerborgh:
We have a different implementation with #646.

melvincarvalho:
Please dont delete the dz_fix_delegation branch, there may be useful code there.

PR #481, opened 2017-01-06
Serve default favicon.ico

dmitrizagidulin:
Since browsers and user agents indiscriminately request /favicon.ico,
serving it (using npm package `serve-favicon`) avoids log spam and makes for a better user experience.


dmitrizagidulin:
@dan-f / @nicola / @deiu - pr ready for review.

timbl:
We should have leaned from the dataserver that nothing should go in the default directory which can’t be pointed to elsewhere.
BTW I think that the default behavior should be the web devs will expect — you have an existing one at the normal place and it uses it — only supplement if that is 404.  This PR currently breaks my favico.ico

Issue #483, opened 2017-01-09
Refactor LDP to use a modular store/backend


dmitrizagidulin:
Currently, the LDP class and the filesystem-based backend code are intertwined.
Separate the two, to enable running solid-server on alternate backends (s3, quad stores, etc).

RubenVerborgh:
Following up in #702.

PR #484, opened 2017-01-09
LDP refactor to use a backend/store

dmitrizagidulin:
Addresses issue #483.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -31,6 +31,8 @@ var corsSettings = cors({
 
 function createApp (argv = {}) {
   var ldp = new LDP(argv)
+  ldp.printDebugInfo()
+
   var app = express()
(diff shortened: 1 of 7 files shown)


RubenVerborgh:
Branches unfortunately diverged, we will need to rethink.

melvincarvalho:
This may be a relevant pattern

https://en.wikipedia.org/wiki/Data_access_object

I've used it before.  Not normally a fan of patterns, but found this to be quite robust.

RubenVerborgh:
@melvincarvalho It's probably just going to be different LDP implementations, i.e., an interface that takes `target`, `operation`, `body` (https://github.com/solid/solid-architecture/blob/master/server/request-flow.md#step-4-perform-the-modification).

PR #487, opened 2017-01-24
Disable crashing verifyDelegator() code in allow() handler

dmitrizagidulin:
Currently, making any request to node-solid-server that has an `On-Behalf-Of:` HTTP header crashes the server (issue #473).
To address current crashes in production, this PR disables `verifyDelegator()` and related code until it can be fixed in PR #475.


dmitrizagidulin:
cc @dan-f / @nicola for review of temp patch.

PR #492, opened 2017-01-30
Refactor IDP/user account management

dmitrizagidulin:
Refactors user account signup management code in solid-server.

- Adds 53 more unit tests for existing functionality
- Adds `favicon.ico` to each user's account folder (which they can override if they want)
- Allows server operators to customize which files and folders are initialized on new account creation (by editing `./config/account-template/`)


dmitrizagidulin:
Ready for review, cc @dan-f / @nicola / @deiu

test/account-creation.js
```diff
@@ -59,45 +57,36 @@ describe('Identity Provider', function () {
...
-            .send('spkac=' + spkac + '&webid=https%3A%2F%2Fnicola.localhost%3A3457%2Fprofile%2Fcard%23me')
-            .expect('Content-Type', /application\/x-x509-user-cert/)
-            .expect(200)
-            .end(done)
-        })
-    })
+        .send('username=nicola&spkac=' + spkac)
+        .expect('Content-Type', /application\/x-x509-user-cert/)
+        .expect(200, done)
+    }).timeout(20000)
```
dan-f: If test cases are running for 20 seconds, there's something wrong with the application

account-template-default/.acl
```diff
@@ -0,0 +1,23 @@
+# Root ACL resource for the user account
+@prefix acl: <http://www.w3.org/ns/auth/acl#>.
+
+<#owner>
+    a acl:Authorization;
+
+    acl:agent <{{webId}}> ;
+
+    # Optional owner email, to be used for account recovery:
+    {{{accountMailto}}}
```
dmitrizagidulin: Np, I'll change it to use `#if`.

account-template-default/profile/card
```diff
@@ -0,0 +1,26 @@
...
+    a foaf:PersonalProfileDocument ;
+    foaf:maker <#me> ;
+    foaf:primaryTopic <#me> .
+
+<#me>
+    a foaf:Person ;
+    a schema:Person ;
+
+    foaf:name "{{name}}" ;
+
```
dmitrizagidulin: How about we put it in the private preferences file? (I'm reluctant to place the email in the public profile without the user's explicit consent).

bin/lib/options.js
```diff
@@ -46,7 +54,7 @@ module.exports = [
       'WebID-TLS',
       'WebID-OpenID Connect'
     ],
-    prompt: true,
+    prompt: false,
```
dmitrizagidulin: Because at this point in the PR, there's only one authentication method. So there's no point in prompting for the choice.

lib/api/accounts/user-accounts.js
```diff
@@ -0,0 +1,143 @@
...
+ * Returns an Express middleware handler for checking if a particular account
+ * exists (used by Signup apps).
+ *
+ * @param accountManager {AccountManager}
+ *
+ * @return {Function}
+ */
+function checkAccountExists (accountManager) {
+  return (req, res, next) => {
+    let accountUri = req.hostname
```
dmitrizagidulin: Not sure what the question is. It's not checking the origin, it's checking the request's hostname. (This is what the existing logic does, as well.)

account-template-default/profile/card
```diff
@@ -0,0 +1,26 @@
...
+    solid:account </> ;  # link to the account uri
+    pim:storage </> ;    # root storage
+
+    solid:inbox </inbox/> ;
+    ldp:inbox </inbox/> ;
+
+    pim:preferencesFile </settings/prefs.ttl> ;  # private settings/preferences
+    solid:publicTypeIndex </settings/publicTypeIndex.ttl> ;
+    solid:privateTypeIndex </settings/privateTypeIndex.ttl> .
```
dmitrizagidulin: It's in the public profile since that slightly speeds up loading the full extended profile. It doesn't seem like sensitive information -- would you rather see it in the preferences file?

lib/api/accounts/user-accounts.js
```diff
@@ -0,0 +1,143 @@
...
+    accountManager.accountUriExists(accountUri)
+      .then(found => {
+        if (!found) {
+          debug(`Account ${accountUri} is available (for ${req.originalUrl})`)
+          return res.sendStatus(404)
+        }
+        debug(`Account ${accountUri} is not available (for ${req.originalUrl})`)
+        next()
+      })
```
dmitrizagidulin: Good catch, forgot a `.catch()` there.

lib/api/accounts/user-accounts.js
```diff
@@ -0,0 +1,143 @@
...
+    } catch (err) {
+      err.status = err.status || 400
+      return next(err)
+    }
+
+    return request.createAccount()
+      .catch(err => {
+        err.status = err.status || 400
+        next(err)
+      })
```
dmitrizagidulin: `next()` does not, in fact, get called here (the successful account creation operation ends up writing to the response object, and Express ends the middleware chain there automatically; adding a next() actually results in an error).

lib/api/index.js
```diff
@@ -1,4 +1,18 @@
+'use strict'
+
+/**
+ * Handles a request to a not implemented `/api/` endpoint.
+ * @param req
+ * @param res
+ * @param next
+ */
+function notImplemented (req, res, next) {
+  res.status(404).send('API not implemented')
```
dmitrizagidulin: I like the more explicit not implemented message, but I'm happy to take this one out.

lib/api/index.js
```diff
@@ -1,4 +1,18 @@
...
+}
+
 module.exports = {
-  accounts: require('./accounts'),
-  messages: require('./messages')
+  authn: require('./authn'),
```
dmitrizagidulin: I hear you.
Part of the reason it's a common pattern is to avoid having to explicitly `require()` every handler in the APIs folder, since you can just require the API object, and then use `API.*` in your route setups.
I'm not attached to it, though - we can take out the various existing occurrences of it.

lib/create-app.js
```diff
@@ -107,66 +82,88 @@ function createApp (argv = {}) {
...
-          idp.middleware(true)(req, res, next)
-        } else if (found && ldp.idp) {
-          idp.middleware(false)(req, res, next)
-        } else {
-          next()
-        }
-      })
-    }
+    // Account Management API (create account, new cert)
+    app.use('/', API.userAccounts.middleware(accountManager))
```
dmitrizagidulin: Sure.

lib/create-app.js
```diff
@@ -107,66 +82,88 @@ function createApp (argv = {}) {
...
-    app.use('/api/accounts', needsOverwrite)
-    app.post('/api/accounts/signin', bodyParser.urlencoded({ extended: false }), API.accounts.signin())
-    app.post('/api/accounts/signout', API.accounts.signout())
-    app.post('/api/messages', authentication, bodyParser.urlencoded({ extended: false }), API.messages.send())
+    // Authentication API (login/logout)
+    app.post('/api/accounts/signin', bodyParser, API.authn.signin())
+    app.post('/api/accounts/signout', API.authn.signout())
+
+    // Messaging API
+    app.post('/api/messages', authentication, bodyParser, API.messages.send())
```
dmitrizagidulin: I agree, I'm working on refactoring that in the next PR.

lib/ldp.js
```diff
@@ -1,5 +1,6 @@
 var mime = require('mime-types')
 var path = require('path')
+const url = require('url')
 var fs = require('fs')
 var $rdf = require('rdflib')
 var async = require('async')
```
dmitrizagidulin: I agree (and in fact, the `async` module is sort of outdated, so I'm removing it in the LDP refactor).

lib/ldp.js
```diff
@@ -209,6 +223,50 @@ class LDP {
...
+   */
+  putGraph (graph, uri, contentType = DEFAULT_CONTENT_TYPE) {
+    return new Promise((resolve, reject) => {
+      let parsedUri = url.parse(uri)
+      let hostname = parsedUri.hostname
+      let path = parsedUri.pathname
+
+      serialize(graph, uri, contentType, (error, content) => {
+        if (error) {
+          error.status = error.status || 400
```
dmitrizagidulin: Np, I'll take it out.

lib/models/account-manager.js
```diff
@@ -0,0 +1,394 @@
...
+   * @param [options.store] {LDP}
+   * @param [options.pathCard] {string}
+   * @param [options.suffixURI] {string}
+   * @param [options.accountTemplatePath] {string} Path to the account template
+   *   directory (will be used as a template for default containers, etc, when
+   *   creating new accounts).
+   */
+  constructor (options = {}) {
+    if (!options.host) {
+      throw TypeError('AccountManager requires a host instance')
```
dmitrizagidulin: Np.

test/account-creation.js
dmitrizagidulin: I wholly agree :) These tests were timing out for me, before I started the refactor (they were running about 10-12 seconds, on my laptop), which is way too long, but that's part of the reason I'm trying to replace this authentication method.

lib/api/accounts/user-accounts.js
dan-f: OK I think I was confusing the `Host` and `Origin` headers.

test/account-creation.js
dan-f: So the question is why are they taking 10 to 12 seconds?

account-template-default/profile/card
dan-f: Seems reasonable to me.

test/account-creation.js
dmitrizagidulin: I mean, the answer could be as simple as 'I have too many apps running on my laptop'. :)

But I suspect it's the cert generation and the individual creation of each account resource via rdflib. (Which, if that's true, this should be faster now, as of this PR)

bin/lib/options.js
dan-f: Ok. I am confused by the existence of the oidc option if it's not actually implemented.

account-template-default/profile/card
dan-f: I'm not exactly sure. I guess I was thinking that it's potentially leaking sensitive information, but an attacker would know where the private type index is located anyways due to the solid spec, so if we really wanted to be secretive about it we'd have to dynamically generate the location of that graph... so I think it can stay as is.

bin/lib/options.js
dmitrizagidulin: Agreed, I'll take it out.

lib/api/accounts/user-accounts.js
dan-f: That strikes me as a bit opaque for library code to meddle with the request/respones objects rather than the route handlers themselves.

lib/api/index.js
dan-f: Like I said I can understand it if the content type is application/json but otherwise it seems redundant to me.

lib/api/index.js
dan-f: In this case I can see that as being reasonable. I think it would be nice if the namespacing in the exports object reflect the directory structure of the API modules themselves in that case to avoid confusion.

lib/api/accounts/user-accounts.js
dmitrizagidulin: It's not really library code though - the CreateAccountRequest class is specifically in charge of handler logic, dealing with the request & response object, etc. It's in an explicit class (as opposed to being in one long monolithic handler function) specifically for better testability.

lib/create-app.js
dan-f: I see. This would also remove the need for that API index module to re-export all those other api modules.

test/account-creation.js
dan-f: > I mean, the answer could be as simple as 'I have too many apps running on my laptop'. :)

I'm not so sure that a teeny tiny single-threaded web server concerned with teeny tiny file IO would be slowed down too much by other apps on the system... :wink:

> But I suspect it's the cert generation...

I didn't realize the server did any cert generation. I thought the client was responsible for that.

> ...and the individual creation of each account resource via rdflib. (Which, if that's true, this should be faster now, as of this PR)

Right, because this server now does that work when it's first booted up.

How about removing the timeouts to verify your claim?

dmitrizagidulin: > How about removing the timeouts to verify your claim?

Hehe, sure :)

lib/api/accounts/user-accounts.js
dan-f: One thing that makes it opaque to me: the fact that the `CreateAccountRequest` object swallows up the request/response objects from express rather than returning back to the express handler and deferring to it to choose how to respond to the client.

The other thing contributing to the opaqueness here IMO is the several layers of classes that interact to get the job done. The account creation handler instantiates a class called `CreateAccountRequest` which has a `createAccount` method, and then calls are made to the `AccountManager`. This is somewhat a matter of taste, and anyways you've already built and tested this particular approach anyways.

That said, I find code easier to understand when it is built as a collection of functions which operate on a small set of data rather than many classes each with their own ad-hoc interface.

dmitrizagidulin: > when it is built as a collection of functions which operate on a small set of data rather than many classes each with their own ad-hoc interface.

I have tried, as much as possible, to break up the logic of these requests into methods that each operate on a small set of data. It's very likely that I may have missed places, though; feel free to let me know if any large or unwieldy pieces jump out.

dmitrizagidulin:
@dan-f thanks for reviewing this!
I think I've addressed all the issues you mentioned in these last commits; let me know if there's anything else.

lib/api/accounts/user-accounts.js
dan-f: Like I said, it's somewhat a matter of taste. If it works, it works. But I find imperatively handing off the response object to be swallowed up by a series of of subroutines that implicitly handle the response to be opaque.

lib/api/accounts/user-accounts.js
```diff
@@ -0,0 +1,143 @@
...
+    accountManager.accountExists()
+      .then(found => {
+        if (!found) {
+          debug('(single user mode) Redirecting to account creation')
+
+          res.redirect(302, '/signup.html')
+        } else {
+          next()
+        }
+      })
```
dan-f: Needs a `.catch()`

lib/api/accounts/user-accounts.js
```diff
@@ -0,0 +1,143 @@
...
+  } else {
+    // In single user mode, if account has not yet been created, intercept
+    // all GET requests and redirect to the Signup form
+    accountManager.accountExists()
+      .then(found => {
+        if (!found) {
+          router.use('/signup.html', express.static('./static/signup.html'))
+          router.get('/*', firstTimeSignupRedirect(accountManager))
+        }
+      })
```
dan-f: Needs `.catch()`

lib/api/index.js
dan-f: Ping regarding content type of the 404 response. Would be nice if this didn't just trigger the default HTML not found page since this is for an API endpoint.

config/defaults.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
```
dmitrizagidulin: Ah, right -- I'm planning to refactor the cli script and its handling of the defaults in a separate PR.

lib/api/accounts/user-accounts.js
dmitrizagidulin: Added.

bin/lib/options.js
dmitrizagidulin: That's been taken out.

dmitrizagidulin:
@dan-f re the two 'ping's on `bin/lib/options.js` -- both of those have been addressed.

lib/api/accounts/user-accounts.js
dan-f: ...? Not seeing it

dmitrizagidulin: Just added in this last commit, `ccb99e6`

lib/api/accounts/user-accounts.js
dmitrizagidulin: I think the notion of 'swallowed up' is also a matter of taste. Each `*Request` class has a single method that writes to the response object. This makes it more isolatable and testable than if the outer route handler wrote to it.

lib/requests/add-cert-request.js
```diff
@@ -0,0 +1,138 @@
...
+   * @throws {Error} HTTP 400 if there were errors during certificate generation
+   *
+   * @returns {Promise}
+   */
+  static addCertificate (request) {
+    let { certificate, userAccount, accountManager } = request
+
+    return certificate.generateCertificate()
+      .catch(err => {
+        err.status = 400
```
dan-f: HTTP 4xx is for a client errors. Why would that apply to this scenario?

test/add-cert-request.js
```diff
@@ -0,0 +1,72 @@
...
+const AccountManager = require('../lib/models/account-manager')
+const AddCertificateRequest = require('../lib/requests/add-cert-request')
+
+var host
+
+beforeEach(() => {
+  host = SolidHost.from({ serverUri: 'https://example.com' })
+})
+
+describe('AddCertificateRequest', () => {
```
dan-f: None of the error conditions are tested

test/account-manager.js
```diff
@@ -0,0 +1,362 @@
...
+        multiUser: true,
+        host: SolidHost.from({ serverUri: 'https://localhost' })
+      }
+      let mgr = AccountManager.from(options)
+
+      let webId = mgr.accountUriFor('alice')
+      expect(webId).to.equal('https://alice.localhost')
+    })
+
+    it('should compose account uri for an account in single user mode', () => {
```
dan-f: In single user mode do accounts get names? If so what if the account name provided to the `accountUriFor` isn't correct? Pretty sure it currently just returns the host.

test/account-manager.js
```diff
@@ -0,0 +1,362 @@
...
+        .then(() => {
+          expect(accountManager.getProfileGraphFor).to
+            .have.been.calledWith(userAccount)
+        })
+    })
+
+    it('should add the cert key to the account graph', () => {
+      accountManager.getProfileGraphFor = sinon.stub()
+        .returns(Promise.resolve(profileGraph))
+      accountManager.addCertKeyToGraph = sinon.stub()
```
dan-f: If you stub out the actual adding of the cert key to the graph, then this test doesn't actually test what it claims to be testing.

test/account-manager.js
```diff
@@ -0,0 +1,362 @@
...
+
+const testAccountsDir = path.join(__dirname, 'resources', 'accounts')
+
+var host
+
+beforeEach(() => {
+  host = SolidHost.from({ serverUri: 'https://example.com' })
+})
+
+describe('AccountManager', () => {
```
dan-f: I'm not seeing the welcome email being tested

dmitrizagidulin: Out of scope for this PR; our whole email subsystem needs testing and refactoring.

test/add-cert-request.js
dmitrizagidulin: Which error conditions are you referring to?

test/account-manager.js
dmitrizagidulin: It's testing that the `addCertKeyToGraph()` is called, and with the specific parameters.

lib/requests/add-cert-request.js
dmitrizagidulin: Errors in cert generation result from invalid client input being sent, though.

test/account-manager.js
dmitrizagidulin: I think the idea here is that it doesn't matter (for the purposes of the account uri method) what the username is, in single-user mode.

test/account-manager.js
dmitrizagidulin: (per conversation w/ Dan) - Ok, will add some welcome email tests.

dmitrizagidulin: @dan-f Added EmailService tests, and welcome email specific tests.

test/account-manager.js
dmitrizagidulin: K, added a test for it. Thanks!

Issue #476, opened 2016-12-26
Missing base path in JSON-LD response


csarven:
`curl -iH"Accept: application/ld+json" https://example.org/container/` returns a JSON-LD body like:

```json
[
  {
    "@id": "https://example.org",
    "http://www.w3.org/ns/ldp#contains": [
      {
        "@id": "https://example.orgdc354280-cb7c-11e6-9c08-3d5cf446a570"
      }
    ]
  }
]
```

The expected response is:
```json
[
  {
    "@id": "https://example.org/container/",
    "http://www.w3.org/ns/ldp#contains": [
      {
        "@id": "https://example.org/container/dc354280-cb7c-11e6-9c08-3d5cf446a570"
      }
    ]
  }
]
```

dmitrizagidulin:
Wow, good catch

dmitrizagidulin:
This bug actually also affects all the other serializations - Turtle, NTriples, etc.

dmitrizagidulin:
Since the fix involves fairly extensive LDP changes, I will fix this as part of PR #484.

dmitrizagidulin:
Addressed by PR https://github.com/solid/node-solid-server/pull/496

PR #496, opened 2017-03-13
Fix base path for RDF parsing

dmitrizagidulin:
Fixes issue #476.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,9 +1,9 @@
 module.exports.uriToFilename = uriToFilename
 module.exports.uriToRelativeFilename = uriToRelativeFilename
 module.exports.filenameToBaseUri = filenameToBaseUri
-module.exports.uriAbs = uriAbs
+module.exports.getBaseUri = getBaseUri
 module.exports.pathBasename = pathBasename
-module.exports.uriBase = uriBase
+module.exports.getFullUri = getFullUri
 module.exports.hasSuffix = hasSuffix
 module.exports.getResourceLink = getResourceLink
 module.exports.parse = parse
@@ -62,7 +62,7 @@ function filenameToBaseUri (filename, uri, base) {
   return uri + '/' + uriPath
 }
 
-function uriAbs (req) {
+function getBaseUri (req) {
   return req.protocol + '://' + req.get('host')
 }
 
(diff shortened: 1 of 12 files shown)


dmitrizagidulin:
cc @csarven for testing -- let me know if this passes the LDN test.

csarven:
The URIs in the output looks good to me.

csarven:
Passes LDN tests as well! Note that it only tests JSON-LD.

However, I'm noticing that if a resource with JSON-LD uses `"@id": ""` (as the subject of triples), and then `'Accept: text/turtle'` is requested on that resource, the Turtle serialization doesn't return that triple. If the `"@id": "http://example.net/note#foo"`, it serializes to Turtle ok.

dmitrizagidulin:
@csarven - can you open an issue for that (the turtle serialization thing you mention above), with some more details (an example file so I can reproduce it, etc)?

dmitrizagidulin:
Is `"@id": ""` a valid JSON-LD statement?

csarven:
Created issue https://github.com/solid/node-solid-server/issues/497

csarven:
Yes, `"@id": ""` is valid JSON-LD. It is equivalent to `<>` in Turtle.

lib/utils.js
```diff
@@ -65,8 +66,22 @@ function uriAbs (req) {
...
+ *   uriBase(req)  // -> 'https://example.com/files/'
+ *   ```
+ *
+ * @param req {IncomingMessage}
+ *
+ * @return {string}
+ */
 function uriBase (req) {
-  return uriAbs(req) + (req.baseUrl || '')
+  return uriAbs(req) + url.resolve(req.baseUrl, req.path)
```
RubenVerborgh: Took me some figuring out, but it seems correct indeed.

I just wonder whether `uriBase` is the right name; something like `fullUri` seems to better convey the meaning.

RubenVerborgh: Similarly, `uriAbs` doesn't seem intuitive; I actually think this one should be called `uriBase`.

RubenVerborgh:
Had a go at the renaming in 8aea09ebd05e01af4dd2ab1c9dafb8ca32be5ec3.

dmitrizagidulin:
@RubenVerborgh nice! :) 👍

PR #501, opened 2017-05-16
enable HTTP/2

elf-pavlik:
it uses `spdy` module since `http2` doesn't work well with current express https://github.com/expressjs/express/issues/2364#issuecomment-260167854

### DIFF
--- a/lib/create-server.js
+++ b/lib/create-server.js
@@ -2,7 +2,7 @@ module.exports = createServer
 
 var express = require('express')
 var fs = require('fs')
-var https = require('https')
+var http2 = require('spdy')
 var http = require('http')
 var SolidWs = require('solid-ws')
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
Thanks, @elf-pavlik !

dmitrizagidulin:
My bad - per conversation with the Solid team, we would prefer not to merge PRs without tests. (In this case, testing that http2 or spdy integration works with solid-server). 
Will be rolling back the merge; please resubmit with tests.

elf-pavlik:
I'll try adding a test which uses https://www.npmjs.com/package/is-http2

RubenVerborgh:
Note that [native HTTP/2 support in Node.js](https://twitter.com/MylesBorins/status/893566926236024834) is coming.

Issue #491, opened 2017-01-26
Upgrade to latest rdflib 0.13.0


dmitrizagidulin:
Some change happened between 0.12 and 0.13 in rdflib (I think in PATCH handling) that is causing several tests to fail in `node-solid-server`. 
Fix whatever is needed in solid-server to pass the tests (so we can continue to keep upgrading rdflib as a dependency).

dmitrizagidulin:
Implemented in PR #503, closing.

PR #503, opened 2017-05-17
Upgrade to rdflib 0.15.0, update PATCH tests

dmitrizagidulin:
Addresses issue #491.
Fixes PATCH- and Glob-related failed tests when upgrading to rdflib 0.15.0

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -7,6 +7,7 @@ var $rdf = require('rdflib')
 var S = require('string')
 var async = require('async')
 var Negotiator = require('negotiator')
+const url = require('url')
 
 var debug = require('debug')('solid:get')
 var debugGlob = require('debug')('solid:glob')
@@ -132,6 +133,7 @@ function globHandler (req, res, next) {
   var root = !ldp.idp ? ldp.root : ldp.root + req.hostname + '/'
   var filename = utils.uriToFilename(req.path, root)
   var uri = utils.uriBase(req)
+  const requestUri = url.resolve(uri, req.path)
 
   var globOptions = {
     noext: true,
@@ -176,7 +178,7 @@ function globHandler (req, res, next) {
       var data = $rdf.serialize(
         undefined,
         globGraph,
-        filename,
+        requestUri,
         'text/turtle')
       // TODO this should be added as a middleware in the routes
       res.setHeader('Content-Type', 'text/turtle')
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -176,7 +176,7 @@ function globHandler (req, res, next) {
       var data = $rdf.serialize(
         undefined,
         globGraph,
-        null,
+        filename,
         'text/turtle')
       // TODO this should be added as a middleware in the routes
       res.setHeader('Content-Type', 'text/turtle')
--- a/test/patch-2.js
+++ b/test/patch-2.js
@@ -38,7 +38,31 @@ describe('PATCH', function () {
 
   describe('DELETE', function () {
     it('reproduce index 1 bug from pad', function (done) {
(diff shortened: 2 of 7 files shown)


dmitrizagidulin:
cc @timbl / @dan-f for review.

test/patch-2.js
```diff
@@ -109,10 +133,7 @@ describe('PATCH', function () {
         .end(function (err, res, body) {
           assert.equal(
             read('sampleContainer/prefixSparql.ttl'),
-            '@prefix schema: <http://schema.org/>.\n' +
-            '@prefix profile: <http://ogp.me/ns/profile#>.\n' +
-            '\n' +
-            '   <#> profile:first_name "Timothy"; a schema:Person .\n')
+            '@prefix : </prefixSparql.ttl#>.\n@prefix schema: <http://schema.org/>.\n@prefix pro: <http://ogp.me/ns/profile#>.\n\n: a schema:Person; pro:first_name "Timothy".\n\n')
```
dan-f: A bit confused about this change given the one above. Let's be consistent with formatting of data. I think the previous change is much more readable.

lib/handlers/get.js
```diff
@@ -176,7 +176,7 @@ function globHandler (req, res, next) {
       var data = $rdf.serialize(
         undefined,
         globGraph,
-        null,
+        filename,
         'text/turtle')
```
dmitrizagidulin: From what I understand, glob functionality is only spec'd/implemented for Turtle.

test/patch-2.js
dmitrizagidulin: No prob, will change it to match.

lib/handlers/get.js
dan-f: From what I understand, the notion of "globbing all the graphs in an LDP container into a single graph" is agnostic to a serialization format.

lib/handlers/get.js
```diff
@@ -176,7 +176,7 @@ function globHandler (req, res, next) {
       var data = $rdf.serialize(
         undefined,
         globGraph,
-        null,
+        filename,
```
dmitrizagidulin: An asterisk is a valid URI character. But whether that's the preferred semantics, as far as the base uri when using file patterns, is a tough call. We'll need to discuss it with @timbl etc.

lib/handlers/get.js
dmitrizagidulin: We can open an issue to add support for multiple content types for globbing, for solid-server. This PR is just intended to upgrade the server to the latest rdflib version.

lib/handlers/get.js
dan-f: My point was not regarding the asterisk. It was regarding the lack of a URI scheme

dmitrizagidulin: Ohh I see. Thanks! will fix

test/patch-2.js
dmitrizagidulin: Changed the expected value formatting to be multi-line. Thanks!

lib/handlers/get.js
dmitrizagidulin: I opened issue #505 to add support for content negotiation with globbing.

lib/handlers/get.js
dmitrizagidulin: Fixed the base serialization uri to be the request uri.

dmitrizagidulin:
> I fail to see how this "fixes Glob-related failed tests". I found several "glob-related tests", but none of them changed in this PR.

I just meant - this PR passes the failing tests that are introduced by upgrading to latest rdflib.

> The first glob-related tests are actually skipped. I tested, on master, un-skipping that describe block, and all of the glob tests failed. This is before the rdflib upgrade.

Yeah, that's weird, I'm not sure what the story was with those. I added an issue, #505, to un-skip and implement those tests.

> The second "glob test" doesn't test globbing. It simply tests for a 200 response code and content-type text/turtle.

Yeah, also odd. I expanded that test to actually check for result contents.

melvincarvalho:
A gentle bump on this.  

Hope it's OK, that I added the blocker label, as this fix will enable me to log in to node solid server, and therefore, work more effectively.

test/resources/sampleContainer/example1.ttl
```diff
@@ -2,6 +2,8 @@
 @prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix ex: <http://example.org/stuff/1.0/> .
 
+</sampleContainer/example*> a :glob.
```
dmitrizagidulin: Added to address @dan-f's comment; tests the `requestUri` being used as a serialization base for globs. (per @timbl's instructions)

dan-f: Uh... I would expect this triple to be *dynamically* added by the glob code, not *statically* added to the `sampleContainer/example1.ttl` graph.

dmitrizagidulin: This isn't an official designation or anything, just a triple @timbl recommended to test the serialization base in the output.

dan-f: I find it a really confusing example. Why not just add <#a> dc:title "foo" . to example1.ttl and then expect to see <example1.ttl#a> dc:title "foo" . in the glob?

I don't want to hold this up any longer

dmitrizagidulin: The example tests the serialization base being the glob request uri. The one you mention, `<#a> dc:title "foo" .` tests the parsing base (relative to example1.ttl). I'll add this line also.

dan-f: Huh. Sounds like I am missing some context from your discussion. I think you should merge if you and @timbl are happy with the test.

deiu: FYI, this is not how you test glob. `</sampleContainer/example*> a :glob.` being in a document makes no sense to me.

dmitrizagidulin: @deiu would something like `</sampleContainer/example*> :test "testing serialization base" .` make more sense, instead? 

I don't have a stake in this argument; was just adding the statement at @timbl's suggestion.

deiu: First, you don't have to keep this PR open on my account. Please merge if you think it's the right thing to do. I'm only pointing out that having a triple with the subject `/example*` in that  document makes **absolutely no sense**. The `*` character must be removed by the server before it starts to glob resources together, it should never be treated as a regular resource (which is what the test currently does).

dmitrizagidulin: @deiu I've replaced the `<example*> a :glob` statement with the example you & @dan-f suggested.

PR #506, opened 2017-05-26
Add 'id token from external servers' integration test

dmitrizagidulin:


dmitrizagidulin:
Adding a minimal integration test using an externally-issued id token on a different pod, using an http rest client. /cc @timbl @dan-f

test/integration/acl-oidc.js
```diff
@@ -305,6 +307,44 @@ describe('ACL HTTP', function () {
...
+    it('client should use an externally issued id token', () => {
+      let decodedToken = JWT.decode(idToken)
+      let tokenIssuer = decodedToken.payload.iss
+      let tokenWebId = decodedToken.payload.sub
+
+      assert.equal(tokenIssuer, 'https://example.com')
+      assert.equal(tokenWebId, 'https://alice.example.com/profile/card#me')
+    })
+
+    it('should recognize an external server id token', (done) => {
```
dan-f: This test isn't what I took away from our discussion. I'm seeing a unit test where the identity provider is faked out. I'd expect, e.g. `example.com`, to be an actual solid server just like the one behind `timAccountUri`. At least, that's what integration means to me in this context.

test/integration/acl-oidc.js
```diff
@@ -305,6 +307,44 @@ describe('ACL HTTP', function () {
...
+
+    const externalServer = path.join(__dirname,
+      '../resources/external-servers/example.com')
+    let testConfig = require(path.join(externalServer, 'openid-configuration.json'))
+    let testJwks = require(path.join(externalServer, 'jwks.json'))
+
+    nock('https://example.com').get('/jwks').reply(200, testJwks)
+    nock('https://example.com').get('/.well-known/openid-configuration').reply(200, testConfig)
+
+    it('client should use an externally issued id token', () => {
```
dan-f: Looks like this just verifies that the [anvilresearch/jose](https://github.com/anvilresearch/jose/) library knows how to parse a JWT. I don't see why this needs to be added.

dmitrizagidulin: No prob, will remove.

dmitrizagidulin:
> I was hoping to see tests against a system of a client, identity provider, and resource provider. Integration tests at that level could verify what the client receives depending on the various things that could succeed or go wrong during a handshake.

Agreed. This is a first step though, just to demonstrate that the ACL system will accept tokens from external servers presented by a REST client. I'll also be adding more elaborate integration tests like you describe (with 3 express servers interacting). Though if at all possible, I'd like to add those after we do our integration demo?

dan-f:
There's definitely value to a reasonable ratio of `unit : integration` tests. This one looks much more like a unit test. Maybe then put it with the rest of the unit tests?

dmitrizagidulin:
I see what you're saying. I put it here with the integration test, tho, because it still requires a running solid server, in addition to the one stubbed by nock.

dmitrizagidulin:
Closing. Superceded by PR #509.

PR #507, opened 2017-05-30
Fix 401 error handling, add tests

dmitrizagidulin:


dmitrizagidulin:
cc @dan-f for review

test/integration/errors-oidc.js
```diff
@@ -0,0 +1,112 @@
...
+    if (ldpHttpsServer) ldpHttpsServer.close()
+    fs.removeSync(path.join(rootPath, 'index.html'))
+    fs.removeSync(path.join(rootPath, 'index.html.acl'))
+  })
+
+  const server = supertest(serverUri)
+
+  describe('Unauthenticated requests to protected resources', () => {
+    describe('accepting json only', () => {
+      it('should return 401 Unauthorized with www-auth header', done => {
```
dan-f: FWIW if this is mocha, you can just `return` promises from the test rather than explicitly call `done`.

test/integration/errors-oidc.js
```diff
@@ -0,0 +1,112 @@
...
+          .expect('WWW-Authenticate', 'Bearer realm="https://localhost:3457", scope="openid"')
+          .expect(401, done)
+      })
+
+      it('should return an html redirect body', done => {
+        server.get('/profile/')
+          .set('Accept', 'text/html')
+          .expect('Content-Type', 'text/html; charset=utf-8')
+          .expect(res => {
+            expect(res.text).to.match(/<meta http-equiv="refresh"/)
```
dan-f: Where are we expected to be redirected to? I think that belongs with the test.

test/integration/errors-oidc.js
```diff
@@ -0,0 +1,112 @@
...
+      it('should return a 401 error', done => {
+        server.get('/profile/')
+          .set('Authorization', 'Bearer abcd123')
+          .expect('WWW-Authenticate', 'Bearer realm="https://localhost:3457", scope="openid", error="invalid_token", error_description="Access token is not a JWT"')
+          .expect(401, done)
+      })
+    })
+
+    describe('with an expired bearer token', () => {
+      const expiredToken = 'eyJhbGciOiJSUzI1NiIsImtpZCI6ImxOWk9CLURQRTFrIn0.eyJpc3MiOiJodHRwczovL2xvY2FsaG9zdDozNDU3Iiwic3ViIjoiaHR0cHM6Ly9sb2NhbGhvc3Q6MzQ1Ny9wcm9maWxlL2NhcmQjbWUiLCJhdWQiOiJodHRwczovL2xvY2FsaG9zdDozNDU3IiwiZXhwIjoxNDk2MjM5ODY1LCJpYXQiOjE0OTYyMzk4NjUsImp0aSI6IjliN2MwNGQyNDY3MjQ1ZWEiLCJub25jZSI6IklXaUpMVFNZUmktVklSSlhjejVGdU9CQTFZR1lZNjFnRGRlX2JnTEVPMDAiLCJhdF9oYXNoIjoiRFpES3I0RU1xTGE1Q0x1elV1WW9pdyJ9.uBTLy_wG5rr4kxM0hjXwIC-NwGYrGiiiY9IdOk5hEjLj2ECc767RU7iZ5vZa0pSrGy0V2Y3BiZ7lnYIA7N4YUAuS077g_4zavoFWyu9xeq6h70R8yfgFUNPo91PGpODC9hgiNbEv2dPBzTYYHqf7D6_-3HGnnDwiX7TjWLTkPLRvPLTcsCUl7G7y-EedjcVRk3Jyv8TNSoBMeTwOR3ewuzNostmCjUuLsr73YpVid6HE55BBqgSCDCNtS-I7nYmO_lRqIWJCydjdStSMJgxzSpASvoeCJ_lwZF6FXmZOQNNhmstw69fU85J1_QsS78cRa76-SnJJp6JCWHFBUAolPQ'
```
dan-f: I think this is too opaque. The code should explicitly generate an expired token for the purposes of a test like this.

lib/handlers/error-pages.js
```diff
@@ -9,12 +9,36 @@ function handler (err, req, res, next) {
 
   var ldp = req.app.locals.ldp
 
-  if (err.status === 401 &&
-      req.accepts('text/html') &&
-      ldp.auth === 'oidc') {
-    debug('401 error - redirect to Select Provider')
-    res.status(err.status)
-    redirectToLogin(req, res, next)
+  if (ldp.auth === 'oidc') {
```
dan-f: Is there a way of including this in the OIDC code rather than in the global error handler?

`WWW-Authenticate` is meant to be used for all 401s, so it should be here for WebID-TLS as well.

lib/handlers/error-pages.js
```diff
@@ -9,12 +9,36 @@ function handler (err, req, res, next) {
...
+    let statusCode = err.status || err.statusCode || 500
```
dan-f: I think we've talked about this before, but I can't think of an instance in which, in application code, a `500` error code should be explicitly set.

lib/handlers/error-pages.js
```diff
@@ -9,12 +9,36 @@ function handler (err, req, res, next) {
...
+      error_description: err.error_description,
+      error_uri: err.error_uri
+    }
+
+    let challengeParams = Object.keys(errorParams)
+      .filter(key => !!errorParams[key])
+      .map(key => `${key}="${errorParams[key]}"`)
+      .join(', ')
+
+    res.set('WWW-Authenticate', 'Bearer ' + challengeParams)
```
dan-f: This header is meant to be used with `401`s, but it's being used here regardless of the error code.

lib/handlers/error-pages.js
```diff
@@ -9,12 +9,36 @@ function handler (err, req, res, next) {
...
+
+    if (statusCode === 401 && req.accepts('text/html')) {
```
dan-f: And if status code is anything else? what does the `return` at the end of this function actually do?

lib/handlers/error-pages.js
dmitrizagidulin: Thanks. (forgot to move it.)

lib/handlers/error-pages.js
dmitrizagidulin: I remember. The reason I added it here is that this is the last place to catch it (if some other module forgot to specify a status code for their error), otherwise it gets a '0 is an invalid status code' error. I can take it out tho.

test/integration/errors-oidc.js
dmitrizagidulin: True - this version of `supertest` doesn't support promises I think, hence the explicit done(). I was planning on upgrading to the latest soon anyway, though.

lib/handlers/error-pages.js
dan-f: Well I guess 500 is better than 0, hah.

dmitrizagidulin:
@dan-f I rewrote the error-pages handler to reflect feedback, and added tests.

test/integration/errors-oidc.js
```diff
@@ -0,0 +1,98 @@
...
+        return server.get('/profile/')
+          .set('Accept', 'text/plain')
+          .expect('WWW-Authenticate', 'Bearer realm="https://localhost:3457", scope="openid"')
+          .expect(401)
+      })
+    })
+  })
+
+  describe('Authenticated responses to protected resources', () => {
+    describe('with an empty bearer token', () => {
```
dan-f: Is there anything in the error telling the client that they messed up the `Authorization` header? I'd expect a 400 in this case, since I think it could count as malformed syntax.

lib/handlers/error-pages.js
```diff
@@ -1,53 +1,167 @@
...
-      return res
-        .status(err.status)
-        .send(err.message || '')
-    }
-
-    res.status(err.status)
-    res.header('Content-Type', 'text/html')
-    res.send(text)
+      res.status(statusCode)
+      res.header('Content-Type', 'text/html')
```
dan-f: This should really obey the `Accept` header.

test/integration/errors-oidc.js
dmitrizagidulin: The Bearer Token spec dictates treating an empty bearer token as a plain 401 Unauthorized (with no error). And a malformed token as a 401 with a `invalid_token` error (which is tested below).

dan-f: Huh. According to [section 3.1](https://tools.ietf.org/html/rfc6750#section-3.1) of the spec, a missing bearer token seems to satisfy the `invalid_request`/`400` scenario:

>    invalid_request
         The request is missing a required parameter, includes an
         unsupported parameter or parameter value, repeats the same
         parameter, uses more than one method for including an access
         token, or is otherwise malformed.  The resource server SHOULD
         respond with the HTTP 400 (Bad Request) status code.

dmitrizagidulin: Not quite - the bearer token is not a required parameter.

lib/handlers/error-pages.js
dmitrizagidulin: That would be nice, agreed. I've opened issue https://github.com/solid/node-solid-server/issues/508 to implement support for more error formats (other than `text/html`) based on the Accept header.

test/integration/errors-oidc.js
dan-f: The bearer token in bearer token authentication is not a required parameter?

dmitrizagidulin: In the sense of - an un-authenticated request (without a token) is still a valid request. It just gets a 401 back (for a protected resource).

dan-f: An unauthenticated request would be one in which there was no attempt to provide credentials. Including an authorization header implies an attempt to authenticate in which case a missing token is an obvious missing parameter. I don't really feel like arguing this point further

dmitrizagidulin: Ok. Is this a blocking issue for you?

dmitrizagidulin: @dan-f Changed to return an `invalid_token` / `Empty access token` error instead.

dmitrizagidulin:
@dan-f any other issues on these 401 tests, or can I merge this?

test/integration/errors-oidc.js
dan-f: Still needs to be `400`, not `401`.

dmitrizagidulin: I completely disagree with your interpretation, but since this seems to be an important issue for you, I will change it to a 400.

dan-f: > invalid_request
The request is missing a required parameter, includes an
unsupported parameter or parameter value, repeats the same
parameter, uses more than one method for including an access
token, or is otherwise malformed. The resource server SHOULD
respond with the HTTP 400 (Bad Request) status code.

dmitrizagidulin: @dan-f now returning a 400 on an empty bearer token header.

PR #509, opened 2017-06-07
Add a 'two pods plus external web app' integration tests

dmitrizagidulin:


dmitrizagidulin:
This PR adds the 'two pods and an external web app' integration test to solid-server. cc @dan-f for review.

dmitrizagidulin:
Also adds support for Proof of Possession tokens to prevent token reuse by Resource Servers.

PR #510, opened 2017-06-22
Make ./data the default root folder

dmitrizagidulin:


dmitrizagidulin:
cc @RubenVerborgh for review

bin/lib/options.js
```diff
@@ -10,9 +10,9 @@ module.exports = [
   // },
   {
     name: 'root',
-    help: "Root folder to serve (defaut: './')",
+    help: "Root folder to serve (defaut: './data')",
```
RubenVerborgh: Maybe fix the `defaut` typo here as well?

dmitrizagidulin: Thanks! :) Will do

PR #511, opened 2017-06-23
Modularize PATCH logic

RubenVerborgh:
This pull request modularizes `PATCH` logic as follows:
- Code that is specific for the content type of a `PATCH` request body is moved to a separate file.
- Common code for all (RDF graph) patches is moved to the general patch logic.

This will enable a more flexible addition of new types of `PATCH` requests bodies, such as an N3 type that I'm planning.

This pull request also removes the logic for `application/sparql` patches, which were never fully implemented in the first place. `application/sparql-update` bodies remain supported.


nicola:
🎉 🎉

PR #513, opened 2017-06-29
Allow container creation via PUT

dmitrizagidulin:


dmitrizagidulin:
cc @dan-f @RubenVerborgh for review.

lib/ldp.js
```diff
@@ -266,11 +266,15 @@ class LDP {
     var root = !ldp.idp ? ldp.root : ldp.root + host + '/'
     var filePath = utils.uriToFilename(resourcePath, root, host)
 
-    // PUT requests not supported on containers. Use POST instead
-    if (filePath.endsWith('/')) {
-      return callback(error(409,
-        'PUT not supported on containers, use POST instead'))
+    const statusCode = fs.existsSync(filePath) ? 204 : 201
```
RubenVerborgh: Nitpick, but might want to have this async?

test/integration/acl-tls.js
```diff
@@ -161,20 +161,26 @@ describe('ACL HTTP', function () {
...
     })
     describe('with defaultForNew in parent path', function () {
-      before(function () {
+      before(() => {
         rm('/acl-tls/write-acl/empty-acl/another-empty-folder/test-file.acl')
         rm('/acl-tls/write-acl/empty-acl/test-folder/test-file')
         rm('/acl-tls/write-acl/empty-acl/test-file')
         rm('/acl-tls/write-acl/test-file')
         rm('/acl-tls/write-acl/test-file.acl')
+        rm('/acl-tls/write-acl/empty-acl/test-folder2/')
```
dan-f: You've got this in the `before` and `after`. Not really sure what the intention is here.

test/integration/http.js
```diff
@@ -393,15 +393,79 @@ describe('HTTP APIs', function () {
...
+        '../resources/sampleContainer/post2.ttl'),
+        { 'encoding': 'utf8' })
+
+      before(() => {
+        rm('/foo/')
+      })
+
+      after(() => {
+        rm('/foo/')
+      })
```
dan-f: Again, not sure why this is in both the `before` and `after`.

dmitrizagidulin:
> When the LDP container already exists, it looks like you re-generate the .meta file and return a 204. But based on the current implementation, the .meta file is overwritten and the rest of the contained data remains unchanged, due to how mkdirp works.

@dan-f Are those not the expected semantics? What would you like to see happen instead?

lib/ldp.js
dmitrizagidulin: If you prefer. I used a sync operation here because `exists` is a pretty lightweight operation, and it avoids introducing another level of callback nesting to the function. Happy to change it tho.

test/integration/http.js
```diff
@@ -393,15 +393,79 @@ describe('HTTP APIs', function () {
...
+
+      it('should create a container (implicit from url)', () => {
+        return server.put('/foo/two/')
+          .expect(201)
+          .then(() => {
+            let stats = fs.statSync(path.join(__dirname, '../resources/foo/two/'))
+
+            if (!stats.isDirectory()) {
```
dan-f: You can just write `expect(statsisDirectory()).toBe(true)` if you're using the `expect` library.

lib/ldp.js
RubenVerborgh: Yes, indeed… I don't mind either way. Can't wait to switch to Node 8 though, where sync and async have essentially the same syntax 😄

test/integration/http.js
```diff
@@ -393,15 +393,79 @@ describe('HTTP APIs', function () {
...
+        return server.put('/foo/four/')
+          .set('content-type', 'text/turtle')
+          .set('link', '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"')
+          .send(containerMeta)
+          .expect(201)
+          .then(() => {
+            let metaFilePath = path.join(__dirname, '../resources/foo/four/.meta')
+            let meta = fs.readFileSync(metaFilePath, 'utf8')
+
+            assert.equal(meta, containerMeta)
```
dan-f: I think @deiu has pointed this out before, but testing for graph equality based on serialization can be brittle. Better to use something like rdflib to assert that the graphs are equal.

test/integration/http.js
```diff
@@ -393,15 +393,79 @@ describe('HTTP APIs', function () {
...
+              .send(newMeta)
+              .expect(204)
+          })
+          .then(() => {
+            let metaFilePath = path.join(__dirname, '../resources/foo/five/.meta')
+            let meta = fs.readFileSync(metaFilePath, 'utf8')
+
+            assert.equal(meta, newMeta)
+          })
+      })
```
dan-f: Missing a test for the case in which there are already graphs in the container.

test/integration/ldp.js
```diff
@@ -93,6 +93,14 @@ describe('LDP', function () {
   })
 
   describe('putGraph', () => {
+    before(() => {
+      rm('sampleContainer/example1-copy.ttl')
+    })
+
+    after(() => {
+      rm('sampleContainer/example1-copy.ttl')
```
dan-f: ?

test/integration/ldp.js
```diff
@@ -126,10 +140,34 @@ describe('LDP', function () {
...
+        assert.equal(status, 201)
+
+        let written = read('new-container/.meta')
+        assert.equal(written, containerMeta)
+
+        done()
+      })
+    })
+
+    it('should update existing container meta', done => {
```
dan-f: I'd much prefer if we didn't share state between tests. Having a test rely on the output of a previous test introduces a possible race condition, especially since, depending on the test runner, tests may be run concurrently.

test/integration/http.js
dmitrizagidulin: I agree, for tests of graph equality. But there isn't any parsing or serialization taking place here. This is just straight file read/write operations.

lib/ldp.js
dan-f: Rule of thumb is async when available for code handling a request, sync for everything else.

test/integration/http.js
dmitrizagidulin: Thanks.

test/integration/http.js
deiu: Can I point out that hardcoding `.meta` is probably a bad idea. I think the config options contains a variable with the name for the meta document.

dan-f: That's an implementation detail of the fact that node-solid-server doesn't yet use a database. It's generally best for integration tests not to assume any particular implementation.

test/integration/ldp.js
RubenVerborgh: This is actually okay; within a single `describe` block, we can have multiple assertions (and in fact, it's in general better to split assertions as much as possible).

dan-f: I'm not objecting to multiple assertions - I'm objecting to the fact that this test (the code in the callback passed to `it`) relies on the fact that some other test created some data on the file system.

test/integration/http.js
dmitrizagidulin: It's in the before and after to ensure the directory is clear before the tests run; fairly standard procedure. Happy to take it out if you think it excessive.

dan-f:
> @dan-f Are those not the expected semantics? What would you like to see happen instead?

Well, I'm not exactly sure, as the [LDP spec](https://www.w3.org/TR/ldp/#ldpc-container) doesn't seem to specify what to do when the container already exists.

That said, I'd expect either a) to recreate the container (removing the existing container entirely) or not to modify it at all. The current implementation rewrites the `.meta` graph and nothing else, which seems a bit unintuitive.

test/integration/http.js
dan-f: I understand cleanup is necessary, but I've never had to do it both before and after a test is run.
Either should suffice.

dmitrizagidulin: Np, will take the before() out.

test/integration/ldp.js
RubenVerborgh: Ah yes, the scoping should be different. That's what I dislike about `super-agent`: it makes it hard to do that separation right (first create context, then assert).

test/integration/http.js
dan-f: Oh I see, these are before/after all the tests in the block, not each test individually. I think this intersects with the issue of shared state between tests.

dmitrizagidulin:
> Well, I'm not exactly sure, as the LDP spec doesn't seem to specify what to do when the container already exists.

> That said, I'd expect either a) to recreate the container (removing the existing container entirely) or not to modify it at all. The current implementation rewrites the .meta graph and nothing else, which seems a bit unintuitive.

Maybe @timbl can clarify Solid semantics of PUT to a container..? (with regards to the .meta etc)?

RubenVerborgh:
Would need a serious rebase.

PR #516, opened 2017-07-02
Add N3 support for PATCH

RubenVerborgh:
This pull request adds **support for N3 patches**, and contains **extensive tests for `PATCH`**.

### Added features
- exposing ACL object and user ID on the request
- fine-grained permission checking of patches after parsing (for both N3 and SPARQL UPDATE)
- support for Notation3 patches

### Pending decisions
- [x] A final vocabulary/vocabularies for N3 patches ([options on the Wiki](https://github.com/solid/node-solid-server/wiki/Existing-patch-approaches))
- [x] Have the right decisions been taken for status codes and permissions? (See _Specification_ below.)
- [x] The PATCH handler performs _partial_ permission checking: it assumes that _Append_ permissions are checked, and—after patch parsing—verifies _Read_ and _Write_ permissions when needed for a particular patch. We might want to either double-check _Append_ permissions, or move permission checking into handlers in general.

As future work, we can create an abstraction that allows non-triple-based patches as well. For this, however, I would strongly recommend rewriting handlers in an object-oriented way, as the current pure functional style could become confusing.

### Specification
<details>
<pre><code>
PATCH
  with a patch document
    with an unsupported content type
      ✓ returns HTTP status code 415
      ✓ has "Unsupported patch content type: text/other" in the response
      ✓ does not modify the file
    containing invalid syntax
      ✓ returns HTTP status code 400
      ✓ has "Patch document syntax error" in the response
      ✓ does not modify the file
    without relevant patch element
      ✓ returns HTTP status code 400
      ✓ has "No patch for https://tim.localhost:7777/read-write.ttl found" in the response
      ✓ does not modify the file
    with neither insert nor delete
      ✓ returns HTTP status code 400
      ✓ has "Patch should at least contain inserts or deletes" in the response
      ✓ does not modify the file
  with insert
    on a non-existing file
      ✓ returns HTTP status code 200
      ✓ has "Patch applied successfully" in the response
      ✓ creates the file
      ✓ writes the correct contents
    on a resource with read-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with append-only access
      ✓ returns HTTP status code 200
      ✓ has "Patch applied successfully" in the response
      ✓ patches the file correctly
    on a resource with write-only access
      ✓ returns HTTP status code 200
      ✓ has "Patch applied successfully" in the response
      ✓ patches the file correctly
  with insert and where
    on a non-existing file
      ✓ returns HTTP status code 409
      ✓ has "The patch could not be applied" in the response
      ✓ does not create the file
    on a resource with read-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with append-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with write-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with read-append access
      with a matching WHERE clause
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a non-matching WHERE clause
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
    on a resource with read-write access
      with a matching WHERE clause
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a non-matching WHERE clause
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
  with delete
    on a non-existing file
      ✓ returns HTTP status code 409
      ✓ has "The patch could not be applied" in the response
      ✓ does not create the file
    on a resource with read-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with append-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with write-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with read-append access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with read-write access
      with a patch for existing data
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a patch for non-existing data
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
      with a matching WHERE clause
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a non-matching WHERE clause
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
  deleting and inserting
    on a non-existing file
      ✓ returns HTTP status code 409
      ✓ has "The patch could not be applied" in the response
      ✓ does not create the file
    on a resource with read-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with append-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with write-only access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with read-append access
      ✓ returns HTTP status code 403
      ✓ has "Access denied" in the response
      ✓ does not modify the file
    on a resource with read-write access
      executes deletes before inserts
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
      with a patch for existing data
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a patch for non-existing data
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
      with a matching WHERE clause
        ✓ returns HTTP status code 200
        ✓ has "Patch applied successfully" in the response
        ✓ patches the file correctly
      with a non-matching WHERE clause
        ✓ returns HTTP status code 409
        ✓ has "The patch could not be applied" in the response
        ✓ does not modify the file
</code></pre>
</details>


lib/ldp-middleware.js
```diff
@@ -24,7 +24,7 @@ function LdpMiddleware (corsSettings) {
   router.copy('/*', acl.allow('Write'), copy)
   router.get('/*', index, acl.allow('Read'), get)
   router.post('/*', acl.allow('Append'), post)
-  router.patch('/*', acl.allow('Write'), patch)
+  router.patch('/*', acl.allow('Append'), patch)
```
RubenVerborgh: @dmitrizagidulin This is a permission downgrade that, with other patch handlers, could allow unintended access. The patch handler is assumed to check for _Write_ (and _Read_) permission when needed after parsing the patch.

(Note that this handler did not check for _Read_ yet.)

dmitrizagidulin: Sounds good; the code will check for Write vs Append downstream, as we discussed.

RubenVerborgh:
Mhm, the sync-ness of the patch handler seems to have been gone some time already. The file I started working on already performed asynchronous writes (https://github.com/solid/node-solid-server/pull/511/files#diff-d83413e73094e31dc0662b194162262dL187); it seems that the "sync" comment was copied verbatim from above (https://github.com/solid/node-solid-server/pull/511/files#diff-d83413e73094e31dc0662b194162262dL56), but that code never performed any actual writing (synchronous nor asynchronous).

In any case, sync is not an appropriate solution for the atomicity problem, given that any Node.js server can be started with multiple threads, which can interfere. If cross-request atomicity is important for Solid (which I assume it is), we might need to give locking a more prominent place in the overall architecture.

dmitrizagidulin:
@timbl I agree with @RubenVerborgh - switching the PATCH handler code to synchronous reads & writes won't do anything for atomicity; that has to be handled separately and explicitly (via locks / mutexes, etc).

lib/handlers/allow.js
```diff
@@ -27,9 +27,11 @@ function allow (mode) {
       suffix: ldp.suffixAcl,
       strictOrigin: ldp.strictOrigin
     })
+    req.acl = acl
 
     getUserId(req, function (err, userId) {
       if (err) return next(err)
+      req.userId = userId
```
dmitrizagidulin: I don't think you need this part; the userId is already recorded on the request object upstream of this handler.

RubenVerborgh: As we discussed, not in all cases (could be in `req.session.userId` or `locals.oidc.webIdFromClaims(req.claims)`). Opened #517 for this.

RubenVerborgh:
TODO for @dmitrizagidulin and/or @timbl:
- [ ] Check whether the decisions regarding status codes are okay (see Specification > Details at the top of this page).

TODO for @dmitrizagidulin:
- [ ] Check whether the assumptions regarding authentication are fine (see _partial permission checking_ argument above).

TODO For @RubenVerborgh:
- [ ] Create and publish patch vocabulary

dmitrizagidulin:
Should a PATCH with an insert to a non-existing file return a 201 Created instead of a 200?

RubenVerborgh:
@dmitrizagidulin Seems logical. Only problem might be that a user with only Append permissions (not Read) might not be authorized to know whether the file already existed.

dmitrizagidulin:
> Only problem might be that a user with only Append permissions (not Read) might not be authorized to know whether the file already existed.

Good point. Is that a use case we should support, tho? I can only think of one Append-but-not-Read access, and that's the Inbox. And there the main interaction is via POST, not PATCH.

dmitrizagidulin:
👍 on the status codes from me. (aside from the minor conversation about the 201 insert via patch)

RubenVerborgh:
Discussed this with @dmitrizagidulin and we decided to keep the 200.

RubenVerborgh:
Ready, pending approval of https://github.com/solid/vocab/pull/25.

RubenVerborgh:
Currently on hold because of #552.

timbl:
This functionality adds the ability to parse N3 patches.  Then, to send them upstream, we need to add an option to the pub/sub protocol to include patches in the upstream info in the websocket.

RubenVerborgh:
#552 has been merged; the `rv/patch/n3` branch has been merged with master, so this pull request is ready to be reviewed and merged.

RubenVerborgh:
@dmitrizagidulin The final commit 1aca23f8810c6e44657cfedc9f1205983c5287e0 is currently a merge commit of `dz_oidc` onto `rv/patch/n3` (resolving some conflicts). It might simplify history if I make the same merge, but then from ` rv/patch/n3` onto `dz_oidc` and then push (instead of merging on GitHub).

PR #518, opened 2017-07-06
Add and test Node 8 support

RubenVerborgh:
The server tests now run with Node 8, and this is tested on Travis.

The main problem was that setting `NODE_TLS_REJECT_UNAUTHORIZED` at runtime through `process.env` didn't work anymore. I have replaced this by the `rejectUnauthorized` configuration option where possible, which fixed the ACL TLS test (and can be useful for local testing). Unfortunately, the supertest library does not allow to set that option, so I set is a an environment variable before calling the Mocha tests.

The remaining problem is that Nock is currently incompatible with Node 8, so for now, I have just skipped those 4 tests that use Nock.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ sudo: false
 language: node_js
 node_js:
   - "6.0"
+  - "8.0"
+  - "node"
 env:
   - CXX=g++-4.8
 
--- a/lib/create-server.js
+++ b/lib/create-server.js
@@ -57,10 +57,10 @@ function createServer (argv, app) {
       throw new Error('Can\'t find SSL cert in ' + argv.sslCert)
     }
 
-    var credentials = {
+    var credentials = Object.assign({
       key: key,
       cert: cert
-    }
+    }, argv)
 
     if (ldp.webid && ldp.auth === 'tls') {
       credentials.requestCert = true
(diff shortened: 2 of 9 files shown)


dmitrizagidulin:
@RubenVerborgh Nice job tracking down these issues! I wonder if we should wait until Nock works with node 8 before merging this... (I don't suppose Travis allows you to execute different scripts for different node versions?)

RubenVerborgh:
I personally wouldn't wait for Nock; I find it very strange that they haven't tackled their issues already. It seems more important to me that we _do_ test for Node 8 (whereas Nock currently doesn't) to avoid breaking things.

It would be possible to have a switch statement per Node version. However, given that we know the proxy tests work on v6, and given that these tests only depend on the proxy and nothing else, I don't see a problem. If somebody wants to work on the proxy in the future, they will likely look at the proxy tests anyway.

RubenVerborgh:
Nock has been [updated](0d96dba9e6e6b541196e17d5739918e2b6), so I can update this PR.

RubenVerborgh:
@dmitrizagidulin Rebased to the latest `dz_oidc`.

dmitrizagidulin:
@RubenVerborgh ok, it's ready to go on your end?

RubenVerborgh:
Totally!

PR #519, opened 2017-07-13
Allow WebID+TLS behind a reverse proxy

RubenVerborgh:
This pull requests extends the Solid server with an `acceptCertificateHeader` option. If enabled (default: no), the Solid server accepts a client certificate in PEM format through the `X-SSL-Cert` HTTP header.

This can be used in combination with the following NGINX options:
```nginx
ssl_verify_client optional_no_ca;
proxy_set_header X-SSL-CERT $ssl_client_cert;
```
That way, NGINX can terminate the SSL connection (and thus run multiple services on port 443), whereas Solid can still check whether the certificate matches the WebID.

### Security considerations
In this configuration, the Solid server does _not_ verify whether the client has the matching private key to the certificate (NGINX does). As such, `acceptCertificateHeader` should **never** be enabled on front-facing servers.


dmitrizagidulin:
@RubenVerborgh looks good! Thanks.

PR #525, opened 2017-07-25
Only set User header with WebID-TLS

### LINKED ISSUE #523: Do not return User header for non-WebID TLS auth schemes

RubenVerborgh:
Closes #523. Breaking change, needs new semver-major.

### DIFF
--- a/lib/api/authn/index.js
+++ b/lib/api/authn/index.js
@@ -16,20 +16,8 @@ function overrideWith (forceUserId) {
   }
 }
 
-/**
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
@dmitrizagidulin Thanks for reviewing! Could you deploy this at https://solidtest.space at some point so we can verify this with https://github.com/solid/solid-auth-client/issues/11?

dmitrizagidulin:
Deployed!

PR #526, opened 2017-07-25
Reject cookies from third-party applications

### LINKED ISSUE #524: Do not accept the Express session cookie from origins other than the server's

RubenVerborgh:
Otherwise, when a user is logged in to their Solid server,
any third-party application could perform authenticated requests
without permission by including the credentials set by the Solid server.

Closes #524. Breaking change, needs new semver-major.

Note: this breaks Warp, which currently relies on cookie-based authentication (https://github.com/linkeddata/warp/issues/23).

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -20,6 +20,7 @@ const OidcManager = require('./models/oidc-manager')
 const config = require('./server-config')
 const defaults = require('../config/defaults')
 const options = require('./handlers/options')
+const debug = require('./debug').authentication
 
 const corsSettings = cors({
   methods: [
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
@dmitrizagidulin Updated check for equality with the hostnames of either the user ID or the server.

lib/create-app.js
```diff
@@ -143,9 +144,32 @@ function initViews (app, configPath) {
...
+    sessionHandler(req, res, () => {
+      // Reject cookies from third-party applications.
+      // Otherwise, when a user is logged in to their Solid server,
+      // any third-party application could perform authenticated requests
+      // without permission by including the credentials set by the Solid server.
+      const origin = req.headers.origin
+      const userId = req.session.userId
+      console.log('check')
+      if (origin && userId) {
+        const userIdHost = userId.replace(/([^:/])\/.*/, '$1')
```
dmitrizagidulin: I think this might grab the path, as well as the host? so, it'll be like `ruben.databox.me/profile/card`, yeah?

dmitrizagidulin: Ah, n/m :) See it.

RubenVerborgh:
@dmitrizagidulin Now with unit tests for the check.

dmitrizagidulin:
@RubenVerborgh 👍 , thank you!

timbl:
There seems to be no discussion about the way this interacts with CORS.   There is no isssue behind this PR?

RubenVerborgh:
@timbl The issue is tagged in the description, it's #524.

There's no interaction with CORS: we simply allow all cross-origin interactions on Solid servers, and also accept cookies with AJAX requests from any origin. As a result from that openness, we need to be careful about what requests we accept from what hosts, and the cookies mechanism wasn't doing that.

PR #528, opened 2017-08-09
Expose ACL and user ID on request

RubenVerborgh:
Cherry-picking this commit from `rv/patch/n3` (#516), since I will also need it for the Auth Proxy (`feature/authproxy`).


dmitrizagidulin:
👍

RubenVerborgh:
Note: this has been removed in favor of #531.

Issue #517, opened 2017-07-05
Move user ID checking to auth


RubenVerborgh:
See discussion at https://github.com/solid/node-solid-server/pull/516/files/d10875391abd400b0a18da0a4681e6cefb5773ac#r125661813

Depends on #516: the storage of `userId` in `request` as done by #516 should be changed.

PR #531, opened 2017-08-09
Split user ID extraction and ACL checking

RubenVerborgh:
The current `allow.js` performs two functions:

1. obtaining the user ID
2. checking whether the user is authenticated

As a consequence, functions are currently tied sequentially.

However, it is not unthinkable that other steps need to be performed in between: other handlers might want to know the user ID without/before checking permissions on the LDP storage. (Case in point: the proxy delegation I'm working on.)

This pull request therefore splits user ID extract and ACL checking.

### DIFF
--- a/lib/api/authn/force-user.js
+++ b/lib/api/authn/force-user.js
@@ -9,7 +9,6 @@ function initialize (app, argv) {
   app.use('/', (req, res, next) => {
     debug(`Identified user (override): ${forceUserId}`)
     req.session.userId = forceUserId
-    req.session.identified = true
     res.set('User', forceUserId)
     next()
   })
--- a/lib/api/authn/force-user.js
+++ b/lib/api/authn/force-user.js
@@ -9,7 +9,9 @@ function initialize (app, argv) {
   app.use('/', (req, res, next) => {
     debug(`Identified user (override): ${forceUserId}`)
     req.session.userId = forceUserId
-    res.set('User', forceUserId)
+    if (argv.auth === 'tls') {
+      res.set('User', forceUserId)
+    }
     next()
   })
 }
--- a/lib/api/authn/force-user.js
+++ b/lib/api/authn/force-user.js
@@ -1,0 +1,20 @@
+const debug = require('../../debug').authentication
+
+/**
+ * Enforces the `--force-user` server flag, hardcoding a webid for all requests,
+ * for testing purposes.
+ */
+function initialize (app, argv) {
+  const forceUserId = argv.forceUser
+  app.use('/', (req, res, next) => {
+    debug(`Identified user (override): ${forceUserId}`)
+    req.session.userId = forceUserId
+    req.session.identified = true
+    res.set('User', forceUserId)
+    next()
+  })
+}
+
+module.exports = {
+  initialize
+}
--- a/lib/api/authn/index.js
+++ b/lib/api/authn/index.js
@@ -1,23 +1,5 @@
-'use strict'
-
-const debug = require('../../debug').authentication
-
-/**
- * Enforces the `--force-user` server flag, hardcoding a webid for all requests,
- * for testing purposes.
- */
-function overrideWith (forceUserId) {
-  return (req, res, next) => {
-    req.session.userId = forceUserId
-    req.session.identified = true
-    debug('Identified user (override): ' + forceUserId)
-    res.set('User', forceUserId)
-    return next()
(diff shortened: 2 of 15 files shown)


RubenVerborgh:
@dmitrizagidulin This PR is not yet ready for merging, but I'm curious to hear your thoughts about this.

I still plan to do some cleaning up, for instance, incorporating `forceUser` into the new user ID handler.

Alternatively, we could just mandate that the existing WebID and OIDC handlers store the user ID in `req.userId` themselves. (However, the commented-out code in `allow.js`  seemed to include a delegation mechanism as well, which could have its place in `userid.js` then.)

RubenVerborgh:
Thinking about this some more, I believe we should change the contract and have the authenticators assign to `request.userId` (or at least to session). The allow.js/userid.js files now need to know about auth methods, and that should not be the case.

dmitrizagidulin:
Sounds good, see also the related issue https://github.com/solid/node-solid-server/issues/517

Action items:

- Remove `req.session.identified` boolean (is redundant), so the contract is now: if `req.session.userId` exists, user is identified & logged in
- as discussed, we don't need to set `req.userId`, since all the authenticators already set `req.session.userId`. (Meaning, downstream consumers like allow can just look in req.session).

RubenVerborgh:
I've removed the `identified` flag and `userId`.

After some additional refactoring, `initAuthentication` is now [really lean](https://github.com/solid/node-solid-server/pull/531/commits/b13377fd68b6aed465eaeb9ec7dcda91a1287782#diff-237045f786d4918cf906b96a1c163eb9R198) and independent of any auth.

A review on the current state is welcome. I'll check in the meantime whether this fixes the new auth proxy's problems.

lib/api/authn/force-user.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Enforces the `--force-user` server flag, hardcoding a webid for all requests,
+ * for testing purposes.
+ */
+function initialize (app, argv) {
+  const forceUserId = argv.forceUser
+  app.use('/', (req, res, next) => {
+    debug(`Identified user (override): ${forceUserId}`)
+    req.session.userId = forceUserId
+    res.set('User', forceUserId)
```
dmitrizagidulin: I think we're no longer returning the User header on non-TLS auth methods?

RubenVerborgh: That's right, but it was still in the `forceUser` code I moved here. Maybe this is useful for faking a TLS authentication? (After all, the client needs _some_ way to find out who the logged-in user is.)

dmitrizagidulin: It was, yeah. I think we just forgot to remove it here. 
So, yes, it's useful to fake TLS authentication, but that should still be taken care of in the webid-tls.js `handler()` function. (As in, forceUser id sets the session.userId, and the TLS handler short-circuits on there being a session.userId already).

RubenVerborgh: I don't think that's the case: the TLS handler is not attached in presence of `forceUser` (neither in my code nor the original): https://github.com/solid/node-solid-server/pull/531/commits/b13377fd68b6aed465eaeb9ec7dcda91a1287782#diff-237045f786d4918cf906b96a1c163eb9L203

dmitrizagidulin: ahh, you're right. so then maybe we should wrap setting that header in an auth method check (only do it if authMethod is tls)

dmitrizagidulin: (inside the forceUser function that is)

RubenVerborgh: That makes sense, let me do that.

RubenVerborgh: Done in d415b74928f5cffc85c90bc3beaac98a4938010b.

dmitrizagidulin:
Looks good! Thanks @RubenVerborgh. 

Will you need to do further refactoring in terms of ACL checking, or is this ready to merge?

RubenVerborgh:
Not in this pull request at least, so please go ahead!

RubenVerborgh:
This pull request can also simplify integration tests, as all authentication methods now have the same contract. Most integration tests can just use `forceUser` instead of having to choose between `tls` or `oidc` (with the [added complexity](https://github.com/solid/node-solid-server/pull/516/commits/338795ee825163f2976535e2196b171e1cc69d23)).

PR #532, opened 2017-08-09
Set default test timeout to 10 seconds.

RubenVerborgh:
Fixes random errors on Travis.

Only 2 overrides left (20 seconds) in `test/integration/account-creation-oidc.js`.


dmitrizagidulin:
Awesome! I didn't realize a global timeout was an option! 👍

PR #534, opened 2017-08-10
Update CHANGELOG for version 4.0.0

RubenVerborgh:


RubenVerborgh:
Should check that all relevant CHANGELOG entries are mirrored in README (#544) and vice versa.

PR #535, opened 2017-08-11
Add startup script for testing purposes

RubenVerborgh:
The [security setting](https://github.com/solid/node-solid-server/pull/518/commits/293c86102277cc035807c0e5aebff5ba119dce22) associated with Node 8 compatibility (#518) means the server will only work with trusted certificates. To avoid this for testing purposes, the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable should be set.

In order to facilitate that, this pull request adds a `bin/solid-test` executable. For consistency, it also renames the default executable to `bin/solid` from `bin/solid.js`, but the latter is symlinked.

Open questions:
- [x] Is `solid-test` an appropriate name? I thought of alternatives such as `solid-insecure`, but that seemed to strong. In any case, the executable is _not_ exported in `package.json`.
- [x] Where should documentation for this feature be added? (In any case, to the changelog [#534].)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,14 @@ $ solid start --root path/to/folder --port 8443 --ssl-key path/to/ssl-key.pem --
 # Solid server (solid v0.2.24) running on https://localhost:8443/
 ```
 
-##### How do I get an SSL key and certificate?
-You need an SSL certificate you get this from your domain provider or for free from [Let's Encrypt!](https://letsencrypt.org/getting-started/).
+### Running in development environments
 
-If you don't have one yet, or you just want to test `solid`, generate a certificate (**DO NOT USE IN PRODUCTION**):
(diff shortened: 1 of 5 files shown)


dmitrizagidulin:
I think naming it `solid-test` would carry across the right connotation. (We can also automatically prepend the `DEBUG="solid:*"` env variable, too, while we're at it).

As for documenting, aside from the CHANGELOG, I think we should also add it to the README, in the Command Line Usage / Running a Server sections. What do you think?

dmitrizagidulin:
I'm also not sure whether it's worth leaving `solid-test` out of the `npm` scripts..

RubenVerborgh:
> (We can also automatically prepend the DEBUG="solid:*" env variable, too, while we're at it).

Maybe not, as people might like to make their own, more specific settings?

> I think we should also add it to the README, in the Command Line Usage / Running a Server sections.

Will do!

> I'm also not sure whether it's worth leaving solid-test out of the npm scripts.

Well, there's `scripts` and there is `bin`. I don't think it has a place in `bin`, but maybe in `scripts`. Then the question is what script? I thought about `start` (as this is commonly used), but there's also already a `solid` entry. Maybe a `solid-test` entry then?

RubenVerborgh:
@dmitrizagidulin Documentation added, ready for review.

PR #536, opened 2017-08-11
Follow Mocha naming conventions

RubenVerborgh:
When following Mocha conventions, the file argument can be removed from the `mocha` script in `package.json`.

As a consequence, individual files can be tested easily:
```
npm run mocha -- test/unit/account-manager-test.js
```


dmitrizagidulin:
Ok, excellent. Lets get rid of those extra `test-*` scripts, and should be good to go.

RubenVerborgh:
Done!

Issue #521, opened 2017-07-19
Allow the WebID and OIDC providers to reside on different domains


RubenVerborgh:
Right now, OIDC providers can only authenticate users on their own domain.

However, the WebID resource could explicitly indicated trusted OIDC domains or providers. For example, then I'd be able to authenticate as https://ruben.verborgh.org/profile/#me from https://drive.verborgh.org/ with OIDC. (Right now, I can only do that via WebID+TLS.)

elf-pavlik:
Searching for 'OpenID Connect delegation', I've found this StackOverflow answer https://stackoverflow.com/a/28022632 which links to http://openid.net/specs/openid-connect-discovery-1_0.html#URLSyntax

Would something like below work?

```turtle
https://ruben.verborgh.org/profile/#me
  http://openid.net/specs/connect/1.0/issuer https://drive.verborgh.org/ .
```

dmitrizagidulin:
great find, @elf-pavlik , re `http://openid.net/specs/connect/1.0/issuer` rel!

dmitrizagidulin:
@elf-pavlik After some discussion with @timbl, we decided not to go with `http://openid.net/specs/connect/1.0/issuer`, use `solid:oidcIssuer` instead:

```ttl
<https://ruben.verborgh.org/profile/#me>
  <http://www.w3.org/ns/solid/terms#oidcIssuer> <https://drive.verborgh.org> .
```

dmitrizagidulin:
I suppose we can add some sort of rdfs/owl equivalence to the specs issuer link, but I'm not familiar enough with ontologies to do that. Feel free to make an addition to https://github.com/solid/vocab/pull/23, though.

RubenVerborgh:
Depends on the semantics. Are they equivalent? Or is `solid:oidcIssuer` something more specific?

dmitrizagidulin:
You're right, I suppose `solid:oidcIssuer` is more specific; it denotes that the server supports the WebID-OIDC protocol specifically.

RubenVerborgh:
Done: https://github.com/solid/vocab/pull/23/commits/c70a6c4e3a9171a7f3ced046b5bed7bfcf5ec166

dmitrizagidulin:
Nice!

RubenVerborgh:
Fixed by #545.

RubenVerborgh:
[...]

PR #545, opened 2017-08-16
Allow login via TLS with externally hosted WebIDs

dmitrizagidulin:
(after verifying that the server is the preferred oidc issuer of that webid profile).

Addresses https://github.com/solid/node-solid-server/issues/521

### DIFF
--- a/lib/models/account-manager.js
+++ b/lib/models/account-manager.js
@@ -350,11 +350,11 @@ class AccountManager {
         this.accountWebIdFor(userData.username)
     }
 
-    if (!userConfig.webId && !userConfig.username) {
-      throw new Error('Username or web id is required')
-    }
+    if (!userConfig.username) {
+      if (!userConfig.webId) {
+        throw new Error('Username or web id is required')
+      }
 
-    if (userConfig.webId && !userConfig.username) {
       userConfig.username = this.usernameFromWebId(userConfig.webId)
     }
 
@@ -391,14 +391,6 @@ class AccountManager {
       .then(() => {
         return template.processAccount(accountDir)
       })
-  }
-
-  externalAccount (webId) {
-    let webIdHostname = url.parse(webId).hostname
-
-    let serverHostname = this.host.hostname
-
-    return !webIdHostname.endsWith(serverHostname)
   }
 
   /**
--- a/lib/models/account-manager.js
+++ b/lib/models/account-manager.js
@@ -345,16 +345,17 @@ class AccountManager {
       username: userData.username,
       email: userData.email,
       name: userData.name,
+      externalWebId: userData.externalWebId,
       webId: userData.webid || userData.webId ||
         this.accountWebIdFor(userData.username)
     }
 
-    if (userConfig.webId && !userConfig.username) {
-      userConfig.username = this.usernameFromWebId(userConfig.webId)
-    }
-
     if (!userConfig.webId && !userConfig.username) {
       throw new Error('Username or web id is required')
+    }
+
+    if (userConfig.webId && !userConfig.username) {
+      userConfig.username = this.usernameFromWebId(userConfig.webId)
     }
 
     return UserAccount.from(userConfig)
--- a/lib/models/authenticator.js
+++ b/lib/models/authenticator.js
@@ -3,6 +3,8 @@
 const debug = require('./../debug').authentication
 const validUrl = require('valid-url')
 const webid = require('webid/tls')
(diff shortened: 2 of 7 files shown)


lib/models/account-manager.js
```diff
@@ -345,18 +345,19 @@ class AccountManager {
...
-    }
-
     if (!userConfig.webId && !userConfig.username) {
       throw new Error('Username or web id is required')
     }
 
+    if (userConfig.webId && !userConfig.username) {
+      userConfig.username = this.usernameFromWebId(userConfig.webId)
+    }
+
```
RubenVerborgh: @dmitrizagidulin MIght want to group those two `if`s in an `if (!userConfig.username)`?

lib/models/authenticator.js
```diff
@@ -304,13 +310,27 @@ class TlsAuthenticator extends Authenticator {
    * @return {UserAccount}
    */
   ensureLocalUser (webId) {
-    if (this.accountManager.externalAccount(webId)) {
-      debug(`WebID URI ${JSON.stringify(webId)} is not a local account`)
+    const serverUri = this.accountManager.host.serverUri
 
-      throw new Error('Cannot login: Selected Web ID is not hosted on this server')
+    // if (this.accountManager.externalAccount(webId)) {
```
RubenVerborgh: @dmitrizagidulin Should this commented code still be here?

lib/models/authenticator.js
```diff
@@ -304,13 +310,27 @@ class TlsAuthenticator extends Authenticator {
...
+    if (domainMatches(serverUri, webId)) {
+      // This is a locally hosted Web ID
+      return Promise.resolve(this.accountManager.userAccountFrom({ webId }))
     }
 
-    return this.accountManager.userAccountFrom({ webId })
+    debug(`WebID URI ${JSON.stringify(webId)} is not a local account, verifying preferred provider`)
+
+    return this.discoverProviderFor(webId)
+      .then(preferredProvider => {
```
RubenVerborgh: I wonder is "preferred" is too strong, given that we only have a predicate to indicate the provider (and we simply assume there is only one).

lib/models/authenticator.js
dmitrizagidulin: Whoops, thanks

lib/models/authenticator.js
dmitrizagidulin: Ah, true. I'll change it to 'authorized provider', then.

lib/models/account-manager.js
dmitrizagidulin: Sure thing.

dmitrizagidulin:
@RubenVerborgh updated. Let me know if there's anything else

RubenVerborgh:
I've tested this with `bin/solid-test`  from #546 merged in.

### Questions beforehand
- What does `"auth": "oidc"` in my config still mean? Will it be both by default now?

### Logging in without an account (and without the right triple in my profile)
- I click "With certificate", and I receive "This server is not the authorized provider for Web ID https://ruben.verborgh.org/profile/#me". Perhaps this error message can be clarified with a solution? Such as "Please authorize add <x> <y> <z> to your profile." Then again, WebID-TLS will only be used by advanced users.

### Logging in without an account (but with the right triple)
- Logging in works. I get `Access denied for https://ruben.verborgh.org/profile/#me`, but that's because my data folder is empty.

So good to go for me, perhaps just the `"auth"` setting needs clarification.

dmitrizagidulin:
> What does `"auth": "oidc"` in my config still mean? Will it be both by default now?

So, it's not quite both, by default. (I'll add the explanation below to the CHANGELOG upgrade notes).

`"auth": "tls"` means: "use the old WebID-TLS auth system", which means that the server prompts for a certificate on each request, etc. And the OIDC components are not initialized & not used.

`"auth": "oidc"` means: "use the WebID-OIDC auth system", and *optionally*, if the config entry `"localAuth": { "tls": <boolean> }` is not set to *false*, TLS will be offered as a *local login method*. However, the server will not prompt for certificates on each request, only on the TLS login api endpoint.

Does that make more sense? 

> This server is not the authorized provider error message

Yeah, agreed, that could be clarified. But like you said, it's for advanced users, there's basically gonna be like 4 users total who would want to use that.

dmitrizagidulin:
What would be the best way to clarify the error message? Maybe link to the section in the spec about discovery from profile RDF?

RubenVerborgh:
A link would probably be all that users need, indeed. Thanks!

PR #546, opened 2017-08-16
Set rejectUnauthorized on solid-test

RubenVerborgh:


RubenVerborgh:
@dmitrizagidulin Another possibility is to rename the option into  `--no-reject-unauthorized`.

dmitrizagidulin:
Renaming it to `--no-reject-unauthorized` seems reasonable. (Denotes that rejecting is the normal state, and that not doing it is the exception.)

RubenVerborgh:
@dmitrizagidulin Thanks, updated.

RubenVerborgh:
@dmitrizagidulin …and without Warp this time.

dmitrizagidulin:
Hehe, thanks

Issue #246, opened 2016-02-12
Return 'wac-allowed' HTTP header with GET & HEAD responses


dmitrizagidulin:
**Problem/Motivation**

There is currently no easy way for solid client code to determine if the current user has read/write/delete etc access to a given resource (short of reading and parsing the ACL chain, which is both discouraged, and not always possible due to ACL limitations on the `.acl` files themselves). (See [solid/#45 - No way to find out if user is authorized to write to a container](https://github.com/solid/solid/issues/45) for context.)

**Proposed Solution**

An easy solution to this would be to add support to LDNode for the following:

With every response, the server should include an additional `WAC-Modes-Allowed:` header in the HTTP response.

For example, doing an unauthenticated HEAD request on a public read-only resource, would yield:

```
WAC-Allowed: user="read", public="read"
```

Doing an HEAD request (while authenticated) to a resource you have full read/write/control access to:

```
WAC-Allowed: user="read;write;control", public=""
```

(Note the `public=""` part -- the empty string denotes no access.)

dmitrizagidulin:
Updated issue to use `Allow:` header instead of CORS `Allow-Methods`

rhiaro:
+1

dmitrizagidulin:
Note that since we don't have a `http://www.w3.org/ns/auth/acl#None` access mode, and since it's problematic to return empty HTTP headers, if the client making the request has no read/write access to a resource, the response will not contain a `WAC-Modes-Allowed` header.

dmitrizagidulin:
Updated it to reflect today's design discussion; return both the permissions of the user, and public permissions.

RubenVerborgh:
Implemented by #550.

PR #550, opened 2017-08-17
Expose the user's permissions through a header

RubenVerborgh:
Implements #246 with `WAC-Allowed: user="read write append",public="read"`.

### DIFF
--- a/lib/header.js
+++ b/lib/header.js
@@ -2,6 +2,7 @@ module.exports.addLink = addLink
 module.exports.addLinks = addLinks
 module.exports.parseMetadataFromHeader = parseMetadataFromHeader
 module.exports.linksHandler = linksHandler
+module.exports.addPermissions = addPermissions
 
 var li = require('li')
 var path = require('path')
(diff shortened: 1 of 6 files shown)


csarven:
Good stuff!

Would "WAC-Mode" be preferable just to keep it closer to *nix's *mode bits*?

RubenVerborgh:
@csarven The header name was chosen in a discussion with @timbl and @dmitrizagidulin yesterday (see #246). I don't mind the naming either way (although I still think "WAC" is cryptic).

dmitrizagidulin:
I just realized we want to do `WAC-Allow`, minus the *ed* on the end. (To match the HTTP `Allow:` header). Sorry bout that :)

RubenVerborgh:
Which of these would you prefer? I don't mind either way; I was following what @dmitrizagidulin wrote down of our discussion in #246.

csarven:
To stick close to convention (like in Web Linking for relation types), probably best to use space as a delimiter. Plus, it is easier to read.

RubenVerborgh:
Makes sense! Are the quotes needed?

csarven:
I think the quotes might make it easier for consumers to tokenise. It can also help to escape characters or have specialised modes or something in the future.

RubenVerborgh:
Okay, mandatory quotes it is.

RubenVerborgh:
250e5132f79ebe5c0d2d84b21ae7de78a7cf76bf brings spaces.

RubenVerborgh:
Currently on hold because of #552.

RubenVerborgh:
#552 has been merged, so this one is ready for review and merge.

PR #552, opened 2017-08-18
Add caching to ACL

RubenVerborgh:
Avoid that the same .acl files are read and parsed multiple times per request.

### DIFF
--- a/lib/acl-checker.js
+++ b/lib/acl-checker.js
@@ -26,17 +26,10 @@ class ACLChecker {
 
     // Check the permissions within the nearest ACL
     return this.getNearestACL(resource)
-    .then(({ acl, graph, isContainer }) =>
-      this.checkAccess(
-        graph, // The ACL graph
-        user, // The webId of the user
-        mode, // Read/Write/Append
-        resource, // The resource we want to access
-        isContainer,
-        acl, // The current Acl file!
-        options
-      )
-    )
+    .then(nearestAcl => {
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
@dmitrizagidulin You might want to have this as a merge commit rather than rebase, in order to keep the commits together.

dmitrizagidulin:
@RubenVerborgh good call, re merge

lib/acl-checker.js
```diff
@@ -1,172 +1,143 @@
...
+
+    // Obtain the permission set for the resource
+    if (!this._permissionSet) {
+      this._permissionSet = this.getNearestACL()
+        .then(acl => this.getPermissionSet(acl))
+    }
+
+    // Check the resource's permissions
+    return this._permissionSet.then(acls => this.checkAccess(acls, user, mode))
+    .catch(err => {
```
dmitrizagidulin: I think this should be indented? Something like:

```js
return this._permissionSet
  .then(acls => this.checkAccess(acls, user, mode))
  .catch(err => {
  })
```
?

RubenVerborgh: Makes sense, will do. (`standard` didn't complain; but I'd be the last person to say that it is a good style checker :wink:)

RubenVerborgh: Done in bda742636e346fff02f6f082b10d37bbd524013b, let me know if more indentation needs adjustments.

dmitrizagidulin:
Oh man, great work on this!

PR #553, opened 2017-08-18
Add support for external WebIDs registering with username & password

dmitrizagidulin:
(Note: changes default templates & views, be sure to clear `config/views` & `config/templates` before testing.)

### DIFF
--- a/default-templates/new-account/index.html
+++ b/default-templates/new-account/index.html
@@ -14,12 +14,7 @@
   <div class="row">
     <div class="col-md-12">
       <p style="margin-top: 3em; margin-bottom: 3em;">
-        Welcome.
-      </p>
-      <p>
-        Web ID:<br />
-
-        <code>{{webId}}</code>
+        Welcome to the account of <code>{{webId}}</code>
       </p>
     </div>
   </div>
--- a/default-templates/new-account/index.html
+++ b/default-templates/new-account/index.html
@@ -14,10 +14,10 @@
   <div class="row">
     <div class="col-md-12">
       <p style="margin-top: 3em; margin-bottom: 3em;">
-        Welcome to your Solid user profile.
+        Welcome.
       </p>
       <p>
-        Your Web ID is:<br />
+        Web ID:<br />
 
         <code>{{webId}}</code>
       </p>
--- a/default-templates/new-account/profile/card
+++ b/default-templates/new-account/profile/card
@@ -18,7 +18,6 @@
     solid:account </> ;  # link to the account uri
     pim:storage </> ;    # root storage
 
-    solid:inbox </inbox/> ;
     ldp:inbox </inbox/> ;
 
     pim:preferencesFile </settings/prefs.ttl> ;  # private settings/preferences
--- a/default-templates/new-account/profile/card
+++ b/default-templates/new-account/profile/card
@@ -6,10 +6,10 @@
 
 <>
     a foaf:PersonalProfileDocument ;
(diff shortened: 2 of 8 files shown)


RubenVerborgh:
### Logging in without an account
> This server is not the authorized provider for Web ID https://ruben.verborgh.org/profile/#me. See https://github.com/solid/webid-oidc-spec#authorized-oidc-issuer-discovery

Nice!

### Registering with a garbage WebID

Tried registering with "xyz" as a WebID, it let me… Might be good to add a URI check for people who don't know what WebID is (and might fill out things like their username, "none", etc.).

### Registering with an actual WebID

Works fine; I can login with username as well as WebID, and I get the proper error message when the triple is not set in my profile.

default-templates/new-account/index.html
```diff
@@ -14,10 +14,10 @@
   <div class="row">
     <div class="col-md-12">
       <p style="margin-top: 3em; margin-bottom: 3em;">
-        Welcome to your Solid user profile.
+        Welcome.
       </p>
       <p>
-        Your Web ID is:<br />
+        Web ID:<br />
```
RubenVerborgh: Would it make sense to make this more descriptive? As in `The owner of this page/space is {{webId}}`? There was some confusion in the past with "Your Web ID is", and this removes some confusion, but maybe not all.

default-templates/new-account/profile/card
```diff
@@ -18,6 +18,7 @@
     solid:account </> ;  # link to the account uri
     pim:storage </> ;    # root storage
 
+    solid:inbox </inbox/> ;
```
RubenVerborgh: Do we need both?

lib/handlers/error-pages.js
```diff
@@ -15,7 +15,7 @@ const SELECT_PROVIDER_AUTH_METHODS = ['oidc']
  * @param next {Function}
  */
 function handler (err, req, res, next) {
-  debug('Error page because of ' + err.message)
+  debug('Error page because of:', err)
```
RubenVerborgh: Can debug handle a second argument?

lib/models/account-manager.js
```diff
@@ -346,16 +346,32 @@ class AccountManager {
...
+      } else {
+        throw err
       }
+    }
 
-      userConfig.username = this.usernameFromWebId(userConfig.webId)
+    if (userConfig.username) {
+      if (userConfig.externalWebId && !userConfig.localAccountId) {
+        // External Web ID exists, derive the local account id from username
+        userConfig.localAccountId = this.accountWebIdFor(userConfig.username)
```
RubenVerborgh: So this means I can automatically get a username assigned?

default-templates/new-account/index.html
dmitrizagidulin: Yeah, good idea

lib/handlers/error-pages.js
dmitrizagidulin: Yeah, it's pretty great, like console.log

lib/models/account-manager.js
dmitrizagidulin: Sort of. the localAccountId just maps from the local server's username-based url to the external Web ID

dmitrizagidulin:
Whoops, I forgot this PR was still unmerged.

default-templates/new-account/profile/card
dmitrizagidulin: Good point, we do not.

PR #555, opened 2017-08-18
Remove dependency on async library

RubenVerborgh:


lib/ldp.js
```diff
@@ -156,42 +154,40 @@ class LDP {
...
-      function (err, data) {
+    // add container stats
+    new Promise((resolve, reject) =>
+      ldpContainer.addContainerStats(ldp, reqUri, filename, resourceGraph,
+        err => err ? reject(err) : resolve())
+    )
+    // read directory
+    .then(() => new Promise((resolve, reject) =>
+      ldpContainer.readdir(filename,
+        (err, files) => err ? reject(err) : resolve(files))
```
dan-f: It seems like a node convention for pre-Promise API async functions to take callbacks which themselves take a possible error as the first argument and a successful value as the next argument. We could write a higher-order function to make these kinds of functions return promises:

```js
const promisify = (fn) =>
  (...args) =>
    new Promise((resolve, reject) =>
      fn(...args, (err, success) =>
        err
          ? reject(err)
          : resolve(success)
      )
    )
```

test/integration/cors-proxy-test.js
```diff
@@ -108,30 +107,18 @@ describe('CORS Proxy', () => {
...
-        nock('https://example.org').get('/').reply(200)
-        server.get('/proxy/?uri=https://example.org/')
-          .expect(200, next)
-      }
-    ], done)
+  it('should return the same HTTP status code as the uri', () => {
+    nock('https://example.org').get('/404').reply(404)
+    nock('https://example.org').get('/401').reply(401)
+    nock('https://example.org').get('/500').reply(500)
+    nock('https://example.org').get('/200').reply(200)
```
dan-f: You can replace these four lines with:

```js

nock('https://example.org')
  .get('/404').reply(404)
  .get('/401').reply(401)
  .get('/500').reply(500)
  .get('/200').reply(200) 
```

dan-f:
Just a few suggestions, but no blockers from me. 👍, nice work!

lib/handlers/get.js
```diff
@@ -161,35 +160,28 @@ function globHandler (req, res, next) {
     let reqOrigin = utils.getBaseUri(req)
 
     debugGlob('found matches ' + matches)
-    async.each(matches, function (match, done) {
+    Promise.all(matches.map(match => new Promise((resolve, reject) => {
       var baseUri = utils.filenameToBaseUri(match, reqOrigin, root)
       fs.readFile(match, {encoding: 'utf8'}, function (err, fileData) {
```
RubenVerborgh: Yeah, unsure. I just ported it.

lib/ldp.js
```diff
@@ -156,42 +154,40 @@ class LDP {
...
-                filename, file, cb)
-            },
-            next)
-        }
-      ],
-      function (err, data) {
+    // add container stats
+    new Promise((resolve, reject) =>
+      ldpContainer.addContainerStats(ldp, reqUri, filename, resourceGraph,
+        err => err ? reject(err) : resolve())
```
RubenVerborgh: I can address this, as there are some other points as well. What indentation would you use here?

dan-f: It's definitely a matter of taste but this is how I generally like things: when breaking function arguments up into several lines, I like to keep the closing paren `)` on its own line, indented to match the start of the line of the function call.

For example:

```js
ldpContainer.addContainerStats(
  ldp, reqUri, filename, resourceGraph,
  err => err ? reject(err) : resolve()
)
```

I also see this a lot when using callback functions - breaking the line after the arguments to the callback:
```js
ldpContainer.addContainerStats(ldp, reqUri, filename, resourceGraph, err =>
  err ? reject(err) : resolve()
)
```

lib/ldp.js
RubenVerborgh: We could do that—and there are a couple of libraries that expose such a function (starting with Node 8 even in core). I've used that in other projects.

I was hesitant to do this here and in other places of the Solid code, because callbacks are sometimes used inconsistently (they don't always have an error argument). So I could start using it here, but then it would perhaps be good to keep an eye on it in other places of the code as well. And eventually, we should migrate to an `await`/`async` style, I suppose.

So, what do you think, should we start the promisify practice here?
Perhaps we should consider switching to `async`/`await` (with build step) for node-solid-server v5? Or already start it here?

test/integration/cors-proxy-test.js
RubenVerborgh: Thanks, will do.

lib/ldp.js
RubenVerborgh: We should really consider using a linter that goes beyond `standard`. There are still lots of variations in the code.

This particular instance can perhaps be solved with `promisify` (see below).

lib/ldp.js
dan-f: Good points!

> I was hesitant to do this here and in other places of the Solid code, because callbacks are sometimes used inconsistently (they don't always have an error argument).

Oof yeah that complicates things!

> And eventually, we should migrate to an await/async style, I suppose.

Sure thing. Though async/await is syntactic sugar on top of Promises, so I think that "promisifying" our interfaces goes hand-in-hand with async/await-ifying them :)

I like your point that this change is more of an interface redesign than simply "promisifying" the existing APIs. How about we simply leave things as they are here, and I'll create a ticket for making the async APIs more consistent.

RubenVerborgh: :+1:, I'll make the other changes and merge.

dan-f: (created issue #557)

lib/ldp.js
dan-f: Yeah, fair point. A code formatter would be nice. I thought I saw you discussing [prettier](https://github.com/prettier/prettier) at some point in gitter. I'd be in favor of adding that, for no other reason than not having to think about style ever again :)

RubenVerborgh: [Won't have been me](https://github.com/RubenVerborgh/N3.js/issues/111) 😄  I'm in favor of a [strict .eslintrc](https://github.com/RubenVerborgh/N3.js/blob/master/.eslintrc) (and really dislike `standard` :wink:).

RubenVerborgh: I'm leaving this as-is, since other parts in the same file follow the same indentation.

dan-f: That's fine. Consistency is more important :)

PR #556, opened 2017-08-21
Expose WAC-Allow to browser clients

RubenVerborgh:
Sets the proper CORS permissions on the header introduced in #550.

### DIFF
--- a/test/integration/header-test.js
+++ b/test/integration/header-test.js
@@ -23,17 +23,26 @@ describe('Header handler', () => {
   describe('WAC-Allow', () => {
     describeHeaderTest('read/append for the public', {
       resource: '/public-ra',
-      headers: { 'WAC-Allow': 'user="read append",public="read append"' }
+      headers: {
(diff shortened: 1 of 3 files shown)


test/integration/header-test.js
```diff
@@ -23,17 +23,26 @@ describe('Header handler', () => {
...
+        'WAC-Allow': 'user="read append",public="read append"',
+        'Access-Control-Expose-Headers': /(^|,\s*)WAC-Allow(,|$)/
+      }
     })
 
     describeHeaderTest('read/write for the user, read for the public', {
       resource: '/user-rw-public-r',
-      headers: { 'WAC-Allow': 'user="read write append",public="read"' }
+      headers: {
+        'WAC-Allow': 'user="read write append",public="read"',
```
dan-f: Is `append` implied by `write`?

RubenVerborgh: Yes, the current ACL checker sets Append whenever it finds Write.

PR #561, opened 2017-08-30
Disable rejectUnauthorized on the WebID-TLS endpoint

RubenVerborgh:


RubenVerborgh:
Mmm broken on Node 6, let me investigate.

PR #563, opened 2017-08-30
Only set session.returnToUrl if not already set

dmitrizagidulin:


RubenVerborgh:
@dmitrizagidulin What is the context of this PR? Should this be in v4.0.0?

dmitrizagidulin:
Closing this PR for now; the issue it was trying to fix seems not to be a problem.

PR #565, opened 2017-08-31
Fix requirement for additional verification logging in with WebID-TLS

dmitrizagidulin:
Fixes the issue of requiring the `solid:oidcIssuer` triple in the profile when logging in via WebID-TLS
with externally hosted Web IDs.

### DIFF
--- a/lib/models/authenticator.js
+++ b/lib/models/authenticator.js
@@ -214,7 +214,7 @@ class TlsAuthenticator extends Authenticator {
 
       .then(cert => this.extractWebId(cert))
 
-      .then(webId => this.ensureLocalUser(webId))
+      .then(webId => this.loadUser(webId))
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/models/authenticator.js
```diff
@@ -304,36 +304,23 @@ class TlsAuthenticator extends Authenticator {
...
-    }
-
-    debug(`WebID URI ${JSON.stringify(webId)} is not a local account, verifying authorized provider`)
-
-    return this.discoverProviderFor(webId)
-      .then(authorizedProvider => {
-        debug(`Authorized provider for ${webId} is ${authorizedProvider}`)
+      return this.accountManager.userAccountFrom({ webId })
+    } else {
+      debug(`WebID URI ${JSON.stringify(webId)} is not a local account, using it as an external WebID`)
```
RubenVerborgh: Yes, but are we still checking that username/password authentication for that account is not enabled? Because otherwise, I could register on any Solid pod with anybody's Web ID.

dmitrizagidulin: You can't register with anybody's Web ID since the server still checks your profile against your browser cert. So if you don't have the cert, the login will still fail

dmitrizagidulin: The regular webid-tls checking happens earlier in the workflow, in `extractWebId()` in authenticator / line 274

RubenVerborgh: Okay, happy that's covered!

PR #567, opened 2017-08-31
Update Data Browser html file

dmitrizagidulin:
To use the new auth client.

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -10,5 +10,6 @@ module.exports = {
   'dbPath': './.db',
   'port': 8443,
   'serverUri': 'https://localhost:8443',
-  'webid': true
+  'webid': true,
+  'dataBrowserPath': 'default'
 }
--- a/static/databrowser.html
+++ b/static/databrowser.html
@@ -5,16 +5,19 @@
     <script type="text/javascript" src="https://linkeddata.github.io/mashlib/dist/mashlib-prealpha.js"></script>
 <script>
 document.addEventListener('DOMContentLoaded', function() {
-    var UI = require('mashlib')
-    var $rdf = UI.rdf
+  var UI = Mashlib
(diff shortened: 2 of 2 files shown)


config/defaults.js
```diff
@@ -10,5 +10,6 @@ module.exports = {
   'dbPath': './.db',
   'port': 8443,
   'serverUri': 'https://localhost:8443',
-  'webid': true
+  'webid': true,
+  'dataBrowserPath': 'default'
```
RubenVerborgh: What are possible alternative values? Why isn't this `static/databrowser.html`?

RubenVerborgh:
Any steps to test?

config/defaults.js
dmitrizagidulin: Not quite sure why it was done this way. I do know that it's what activates it.

dmitrizagidulin:
To test, start up the server and request some rdf file on the server, say /settings/prefs.ttl or something. (Test on a browser that doesn't have an rdf plugin)

RubenVerborgh:
I get `Uncaught ReferenceError: Mashlib is not defined`.

dmitrizagidulin:
Ah damn, I think I forgot a git push

dmitrizagidulin:
Maybe there's a delay in propagation to github.io..

RubenVerborgh:
No change unfortunately.

dmitrizagidulin:
Found the issue - mashlib repo was still set to serve gh-pages instead of master on github.io. Should be fixed shortly..

dmitrizagidulin:
(might require a shift-refresh)

RubenVerborgh:
(BTW regarding gh-pages and master, see https://github.com/linkeddata/rdflib.js/pull/201.)

PR #569, opened 2017-09-01
Add a /public folder in new accounts

RubenVerborgh:
When an account is created, there's currently no folder that is publicly readable. This pull request adds `/public` with all permissions for the owner, and (inherited) read permissions for the public.


RubenVerborgh:
Needs a CHANGELOG entry (#534).

dan-f:
lgtm 👍

PR #570, opened 2017-09-01
Rename the idp option into multiuser

### LINKED ISSUE #515: Add multiuser as alternative to idp configuration property
The `idp` configuration property is non-intuitive; instead, we should consider adding `multiuser` (but do so in a backwards-compatible way).

RubenVerborgh:
Closes #515.

Please let me know whether you prefer `multiuser` or `multiUser`. I chose the former because it just seemed easier.

### DIFF
--- a/lib/create-server.js
+++ b/lib/create-server.js
@@ -23,7 +23,7 @@ function createServer (argv, app) {
   debug.settings('Base URL (--mount): ' + mount)
 
   if (argv.idp) {
-    console.error('The idp configuration option has been renamed to multiuser.')
+    console.warn('The idp configuration option has been renamed to multiuser.')
     argv.idp = argv.multiuser
     delete argv.idp
   }
--- a/lib/create-server.js
+++ b/lib/create-server.js
@@ -22,9 +22,15 @@ function createServer (argv, app) {
   app.use(mount, ldpApp)
   debug.settings('Base URL (--mount): ' + mount)
 
+  if (argv.idp) {
+    console.error('The idp configuration option has been renamed to multiuser.')
+    argv.idp = argv.multiuser
+    delete argv.idp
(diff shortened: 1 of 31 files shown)


RubenVerborgh:
Needs a CHANGELOG entry (#534).

lib/create-server.js
```diff
@@ -22,9 +22,15 @@ function createServer (argv, app) {
   app.use(mount, ldpApp)
   debug.settings('Base URL (--mount): ' + mount)
 
+  if (argv.idp) {
+    console.error('The idp configuration option has been renamed to multiuser.')
```
dan-f: Nitpick: may as well use `console.warn` so as not to print out a stack trace.

test/integration/account-creation-tls-test.js
```diff
@@ -18,7 +18,7 @@
 //     sslCert: path.join(__dirname, '../keys/cert.pem'),
 //     auth: 'tls',
 //     webid: true,
-//     idp: true,
+//     multiuser: true,
```
dan-f: This test file is completely commented out...

lib/create-server.js
RubenVerborgh: Good idea! This was copied from a similar warning elsewhere; will adjust that too.

test/integration/account-creation-tls-test.js
RubenVerborgh: I know; I'm not sure about the purpose of this file, I just search & replaced even in the commented code in case it is ever taken up again.

lib/create-server.js
RubenVerborgh: Done in 9e1b59fa631008a98c06b1f0689711fbfe775793.

PR #573, opened 2017-09-03
Use data browser for containers as well

### LINKED ISSUE #537: Fix or replace directory browser
The current Solid server integrates Warp, but Warp broke because of #526.

RubenVerborgh:
Removes Warp as well as the separate `--file-browser` option. Closes #537.

### DIFF
--- a/test/integration/http-test.js
+++ b/test/integration/http-test.js
@@ -276,11 +276,11 @@ describe('HTTP APIs', function () {
         .end(done)
     })
 
-    it('should redirect to file browser if container was requested as text/html', function (done) {
+    it('should show data browser if container was requested as text/html', function (done) {
       server.get('/sampleContainer2/')
         .set('Accept', 'text/html')
         .expect('content-type', /text\/html/)
-        .expect(303, done)
+        .expect(200, done)
     })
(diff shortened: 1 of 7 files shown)


RubenVerborgh:
Requires an entry in CHANGELOG (#534).

test/integration/http-test.js
```diff
@@ -276,11 +276,11 @@ describe('HTTP APIs', function () {
         .end(done)
     })
 
-    it('should redirect to file browser if container was requested as text/html', function (done) {
+    it('should show data browser if container was requested as text/html', function (done) {
       server.get('/sampleContainer2/')
         .set('Accept', 'text/html')
         .expect('content-type', /text\/html/)
-        .expect(303, done)
+        .expect(200, done)
```
dan-f: Not your problem, but this test does not in any way "show data browser if container was requested as text/html" 🤦‍♂️

dan-f:
👍 because this un-breaks file browsing on `node-solid-server`.

For the long term, I think it's worth discussing file browsing UX, and how to make the default experience accessible.

RubenVerborgh:
Thanks, @dan-f, added #578 for the UX discussion.

Issue #574, opened 2017-09-04
CLI - express injection


tomasklapka:
I would like to run another express middleware with the solid server while not losing the CLI and its processing of args and config.json. It means I need to mount the middleware before solid mounts to it's mount path.

There is an optional argument to inject express in function `createServer (argv, app)` but it is not used in CLI.

I'm thinking about refactoring the CLI part to be a reusable module with possibility to inject both solid and express.

Do you have another suggestion?

RubenVerborgh:
Actually, we have been discussing splitting the library into several parts, and all those parts could consist of Express-style handlers. For instance, the LDP handlers could live in a separate library, as could the permission handlers.

tomasklapka:
resolved by #576
thanks for accepting 👍

PR #576, opened 2017-09-05
CLI modularized, express injection allowed

tomasklapka:
As issued in #574 
This is a simple modularization of CLI so it is possible to reuse it and inject own express like this:
```js
#!/usr/bin/env node

const startCli = require('solid-server').startCli
const express = require('express')
const rdfFormatsProxy = require('http-rdf-formats-proxy')
const server = express()

server.get('/version', function (req, res) {
  res.json({ "version": require('./package.json').version })
})
server.use('/rdf-formats-proxy', rdfFormatsProxy())

startCli(server)
```
EDIT: renamed cli() to startCli()

### DIFF
--- a/bin/lib/cli.js
+++ b/bin/lib/cli.js
@@ -1,0 +1,15 @@
+var program = require('commander')
+var packageJson = require('../../package.json')
+var loadInit = require('./init')
+var loadStart = require('./start')
+
+module.exports = function cli (server) {
+  program
+  .version(packageJson.version)
+
+  loadInit(program)
+  loadStart(program, server)
+
+  program.parse(process.argv)
+  if (program.args.length === 0) program.help()
+}
--- a/bin/lib/cli.js
+++ b/bin/lib/cli.js
@@ -1,11 +1,10 @@
-var program = require('commander')
-var packageJson = require('../../package.json')
-var loadInit = require('./init')
-var loadStart = require('./start')
+const program = require('commander')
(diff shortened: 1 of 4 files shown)


bin/lib/cli.js
```diff
@@ -0,0 +1,15 @@
+var program = require('commander')
+var packageJson = require('../../package.json')
+var loadInit = require('./init')
+var loadStart = require('./start')
+
+module.exports = function cli (server) {
+  program
+  .version(packageJson.version)
```
RubenVerborgh: (nitpick) Make this one line?

bin/lib/cli.js
```diff
@@ -0,0 +1,15 @@
+var program = require('commander')
+var packageJson = require('../../package.json')
+var loadInit = require('./init')
+var loadStart = require('./start')
+
+module.exports = function cli (server) {
```
RubenVerborgh: Could the function name be more descriptive? Because it doesn't return anything. Maybe `startCli` or something like this?

bin/lib/cli.js
```diff
@@ -0,0 +1,15 @@
+var program = require('commander')
+var packageJson = require('../../package.json')
+var loadInit = require('./init')
+var loadStart = require('./start')
```
RubenVerborgh: Perhaps all of these can become `const` now.

RubenVerborgh:
Thanks!

Issue #275, opened 2016-03-07
Persist Content-Type of resources without file extensions


dmitrizagidulin:
(Creating a tracking issue for PRs https://github.com/linkeddata/ldnode/pull/159 and https://github.com/linkeddata/ldnode/pull/229)

**Problem:** When writing a resource without a file extension, there is currently no way for the server to remember what its content-type was. So, when it's next requested, the server doesn't know what that resource's content type is, and returns `text/turtle` by default.

**Potential Solutions:**
1. Add the content type to the resource's `.meta` file. Have the server check the `.meta` (if the resource doesn't have a file extension), discover the content type, and return it with the response.
2. (Route taken by PR #229) Append a transparent file extension to the resource, only when the resource does not include a file extension. For example, if you PUT a resource named `article` with an HTML content type, the file will be named `article.$.html` on disk. When reading the resource, ldnode detects the special extension (and filters it out), and returns the resource with the correct content-type.

sandhawke:
Do we have a spec for how resources live on the filesystem?    It seems like something a lot of tools might want to agree on, even though it's not the wire protocol, and not everyone has to use a filesystem, of course.   My copy-web-to/from-filesystem tool needs the same thing.

dmitrizagidulin:
@sandhawke  - What about solid servers that don't use the filesystem at all, like Meccano? 
There shouldn't really be a spec for that, since the only way apps (like the migrator app) should be accessing those files is through the API.

sandhawke:
@dmitrizagidulin I'm just talking about the systems that happen to want to join in, in using the same filesystem conventions.     Meccano may want to use it, too, for its export-in-a-zip-file and import-from-a-zip-file functions.

Other elements of this spec would be where metadata and acls go.   I'm thinking $describedby.ttl and $acl.ttl  (and this can be extended, where basically the word after the $ is the rel text)

nicola:
I agree with @dmitrizagidulin. Although it is nice to have everyone being on the same page on how to represent data on the file system, this is beyond the solid spec - in other words, everyone can implement it the way they want, the *import/export* that you mention should be done in a ldp way, exactly to avoid this problem, where all of this is abstracted.

sandhawke:
I completely agree it wouldn't be part of solid, per se.  That's why I asked if we "have a spec".   (Although I guess if you view solid as a collection of specs my phrasing is ambiguous.)

[...]

PR #584, opened 2017-09-06
Add file extension to generated paths on POST

RubenVerborgh:
This commit appends an extension to generated paths for POST requests,
reflecting the Content-Type of the request. This enables the server
to serve that file back with the correct content type.
For example, if a text/html request is posted, it will receive
the .html extension and will hence be served back as text/html.

Addresses the POST case of #275.

### DIFF
--- a/lib/handlers/post.js
+++ b/lib/handlers/post.js
@@ -6,6 +6,7 @@ var path = require('path')
 var header = require('../header')
 var patch = require('./patch')
 var error = require('../http-error')
+var { extensions } = require('mime-types')
 
 function handler (req, res, next) {
   var ldp = req.app.locals.ldp
@@ -80,20 +81,19 @@ function handler (req, res, next) {
 
   function one () {
     debug('Receving one file')
-    var linkHeader = header.parseMetadataFromHeader(req.get('Link'))
-    var slug = req.get('Slug')
-    ldp.post(
-      req.hostname,
-      containerPath,
-      slug,
-      req,
-      linkHeader.isBasicContainer,
+    const { slug, link, 'content-type': contentType } = req.headers
+    const links = header.parseMetadataFromHeader(link)
+    const mimeType = contentType ? contentType.replace(/\s*;.*/, '') : ''
+    const extension = mimeType in extensions ? `.${extensions[mimeType][0]}` : ''
+
+    ldp.post(req.hostname, containerPath, req,
+      { slug, extension, container: links.isBasicContainer },
       function (err, resourcePath) {
         if (err) {
           return next(err)
         }
         debug('File stored in ' + resourcePath)
-        header.addLinks(res, linkHeader)
+        header.addLinks(res, links)
         res.set('Location', resourcePath)
         res.sendStatus(201)
         next()
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -17,6 +17,7 @@ var ldpContainer = require('./ldp-container')
 var parse = require('./utils').parse
 
 const DEFAULT_CONTENT_TYPE = 'text/turtle'
+const DEFAULT_EXTENSION = '.ttl'
 
 const RDF_MIME_TYPES = [
   'text/turtle',            // .ttl
@@ -184,7 +185,7 @@ class LDP {
     .then(result => callback(null, result), callback)
   }
 
-  post (hostname, containerPath, slug, stream, container, callback) {
+  post (host, containerPath, stream, { container, slug, extension }, callback) {
     var ldp = this
     debug.handlers('POST -- On parent: ' + containerPath)
     // prepare slug
@@ -195,8 +196,10 @@ class LDP {
         return
       }
     }
+    // No need to use the default extension, as we know its MIME type already
+    if (extension === DEFAULT_EXTENSION) extension = ''
     // TODO: possibly package this in ldp.post
-    ldp.getAvailablePath(hostname, containerPath, slug, function (resourcePath) {
+    ldp.getAvailablePath(host, containerPath, { slug, extension }).then(resourcePath => {
       debug.handlers('POST -- Will create at: ' + resourcePath)
       let originalPath = resourcePath
       if (container) {
@@ -206,7 +209,7 @@ class LDP {
           originalPath += '/'
         }
       }
-      ldp.put(hostname, resourcePath, stream, function (err) {
+      ldp.put(host, resourcePath, stream, function (err) {
         if (err) callback(err)
         callback(null, originalPath)
       })
@@ -281,15 +284,15 @@ class LDP {
     })
   }
 
-  exists (host, reqPath, callback) {
-    var options = {
-      'hostname': host,
-      'path': reqPath,
-      'baseUri': undefined,
-      'includeBody': false,
-      'possibleRDFType': undefined
+  exists (hostname, path, callback) {
+    const options = { hostname, path, includeBody: false }
+    if (callback) {
+      return this.get(options, callback)
+    } else {
+      return new Promise((resolve, reject) => {
+        this.get(options, err => err ? reject(err) : resolve(true))
+      })
     }
-    this.get(options, callback)
   }
 
   /**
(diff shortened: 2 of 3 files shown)


csarven:
Agree with Tim's points. Okay, with the merge. For the record, I'm not in favour of basing the decision solely on the Content Type as opposed to what's in the body. See also https://github.com/solid/node-solid-server/issues/275#issuecomment-328080646

lib/handlers/post.js
```diff
@@ -80,20 +81,19 @@ function handler (req, res, next) {
...
-    var slug = req.get('Slug')
-    ldp.post(
-      req.hostname,
-      containerPath,
-      slug,
-      req,
-      linkHeader.isBasicContainer,
+    const { slug, link, 'content-type': contentType } = req.headers
+    const links = header.parseMetadataFromHeader(link)
+    const mimeType = contentType ? contentType.replace(/\s*;.*/, '') : ''
```
dan-f: I recommend you extract this logic into a function so that it can be tested in isolation.

lib/handlers/post.js
```diff
@@ -80,20 +81,19 @@ function handler (req, res, next) {
...
+    const extension = mimeType in extensions ? `.${extensions[mimeType][0]}` : ''
```
dan-f: Ditto here.

lib/handlers/post.js
```diff
@@ -80,20 +81,19 @@ function handler (req, res, next) {
...
+
+    ldp.post(req.hostname, containerPath, req,
+      { slug, extension, container: links.isBasicContainer },
       function (err, resourcePath) {
         if (err) {
           return next(err)
         }
         debug('File stored in ' + resourcePath)
-        header.addLinks(res, linkHeader)
+        header.addLinks(res, links)
```
dan-f: Why do we send back the link headers in the response?

lib/ldp.js
```diff
@@ -184,7 +185,7 @@ class LDP {
     .then(result => callback(null, result), callback)
   }
 
-  post (hostname, containerPath, slug, stream, container, callback) {
+  post (host, containerPath, stream, { container, slug, extension }, callback) {
```
dan-f: Nitpick, but is there a reason to put only `container`, `slug`, and `extension` in the options object?

lib/ldp.js
```diff
@@ -17,6 +17,7 @@ var ldpContainer = require('./ldp-container')
 var parse = require('./utils').parse
 
 const DEFAULT_CONTENT_TYPE = 'text/turtle'
+const DEFAULT_EXTENSION = '.ttl'
```
dan-f: I think that `DEFAULT_CONTENT_TYPE` and `DEFAULT_EXTENSION` should be maintained in terms of a mapping from one to the other since they (in purpose) represent the same thing.

lib/ldp.js
```diff
@@ -478,27 +481,18 @@ class LDP {
...
-          // Otherwise, generate a new path
-          } else {
-            const id = uuid.v1().split('-')[ 0 ] + '-'
-            newPath = path.join(containerURI, id + slug)
-            resolve(ensureNotExists(newPath))
-          }
-        })
-      })
+  getAvailablePath (host, containerURI, { slug = uuid.v1(), extension }) {
+    function ensureNotExists (self, newPath) {
```
dan-f: I'd refactor this closure into a function fully determined by its arguments, that way it doesn't have to be redefined at runtime.

test/integration/http-test.js
```diff
@@ -608,6 +608,73 @@ describe('HTTP APIs', function () {
...
+        before(() =>
+          server.post('/post-tests/')
+                .set('slug', 'slug1')
+                .set('content-type', 'text/turtle; charset=utf-8')
+                .then(res => { response = res })
+        )
+
+        it('is assigned an extensionless URL', () => {
+          expect(response.headers).to.have.property('location')
+          expect(response.headers.location).to.match(/^\/post-tests\/slug1+$/)
```
dan-f: Do you want that `+` before the `$`? This test will accept `/post-tests/slug111111`

test/integration/http-test.js
```diff
@@ -608,6 +608,73 @@ describe('HTTP APIs', function () {
         })
     })
 
+    describe('content-type-based file extensions', () => {
+      // ensure the container exists
+      before(() =>
+        server.post('/post-tests/')
+          .send(postRequest1Body)
```
dan-f: `postRequest1Body`??

test/integration/http-test.js
```diff
@@ -608,6 +608,73 @@ describe('HTTP APIs', function () {
...
+          .set('content-type', 'text/turtle')
+      )
```
dan-f: Make sure to have an `after` hook which cleans up!

lib/handlers/post.js
RubenVerborgh: Which part, the single line or the extraction of headers in general?

lib/handlers/post.js
RubenVerborgh: I have no idea; just preserving what was there…

lib/handlers/post.js
RubenVerborgh: Actually this line is really straightforward. Extension is a hash with non-empty arrays of extensions; so either it's in there and I return the first, or it isn't and I return the empty string. Testing this line in isolation would be just testing the `extensions` data structure, so that doesn't seem really meaningful (unless you meant something else than the line).

RubenVerborgh: Then again, I suppose I could put both lines in a `util` library. Is this what you meant?

lib/ldp.js
RubenVerborgh: You're right; this is going to get dropped though as per @timbl's comment.

lib/ldp.js
RubenVerborgh: Yes: the first three seem like elementary objects ("put this stream on this path on this host"), whereas the other are of lesser importance. Would that make sense?

lib/ldp.js
RubenVerborgh: It's a bit more tricky, since the `ensureNotExists` function is calling itself recursively. Basically, the flow is that it receives a filename, and if that name exists, generates another name (which it needs to check again).

Note that "redefined" is very likely not how V8 interprets this. The function exists once as machine code and will just be populated with different arguments.

test/integration/http-test.js
RubenVerborgh: I know… just reusing what the other tests do. The whole test file needs refactoring.

test/integration/http-test.js
RubenVerborgh: Good point, it's further down.

test/integration/http-test.js
RubenVerborgh: I didn't! Good catch.

lib/handlers/post.js
dan-f: Ah I was thinking that the core of this PR was "given a request, how do we infer the mime type and extension", so I was thinking that the logic for determining `mimeType` and `extension` could live in their own functions. Not a big deal, since this is covered already in tests.

lib/handlers/post.js
dan-f: Hope that my response above clarifies my thought.

lib/ldp.js
dan-f: I was trying to think this through. I wish I knew better how JIT compilers work (/me makes note to do more reading on this subject). My hypothesis was that, because `ensureNotExists` depends on the external values of the parameters to `getAvailablePath` (`host`, `containerURI`, `slug`, and `extension`), it cannot be translated into machine code, because its very definition is determined at runtime.

That, or I have no idea what I'm talking about 😆

dan-f:
@RubenVerborgh - please merge when ready.

RubenVerborgh:
Updated PR to address @dan-f's review, and to address @timbl's comment that extensions should always be added (even in the case of Turtle).

PR #586, opened 2017-09-09
Give hosts entry each their own line fixes #585

### LINKED ISSUE #585: test fails on linux when using comma separated localhosts in /etc/hosts
The test in test/account-creation.js fails on ubuntu when using a comma separated list in ubuntu

Using one entry on each line or space separated seems to fix the issue, for me at least

Simple fix would be to change the README to suggest putting one entry on each line, which I would expect to be OS independent

melvincarvalho:
Give hosts entry each their own line which fixes a test on ubuntu

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,9 @@ Edit your `/etc/hosts` file, and append:
 
 ```
 # Used for unit testing solid
-127.0.0.1 nic.localhost, tim.localhost, nicola.localhost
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
I already found it weird this syntax worked; apparently only on OSX. Thanks!

RubenVerborgh:
(Accidentally merged this to `master`, but I suppose it is not a problem.)

PR #594, opened 2017-09-14
Create new landing page

### LINKED ISSUE #579: Dead-end UX upon landing "in" Solid
dead_end_ux

"Welcome to the account of `<url>`" doesn't help me discover what I can *do* with a solid account.

"Solid User Profile" would make sense if I were viewing a solid user profile.

I realize we don't yet have a dashboard, but I think there's a bare minimum experience we should offer on the home page:

- log out
- affordance to edit my profile
- file browser
- example apps

RubenVerborgh:
Fixes #579

### DIFF
--- a/default-templates/new-account/index.html
+++ b/default-templates/new-account/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{#if name}}{{name}} -- {{/if}}Solid Home</title>
+  <title>{{#if name}}{{name}} &ndash; {{/if}}Solid Home</title>
   <link rel="stylesheet" href="/common/css/bootstrap.min.css">
 </head>
 <body>
@@ -11,7 +11,7 @@
   <nav class="navbar navbar-default">
     <div class="container-fluid">
       <div class="navbar-header">
-        <a href="/" class="navbar-brand">{{#if name}}{{name}} -- {{/if}}Solid Home</a>
+        <a href="/" class="navbar-brand">{{#if name}}{{name}} &ndash; {{/if}}Solid Home</a>
       </div>
     </div>
   </nav>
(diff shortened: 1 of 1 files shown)


dan-f:
Nice. A few broad thoughts:

- I think naming is important, and I'm wondering if we've got consensus on calling this "Solid Homepage". "Homepage" (page) implies something static, which this is, though in the future it won't be. I think it makes sense for this release, but wanted to raise the flag now.
- "Solid Homepage of {{name}}" sounds a bit odd to me. I think "{{name}}'s Solid Homepage" could make more sense, but then we run into issues with names that end in "s".
- There's definitely a big branding decision here. For now, "Solid" makes sense, but I think in the future we should think about separating the user-facing identity of the solid server from the identity of the broader solid architecture. Perhaps we could say something like "{{name}}'s Solid"


And some specific thoughts:

- Linking to data is cool, but people are going to need to know how to *do* things. Let's see if we can update the profile builder (for example) with the new auth, so that we can afford the user an opportunity to build their profile, not just see the static version of it.
- I think that linking to folders can be useful, but only if the databrowser is sufficiently discoverable. I'm concerned that in the `inbox` example, people won't know what do do with it. Again, a link to the inbox app will let people do things with their inbox.

We should chat about timeline. I realize most of our apps have to be updated for the new auth.

timbl:
Yes, home page is, while not static, a welcome for others rather than for oneslef. (Actually originally it was the start page was (and is) the initial you set you browser to, which does in fact work for this, but that is rather been usurped in common parlance as a public page representing you to others)

agree you can't do the 's.   

If someone sell is siing it you want to emphasize the person   Dan Friedman (on Solid)  vs  Control Center (Dan Friedman)  where the  bit son parens may be in a top right corner etc.    
- Solid home
- Solid stuff
- Dashboard
- Nexus
- Control Panel
- Start  (no, windows..)
- Home
Maybe a big i18n picture of a house.
Maybe make a RDF class for it like Home page and put in labels in diff languages and later work towars selecting the right one from language prefs.

RubenVerborgh:
The _`s_  thing we can avoid with something like "Jane Doe – Home" I suppose.

PR #599, opened 2017-09-15
Apply ACL also to authProxy

### LINKED ISSUE #595: Make .acl also apply to authProxy
If we use authProxy on a path like `/sensors/` then the regular .acl rules should apply.

RubenVerborgh:
Closes #595.

### DIFF
--- a/test/integration/auth-proxy-test.js
+++ b/test/integration/auth-proxy-test.js
@@ -18,7 +18,7 @@ describe('Auth Proxy', () => {
 
       // Set up Solid server
       server = ldnode({
-        root: path.join(__dirname, '../resources'),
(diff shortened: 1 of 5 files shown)


test/integration/auth-proxy-test.js
```diff
@@ -36,18 +37,100 @@ describe('Auth Proxy', () => {
...
+
+    describe('responding to GET', () => {
+      describe('for a path with read permissions', () => {
+        let response
+        before(() =>
+          request(server).get('/server/a/r')
+            .then(res => { response = res })
+        )
+        it('returns status code 200', () => {
+          expect(response).to.have.property('statusCode', 200)
```
dan-f: `expect(response.statusCode).toBe(200)`

Issue #604, opened 2017-09-17
Error: cannot find module '../package,json'


jdh0139:
1. os: Linux, Ubuntu 16.04.3 LTS
2. node: 6.11.3
    npm: 3.10.10
3. solid server install command.
    sudo npm install -g solid-server
4. The log given by the failure.
jdh0139@jdh0139-VirtualBox:~$ solid init
module.js:471
    throw err;
    ^
Error: Cannot find module '../package.json'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at getVersion (/usr/lib/node_modules/solid-server/bin/lib/cli.js:24:25)
    at startCli (/usr/lib/node_modules/solid-server/bin/lib/cli.js:7:19)
    at Object.<anonymous> (/usr/lib/node_modules/solid-server/bin/solid:3:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)

RubenVerborgh:
Tried the same steps (with Node 8) and couldn't reproduce.

dmitrizagidulin:
I suspect this has to do with a global install of ldnode. So line 24 in cli.js is breaking -- it likely needs to use path.join() instead of `../`

melvincarvalho:
#606 has been merged and should fix this.  Please reopen if this issue persists.

PR #606, opened 2017-09-19
Fix require of package.json for global installs

dmitrizagidulin:
Fixes issue #604

### DIFF
--- a/bin/lib/cli.js
+++ b/bin/lib/cli.js
@@ -2,6 +2,7 @@ const program = require('commander')
 const loadInit = require('./init')
 const loadStart = require('./start')
 const { spawnSync } = require('child_process')
+const path = require('path')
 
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
cc @timbl / @melvincarvalho  for review

bin/lib/cli.js
```diff
@@ -21,7 +22,7 @@ function getVersion () {
     return stdout.trim()
   } catch (e) {
     // Obtain version from package.json
-    const { version } = require('../package.json')
+    const { version } = require(path.join(__dirname, '../../package.json'))
```
melvincarvalho: Minor tip:  did you know you can do 

path.join(__dirname, '..', '..', 'package.json')

instead of

path.join(__dirname, '../../package.json')

PR #607, opened 2017-09-20
Added files and instructions to build and run with docker

retog:
With Docker one can install and node-solid-server without having to install node and npm. The resulting Docker image can easily be deployed to servers or published on dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM node:6.11.1-onbuild
+EXPOSE 8443
+RUN cp config.json-default config.json
+RUN openssl req \
+    -new \
+    -newkey rsa:4096 \
+    -days 365 \
+    -nodes \
+    -x509 \
+    -subj "/C=US/ST=Denial/L=Springfield/O=Dis/CN=www.example.com" \
+    -keyout cert.key \
+    -out cert.pem
+CMD npm run solid start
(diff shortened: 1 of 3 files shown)


retog:
Just noticed that a similar PR #327 has been there for more than a year. This PR is slightly different in that it defaults to multiuser and creates a self-signed certificate when building. But given that the other PR has been there for so long I guess there's little interest in this feature.

melvincarvalho:
@retog No, it is interesting!  Would definitely like to run via docker.  I think with the focus on getting 4.0.0 released some PRs may have fallen of the radar.  Is a good time to revisit.

melvincarvalho:
I'm happy to merge this, as it looks the most up to date.  

@rhiaro thanks for your previous work on this.  Would you like to review or make changes before merging?

melvincarvalho:
Just a couple of comments

- @rhiaro's PR had a 10 year expiry vs 1 year for the cert -- personally I like to go with 100 years for self signed
- the other difference is a 4096 bit RSA key vs 2048 bit.  While the newer 4096 is recommended, the majority of RSA keys are 2048 and well tested with tooling, I have had occasional issues with 4096 bit keys

Other than those minor comments, this seems to look pretty good!

retog:
Let's encrypt should support wildcard certs very soon, the it would be nice to have this docker taking care of getting actually signed certs. So I would address all cert issues when we can add letsencrypt support. In the meantime this docker is better than no docker at all.

Issue #436, opened 2016-08-29
Command line args, init questions, config file and --help must be consistent


timbl:
Currently the solid start --help gives you a list of command line options not all of which work. Some like --no-webid with in the config file but not in the args.  Be kind to users - be consistent.

Currently the init questions force you to give paths to SSL certs even if you are not using webid (which is this program's very obscure way of saying not use HTTPS!)   If the answer to "Are you using HTTPS and using webids?" is false, init should skip the SSL cert path questions.

The command line should have an argument to point to the config file, and the help should explain where the default config file is.

PR #613, opened 2017-10-18
+ command like arg: --config-file [./config.json]

ericprud:
This added arg allows you to override "`./config.json`". It's kinda lame in that it doesn't actually use the default 'cause config.json is read before `require('commander').parse(process.argv)` is run so you have to include the default a la:
```
let configFile = argv['configFile'] || './config.json'
```

### DIFF
--- a/bin/lib/start.js
+++ b/bin/lib/start.js
@@ -24,8 +24,9 @@ module.exports = function (program, server) {
 
   start.action((opts) => {
     let argv = extend({}, opts, { version: program.version() })
+    let configFile = argv['configFile'] || './config.json'
 
-    fs.readFile(process.cwd() + '/config.json', (err, file) => {
+    fs.readFile(configFile, (err, file) => {
       // No file exists, not a problem
       if (err) {
         console.log(colors.cyan.bold('TIP'), 'create a config.json: `$ solid init`')
--- a/bin/lib/options.js
+++ b/bin/lib/options.js
@@ -53,6 +53,12 @@ module.exports = [
     prompt: true
   },
   {
+    name: 'config-file',
+    question: 'Path to the config file (for example: ./config.json)',
+    default: './config.json',
+    prompt: true
+  },
+  {
     name: 'db-path',
     question: 'Path to the server metadata db directory (for users/apps etc)',
     default: './.db',


bin/lib/start.js
```diff
@@ -24,8 +24,9 @@ module.exports = function (program, server) {
 
   start.action((opts) => {
     let argv = extend({}, opts, { version: program.version() })
+    let configFile = argv['configFile'] || './config.json'
```
melvincarvalho: Might want to use path.join here with __dirname is that seems to be consistent with much of the code base

e.g. https://github.com/solid/node-solid-server/blob/dea121afbb92dd8e9ef8929d1d09246cd5ba5180/test/integration/capability-discovery-test.js#L12

melvincarvalho:
Generally, looks good!

I think also solves an issue raised in 

https://github.com/solid/node-solid-server/issues/436

bin/lib/start.js
ericprud: I'm happy to do that; was just making a minimal change to master. Does that buy us anything over just a relative path to config.json? I made pretty much the reverse change in shex.js when I made it able to load examples from a URL I suspect the only change is for node-solid-server is in debugging output.

melvincarvalho: No, looks good, happy to merge!

dmitrizagidulin:
re relative path - we constantly run into trouble with relative paths with people who install solid server globally (npm i -g)

melvincarvalho:
@dmitrizagidulin good point!  However the defaults for data and db are already ./data and ./.db respectively, right?

So ./config as a default seems consistent, for a quick change.  If I got that right.

We could perhaps cleanup the config a bit more later, e.g. in line with https://github.com/solid/node-solid-server/issues/436, alphabetizing the help etc.

ericprud:
Not necessarily critical path for this PR, but what *is* the right behavior for servers installed with `-g`? Should their be platform-specific detection of whether it's installed in a user vs. non-user directory and in the latter case default to `/usr/local/etc` (and some equiv on windows)? This affects both install (defaults) and runtime (where to look for config.json).

melvincarvalho:
@ericprud would you mind checking that the http options still works after this patch?

there was an issue raised : https://gitter.im/solid/node-solid-server?at=5ab044e8f3f6d24c68934a79

PR #625, opened 2018-01-17
Update node-webid so server does not crash with no content type fixes

### LINKED ISSUE #623: Add test for WebID with no content type
Occasionally a non standard WebID without a content type will try to authenticate on node-solid-server

There is some [evidence](https://gitter.im/solid/chat?at=5a563275d0514c785b137c47) that this could cause the server to crash

It would be helpful to add a test to capture this case, and it should return an HTTP 401 error code, without causing the server to go down

melvincarvalho:
#623

Doesnt quite add the test but if fixes the issue of server crashing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "uuid": "^3.0.0",
     "valid-url": "^1.0.9",
     "vhost": "^3.0.2",
-    "webid": "^0.3.7"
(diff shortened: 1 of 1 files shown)


timbl:
This branch has a new dependency 
global-tunnel-ng 
?

PR #637, opened 2018-02-23
Fix OidcManager / delayBeforeRegisteringInitialClient param

dmitrizagidulin:
Work done in Dublin with @pmcb55


melvincarvalho:
Looks good but where is this param used?

PR #640, opened 2018-02-26
Update databrowser.html to latest mashlib

RubenVerborgh:
The latest version of the public mashlib at https://linkeddata.github.io/mashlib/dist/mashlib.min.js is no longer compatible with the old mashlib API. As such, databrowser.html has to change.

This PR is a combination of the current databrowser.html at timbl.com and https://github.com/linkeddata/mashlib/blob/c87ca0ac4c71177d29efd4a68ea105650401f0aa/dist/databrowser.html.

### DIFF
--- a/static/databrowser.html
+++ b/static/databrowser.html
@@ -10,16 +10,22 @@
     <script type="text/javascript" src="https://linkeddata.github.io/mashlib/dist/mashlib.min.js"></script>
     <script>
       document.addEventListener('DOMContentLoaded', function () {
-        var UI = Mashlib
+        const panes = require('mashlib')
+        const UI = panes.UI
 
-        UI.rdf.Fetcher.crossSiteProxyTemplate = document.origin + '/proxy?uri={uri}'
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
Problem (which also existed with the previous version): [`document.origin` currently only has limited browser support](https://bugzilla.mozilla.org/show_bug.cgi?id=931884) (Chrome only)

timbl:
Looks good to me and important.   Maybe in future we make the databrowser.html file an export of dome sort from mashlib?

RubenVerborgh:
> (where can I find a list of github things waiting for my review?)

Right here: https://github.com/solid/node-solid-server/pulls/review-requested/timbl

> Maybe in future we make the databrowser.html file an export of dome sort from mashlib?

Yes, and/or replace the JavaScript by a single line, such as

```JavaScript
Mashlib.init({ pane: 'outliner', subject: location.href, element: 'outline' })
```

PR #643, opened 2018-03-08
Add ResourceMapper class for URL/file mapping

RubenVerborgh:


RubenVerborgh:
@melvincarvalho Thanks and yeah, did an iterative review process with @timbl here 😄 
A couple of extra test cases coming up, and this one will be good.

Issue #571, opened 2017-09-02
Fragment identifier on main page URI is stripped in login process


timbl:
e.g you open a new Chrome, go to incognito mode, and load:

https://timbl.com/timbl/2017/Solid/HighLevelIssues/index.ttl#this

You end up getting to

https://timbl.com/timbl/2017/Solid/HighLevelIssues/index.ttl

This is only a problem the first time, because of all the redirects, not later when the login state has been stored.
Seems to be a problem with eith TLS or OIDC forms of login.   I thought @RubenVerborgh fixed this, at least for OIDC.

RubenVerborgh:
So the previous bug was #562 and involved a fix in the general 401 response. I'll check what's going on.

dmitrizagidulin:
We fixed the server side redirect, but not for cases where the auth process is initiated from witnin a mashlib app. Will post details shortly.

dmitrizagidulin:
The fix in https://github.com/solid/node-solid-server/pull/562 addressed a slightly different use case.

In this case, however, the hash fragment is being lost - first during the oidc identity provider's authentication response (the provider overwrites the hash fragment with auth response credentials), and second when the auth client parses the response and clears the credentials from the uri.

The upstream issues to fix are:

- [ ] [`oidc-op` #38-Preserve redirect_uri hash fragment during implicit flow AuthenticationResponse](https://github.com/anvilresearch/oidc-op/issues/38) and
- [ ] [`solid-auth-client` #28-Preserve original redirect_uri hash fragment when parsing authentication response](https://github.com/solid/solid-auth-client/issues/28)

RubenVerborgh:
Given that https://github.com/solid/solid-auth-client/issues/28 is done, we should be able to close this.

I'm not sure that we need https://github.com/solid/solid-auth-client/issues/28, as the hash is added by solid-auth-client. @dmitrizagidulin, do you think this is still necessary?

We might need to be careful to avoid both components trying to add the same hash. It they do, we might want to lock oidc-op to 0.3.0 for now.

dmitrizagidulin:
Yeah, i think this issue is now resolved and can be closed. And no, don't worry, the oidc-op component will not interfere w the auth client restoring the hash.

RubenVerborgh:
Okay, let's wait for #590 to test.

RubenVerborgh:
Can be tested after #593, and closed if succesful.

dan-f:
Fixed in #593

[...]

PR #648, opened 2018-03-23
Send full returnToUrl to oidc-auth-manager

RubenVerborgh:
Addresses the node-solid-server part of #571.

oidc-auth-manager still loses the URL we send, so fixes on that side are needed to fully solve #571. Tracking that in https://github.com/solid/oidc-auth-manager/issues/17.

### DIFF
--- a/default-views/auth/select-provider.hbs
+++ b/default-views/auth/select-provider.hbs
@@ -28,5 +28,14 @@
     <button type="submit" class="btn btn-primary" id="discover">Submit</button>
   </form>
 </div>
+<script>
+  // Temporary solution to preserve return URL
+  // until https://github.com/solid/oidc-auth-manager/issues/17 is resolved
+  const locationUrl = new URL(location)
+  const returnToUrl = locationUrl.searchParams.get('returnToUrl')
+  localStorage.removeItem('returnToUrl')
+  if (returnToUrl && new URL(returnToUrl).host === locationUrl.host)
+    localStorage.setItem('returnToUrl', returnToUrl)
+</script>
 </body>
 </html>
--- a/default-views/auth/select-provider.hbs
+++ b/default-views/auth/select-provider.hbs
@@ -20,12 +20,10 @@
   {{/if}}
 </div>
 <div class="container">
-  <form method="post" action="/api/auth/select-provider">
+  <form method="post">
     <div class="form-group">
       <label for="webid">WebID or server URL:</label>
-      <input type="text" class="form-control" name="webid" id="webid"
-             value="{{serverUri}}" />
-      <input type="hidden" name="returnToUrl" id="returnToUrl" value="" />
+      <input type="text" class="form-control" name="webid" id="webid" value="{{serverUri}}" />
     </div>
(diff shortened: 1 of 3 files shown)


default-views/auth/select-provider.hbs
```diff
@@ -20,12 +20,10 @@
   {{/if}}
 </div>
 <div class="container">
-  <form method="post" action="/api/auth/select-provider">
+  <form method="post">
```
RubenVerborgh: Always POST to self, so we don't lose query string.

default-views/auth/select-provider.hbs
```diff
@@ -20,12 +20,10 @@
...
     <div class="form-group">
       <label for="webid">WebID or server URL:</label>
-      <input type="text" class="form-control" name="webid" id="webid"
-             value="{{serverUri}}" />
-      <input type="hidden" name="returnToUrl" id="returnToUrl" value="" />
```
RubenVerborgh: This one was always empty, so rather pointless. Instead, the query string should be used.

lib/handlers/error-pages.js
```diff
@@ -163,15 +163,12 @@ function redirectToSelectProvider (req, res) {
...
 
-  let currentUrl = util.fullUrlForReq(req)
-  req.session.returnToUrl = currentUrl
-
-  let locals = req.app.locals
-  let loginUrl = locals.host.serverUri +
-    '/api/auth/select-provider?returnToUrl=' + currentUrl
+  const locals = req.app.locals
+  const currentUrl = util.fullUrlForReq(req)
+  const loginUrl = `${locals.host.serverUri}/api/auth/select-provider?returnToUrl=${encodeURIComponent(currentUrl)}`
```
RubenVerborgh: Escaping was definitely important here (for query strings and such).

lib/handlers/error-pages.js
```diff
@@ -163,15 +163,12 @@ function redirectToSelectProvider (req, res) {
   res.status(401)
   res.header('Content-Type', 'text/html')
 
-  let currentUrl = util.fullUrlForReq(req)
-  req.session.returnToUrl = currentUrl
```
RubenVerborgh: We can't rely on session, because the server cannot see the hash.

dmitrizagidulin: This fixed mechanism (not relying on session) will still work on direct requests by the browser (like, requesting an image or a pdf file or whatever), not mediated by a client?

RubenVerborgh: If we end up in this code, we are in the process of rendering an HTML error page. So if that's the behavior we exhibit in response to a browser directly requesting an image (I hope not), it's broken for another reason already :smile:

dmitrizagidulin: Ahh, I see. Makes sense.

RubenVerborgh:
@timbl e7be5dd8b9325de39bd92f51f63b65c9ff01b9bf adds a temporary solution to preserve the full return URL (with hash) when logging in with WebID-TLS. Could you test `fix/redirect-hash` branch on timbl.com?

PR #650, opened 2018-04-19
Restore WebID-TLS login in OIDC mode

RubenVerborgh:
This is a kludge to make WebID-TLS login work again when the server is running in OIDC mode. The problem I observed is as follows:

- When I would click the "Log in with WebID-TLS" button, the browser would send an AJAX `POST` request to `/login/tls`.
- That `POST` request would trigger re-negotiation on the server and ask for a client certificate.
- The server would extract my WebID from my client certificate (which is `https://ruben.verborgh.org/profile/#me`).
- The server would:
   1. incorrectly assume that I want to be directed to my own pod instead of the redirect URI (see #648);
   2. incorrectly assume that my home pod is `https://ruben.verborgh.org/`.
- Hence, the server would send a 302 response redirecting to `https://ruben.verborgh.org/`.
- Therefore, the browser (still executing the AJAX request) would redirect to `https://ruben.verborgh.org/` (without letting the AJAX script know), but that would fail because the AJAX request uses `credentials: 'include'`, which doesn't work there (nor should it).
- Since the redirected request would fail, the whole AJAX call would, and the success callback would never get executed.

As you can see from the above accumulation, the root cause of the problem is that the whole login flow is messy and, in my opinion, reeds a deeper rewriting.

For a more immediate effect, I wrote a kludge that fixes the symptoms as follows:
- Return a 204 in case of an external WebID rather than a redirect, which causes the AJAX request to succeed.

The redirect is still the wrong one; we depend on #648 to address that.

### DIFF
--- a/default-views/auth/login-tls.hbs
+++ b/default-views/auth/login-tls.hbs
@@ -30,7 +30,10 @@
         return response
       })
       .then(function(response) {
-        window.location.href = response.url
+        // TODO: redirect to proper location stored in hidden field redirect_uri
+        // depends on https://github.com/solid/node-solid-server/pull/648
+        // and https://github.com/solid/oidc-auth-manager/issues/17
+        window.location.href = '/'
       })
   })
 </script>
--- a/lib/requests/login-request.js
+++ b/lib/requests/login-request.js
@@ -30,6 +30,7 @@ class LoginRequest extends AuthRequest {
     super(options)
 
     this.authenticator = options.authenticator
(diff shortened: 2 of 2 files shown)


melvincarvalho:
Just cc'ing @dmitrizagidulin on this as he design a lot of this work flow ...

dmitrizagidulin:
@RubenVerborgh apologies for late reply -- looks great to me!

PR #655, opened 2018-05-01
Use block level scope instead of function level scope

ninachaubal:
Refactors code in ./lib to use block level scope (const, let) instead of function level scope (var).


ninachaubal:
Rebased to develop. Could you take another look at this?

RubenVerborgh:
Perfect, thanks!

PR #666, opened 2018-05-06
Prevent double slashes in the "service" link header

### LINKED ISSUE #634: OPTIONS returns two slashes in the "service" link header instead of one
### Description
OPTIONS returns two slashes in the "service" link header instead of one

#### Steps to Reproduce

    curl -i -X OPTIONS https://melvincarvalho.solidtest.space/

#### Expected Results
    Link: <https://melvincarvalho.solidtest.space/.well-known/solid>; 
    rel="service", <https://solidtest.space>; rel="http://openid.net/specs/connect/1.0/issuer"

#### Actual Results
    Link: <https://melvincarvalho.solidtest.space//.well-known/solid>; 
    rel="service", <https://solidtest.space>; rel="http://openid.net/specs/connect/1.0/issuer"

Note: two slashes

ninachaubal:
Resolves #634

### DIFF
--- a/lib/handlers/options.js
+++ b/lib/handlers/options.js
@@ -1,5 +1,6 @@
 const addLink = require('../header').addLink
 const utils = require('../utils')
+const url = require('url')
 
 module.exports = handler
 
@@ -22,7 +23,7 @@ function linkAuthProvider (req, res) {
 }
 
 function linkServiceEndpoint (req, res) {
-  let serviceEndpoint = `${utils.getFullUri(req)}/.well-known/solid`
+  let serviceEndpoint = url.resolve(utils.getFullUri(req), '.well-known/solid')
   addLink(res, serviceEndpoint, 'service')
 }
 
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
Thanks @ninachaubal! :)

Could you by any chance add a unit test or something, to that function?

RubenVerborgh:
@ninachaubal Thanks! However, interpolation is not the problem, but the extra slash is. No need to perform an actual resolve, since we know the first part will end in a slash and the second part is a constant without special path expressions. Let's add a test and then we can merge.

ninachaubal:
Hey, sorry I couldn't get to this sooner. 

I agree that we don't need to perform an actual resolve here, but I'm not certain we can assume that first part will end in a slash. As far as I can tell, the first part will end in a slash if req.path ends in a slash and express doesn't make any guarantees about that. 

Would it be reasonable to edit utils.getFullUri() to guarantee an ending slash? Alternatively, we could use something like [url-join](https://www.npmjs.com/package/url-join) to ensure there aren't multiple slashes without doing a resolve.

On the tests, what's the best place to put them? Should I add a new file under test/unit for options or is there an existing test suite you would like me to add these to?

RubenVerborgh:
> I'm not certain we can assume that first part will end in a slash.

Resolve is perfect then, thanks!

> On the tests, what's the best place to put them? 

Looks like `test/integration/capability-discovery-test.js` is the place to go.

ninachaubal:
@RubenVerborgh, could you take another look at this PR?

RubenVerborgh:
Thanks!

PR #678, opened 2018-05-27
update deps

mediaprophet:
update package version no.#.


melvincarvalho:
We are in the process of upgrading @trust/oidc-op and rdflib

If node mailer is breaking something let's handle that separately

Issue #683, opened 2018-05-28
Remove bookmark list and profile editor from dashboard


melvincarvalho:
## Summary

Remove bookmark list and profile editor from dashboard.

For me both apps are completely broken.

The profile editor just gives a white screen when I click it.

The bookmark app on heroku takes a long time to load.  And says

Couldn't find your Mark installation
Unauthorized

![bookmark](https://user-images.githubusercontent.com/65864/40602873-7e63159e-625a-11e8-849f-e7edb2886dcb.png)

Apps can be added back when fixed or working apps can replace them.

melvincarvalho:
## Code

The relevant code is at : 

https://github.com/solid/node-solid-server/blob/cc7b5bcf70892d1bd6f5d046df97528f4c1fe743/default-templates/new-account/index.html#L60

## New Users

A simple deletion should be fine

## Existing Users

Will still have the old template.  It's possible that a script could replace the root index.html in user directories with an updated version.

However, the handlebars substitutions would require user specific variables.

- name
- webId

melvincarvalho:
[Suggestion](https://gitter.im/solid/node-solid-server?at=5b0bd87ea45f930a65de9fc0) from @timbl 

> Maybe we can make a pointer from there to a github md page on “what to do when you first get a solid+databrowser account"

I'll raise a new issue for this.

PR #684, opened 2018-05-28
Remove bookmark app and profile editor app addreses #683

melvincarvalho:


dmitrizagidulin:
Is the Friends app working?

melvincarvalho:
@dmitrizagidulin when I checked it was at least functional.  The other two completely failed, hence removal.

I think we can (and should) curate this page a few times, and perhaps allow users to regenerate it from the template.

PR #691, opened 2018-05-30
Update version of nodejs used in Dockerfile.

tmciver:


RubenVerborgh:
Thanks, but should be rebased such that only 56fc0d39e3c8b318e985ccfd1e2e8a325339d75e remains.

tmciver:
Oh, was I supposed to do this on the `develop` branch?  I can create a new PR for that.

Issue #694, opened 2018-06-01
Consider special handling for robots.txt


csarven:
If robots.txt doesn't exist, return a 404, instead of 401, and then its HTML meta redirecting to auth/select-provider (for login).

dmitrizagidulin:
Ah, good idea. 

Do we actually want to include a proper robots.txt, or..?

csarven:
As a general rule of thumb: only if there are paths that are public-read that should not be indexed or archived by other services. Note that, by suggesting that some path should not be crawlable would make proper archiving less effective. So, here the general assumption should be that, if public-read, then go ahead and index/archive.

Having said that, I would default to not having a robots.txt or an empty robots.txt. Return 404. The admin should make these decisions.

dmitrizagidulin:
@csarven - so, there's no good way to have special handling of `robots.txt` requests, in solid-server.

I've added PR #700, though, which adds empty/default robots.txt files to new account templates and server templates, with their corresponding .acls. So, that should cut down on the 401 error spam (which is what I assume prompted you to open this issue).

dmitrizagidulin:
Addressed by PR #700, closing.

PR #700, opened 2018-06-07
Add robots.txt and .acl to default account and server templates

dmitrizagidulin:
Addresses issue https://github.com/solid/node-solid-server/issues/694

(Also adds favicon.ico to the default multi-user server template as well).


default-templates/server/favicon.ico.acl
```diff
@@ -0,0 +1,15 @@
+# ACL for the default favicon.ico resource
+# Server operators will be able to override it as they wish
```
RubenVerborgh: I don't see any Control option being set though.

default-templates/server/robots.txt.acl
```diff
@@ -0,0 +1,15 @@
+# ACL for the default robots.txt resource
+# Server operators will be able to override it as they wish
```
RubenVerborgh: No Control option.

default-templates/server/favicon.ico.acl
dmitrizagidulin: Ok, so, that’s intentional. For the server templates, we don’t currently have any notion of giving identity to the server operator (like an admin user or whatever). So it’s unclear who to give write/control permissions to. 
More importantly, the assumption is that server operators will be able to edit this acl file directly. (via the file system)

PR #701, opened 2018-06-08
Updated pull request based on branch V4.1.0 from solid/node-solid-server

smalinin:


RubenVerborgh:
@smalinin Thanks, could you summarize the changes?

Note: lets make this a squash merge, because of the difficult history tail (with "sync" commits instead of merge commits).

RubenVerborgh:
Also, the build is currently broken because of

```
/home/travis/build/solid/node-solid-server/lib/requests/login-request.js:175:11: Infix operators must be spaced.
```

smalinin:
It looks, that only the next fixes were added

lib/create-app.js
```diff
@@ -169,7 +169,7 @@ function initWebId (argv, app, ldp) {
       // without permission by including the credentials set by the Solid server.
       const origin = req.headers.origin
       const userId = req.session.userId
-      if (!argv.host.allowsSessionFor(userId, origin)) {
+      if (req.path !== '/logout' && req.path !== '/goodbye' && !argv.host.allowsSessionFor(userId, origin)) {
```
dmitrizagidulin: Why this addition?

static/databrowser.html
```diff
@@ -9,25 +9,29 @@
...
+    // tabulator = {}
+
+    window.document.title = 'Loading ' + uri
+    var kb = UI.store
+    var subject = kb.sym(uri)
+    var outliner = panes.getOutliner(document)
+    outliner.GotoSubject(subject, true, undefined, true, undefined);
 
-        // Authenticate the user
-        UI.authn.checkUser()
```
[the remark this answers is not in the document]
smalinin: I have fixed the above.

lib/create-app.js
smalinin: The Logout didn't work on WebID-OIDC server, I added request to solid-auth-client

dmitrizagidulin: Makes sense. Let's remove this change, though, the logout cookie behavior will need to be handled separately, elsewhere.

smalinin: I have fixed chmods, may be it was changed my Samba server, it was changed for .acl som another files.

lib/models/account-template.js
```diff
@@ -92,7 +90,8 @@ class AccountTemplate {
       tlsCertWebid: userAccount.tlsCertificateWebid,
       tlsCertModulus: userAccount.tlsCertificateModulus,
       tlsCertExponent: userAccount.tlsCertificateExponent,
-      addOwlSameAs
+      addOwlSameAs,
+      serverUri: accountManager.host.serverUri
```
dmitrizagidulin: If the `accountManager` is only being used to get the `host` instance, maybe just pass in `host` directly? Instead of the full `accountManager`?

lib/requests/login-request.js
```diff
@@ -164,37 +164,42 @@ class LoginRequest extends AuthRequest {
    * @return {string}
    */
   postLoginUrl (validUser) {
-    if (this.clientId) {
+    let uri
+
+    if (this.authQueryParams['client_id']) {
```
dmitrizagidulin: Why the change from `this.clientId` to `this.authQueryParams['client_id']` here? Was `clientId` not being properly initialized?

lib/create-app.js
smalinin: Removed handle for `/logout` and `/goodbye`

lib/requests/login-request.js
smalinin: Fixed

lib/models/account-template.js
smalinin: Fixed

dmitrizagidulin:
@smalinin Thanks for the fixes! There's still 3 .acl files with `chmod` changes, but overall 👍 from me!

smalinin:
@dmitrizagidulin Ok, I see, it looks that in my prev PR to Node-solid-server, I changed chmod for these files to `rwxrwxr-x` mode, and this was missed in prev PR review stage. Now I set chmod to `rw-rw-r--` for all .acl files, that is the properly value.

dmitrizagidulin:
No prob. Thanks again!

RubenVerborgh:
Reminder that we should squash this into one commit, as 41d9bb5691a037c2942dacd1a0ffe2ecef89fcbb is problematic (not a merge commit, will mess with our history).

kidehen:
@RubenVerborgh @dmitrizagidulin ,

What does "Reminder that we should squash this into one commit, as 41d9bb5 is problematic (not a merge commit, will mess with our history)." mean in regards to:

1. Actions Items?
1. Action item ownership?

RubenVerborgh:
> Actions Items?

Whoever merges this, should squash merge, not create a merge commit.

> Action item ownership?

The merger.

RubenVerborgh:
Assigning this to @kjetilk for now, as a reminder we need to look into this.

RubenVerborgh:
@smalinin @kidehen We have discussed this, and @justinwb will be following up on this PR with you. Thanks for your patience.

smalinin:
@RubenVerborgh Basically I could create for from current state on solid:v4.1.0 , apply all changes above and create a new PR(it is not problem). Will it be better for you ?

RubenVerborgh:
@smalinin That would help. Note that a new PR is not needed for that, you can just force push to the branch (but I don't mind either).

kidehen:
@justinwb,

What's happening here?

justinwb:
@kidehen just left you a note on Gitter. Summation - we’ve got it on the backlog to merge as soon as we wrap a couple more priority items for the 5.0 release. One thing that would help speed things up a ton when we merge is if the contributors in that pull can go through and add as many comments as possible ahead of time so we’re not going back and forth with a ton of questions

smalinin:
@RubenVerborgh I have recreated PR https://github.com/solid/node-solid-server/pull/728  with fresh form and collect all changes to one commit.

RubenVerborgh:
Note, the branch has been renamed into feature/webid-delegation, since this will not be part of v4.1.

RubenVerborgh:
Following up in #728.

lib/create-app.js
RubenVerborgh: This addresses the symptoms, but not the cause, which should be addressed by https://github.com/solid/node-solid-server/pull/793

PR #702, opened 2018-06-11
Refactor LDP to use a modular store/backend

dmitrizagidulin:
This is an in-place refactoring of the LDP backend code, to be able to move towards the architecture outlined here: [Solid HTTP request handling flow](https://github.com/solid/solid-architecture/blob/master/server/request-flow.md).

To make the code review burden easier, it'll be done as a series of smaller PRs *into* this branch.

Addresses issue https://github.com/solid/node-solid-server/issues/483

PR #703, opened 2018-06-11
Extract printing of debug info from LDP constructor

dmitrizagidulin:
Extract the config diagnostics out of the LDP constructor and into its own function.


dmitrizagidulin:
(Part of ongoing PR #702)

PR #705, opened 2018-06-11
LDP constructor cleanup

dmitrizagidulin:
* Remove unused `extend` dep (obsoleted by Object.assign)
* Extract `initErrorPages()` out of LDP constructor

(Part of ongoing PR #702)

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -67,21 +67,6 @@ class LDP {
     this.initErrorPages(options)
   }
 
-  printDebugInfo () {
-    debug.settings('Server URI: ' + this.serverUri)
-    debug.settings('Auth method: ' + this.auth)
-    debug.settings('Db path: ' + this.dbPath)
-    debug.settings('Config path: ' + this.configPath)
-    debug.settings('Suffix Acl: ' + this.suffixAcl)
-    debug.settings('Suffix Meta: ' + this.suffixMeta)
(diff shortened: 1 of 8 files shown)


lib/ldp.js
```diff
@@ -44,67 +45,52 @@ const RDF_MIME_TYPES = [
...
-    if (!this.suffixAcl) {
-      this.suffixAcl = '.acl'
-    }
-    if (!this.suffixMeta) {
-      this.suffixMeta = '.meta'
+    this.root = options.root || process.cwd()
+    if (!this.root.endsWith('/')) { this.root += '/' }
+
+    if (this.corsProxy && !this.corsProxy.startsWith('/')) {
+      this.corsProxy = '/' + this.corsProxy
```
RubenVerborgh: So the above is fine, but when I code, I tend to have the rule that the length of the code should represent the complexity of the action. My suggestion would be:

```
this.root.replace(/\/*$/, '/');
this.corsProxy.replace(/$\/*/, '/');
```

dmitrizagidulin: Does the regex variant take care of the `if (this.corsProxy &&` guard?

(I must admit, though I like the conciseness of regex, since i'm less familiar with it, I find the more wordy if/then more readable :) )

RubenVerborgh: Ah no, doesn't have that guard. Leave it as-is then, readability is king.

PR #706, opened 2018-06-12
Refactor LDP API middleware

dmitrizagidulin:
Proposed LDP API refactor, based on @RubenVerborgh et al's [http request flow](https://github.com/solid/solid-architecture/blob/master/server/request-flow.md).

Please give feedback early & often, to make sure we're on the right track.

(Part of ongoing PR #702)


lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Checks to see if target exists
+         */
+        request = LdpHeadRequest.from({req, host})
+        break
+      case 'get':
+        /**
+         * Use cases to handle:
+         *  - LdpListContainerRequest, if target is an ldp container
+         *  - However, if Accept: header is HTML, check if index.html exists
```
dmitrizagidulin: It would be nice to be able to dispatch between "List container" and implicit "I'm requesting a container url but really get me its index.html/index.ttl" type of GET requests here, during parsing & constructing the Request object.

But that requires passing in a backend object (e.g. `LdpFileStore`), to check if said `index.html` exists. Which I was initially reluctant to do, in terms of separation of concerns (and only pass in the backend into the `request.handle()` method). But it seems like it might be a good idea to pass it in at this early stage.

lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+         *   - has `mkdir -p` semantics (creates missing container hierarchy)
+         *
+         * If target is a resource:
+         *   - writes the resource, always overwriting existing contents if any existed
+         *     (since we don't support conditional `If-None-Match` requests)
+         *   - has `mkdir -p` semantics (creates missing container hierarchy)
+         */
+        request = LdpPutRequest.from({req, host})
+        break
+      case 'post':
```
dmitrizagidulin: Currently, the target path for POST is required to be a container. If it's not a container, solid-server currently throws a 405 Method Not Allowed error.

Should we continue to implement this behavior? I'm thinking yes, since POST to a resource doesn't seem to make much sense.

My other question is - should the "must be a container" requirement be signaled by a 405 error, or would it be more appropriate to return a 400 error instead?
(If the former, we need to make sure we return the Allow header per spec - "The response MUST include an Allow header containing a list of valid methods for the requested resource.")

lib/api/ldp/api.js
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      // parse target, operation, (optionally) body, preferences, store authn
+      const request = LdpRequest.from({req, host: this.host})
+
+      // check that operation is permitted (throws error if not)
+      const permissions = await this.acl.allow(request)
+
+      // perform the operation and return a response
+      await request
+        .handle({res, permissions, store: this.store, fetch: this.fetch})
```
RubenVerborgh: Why are the permissions still passed?

lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+         *
+         *  - Use contents of `Slug` header, if provided and no resource with same
+         *    name already exists
+         *  - Use contents of Slug plus UUID, if resource with same name exists
+         *  - Generate a UUID if no Slug given
+         *
+         * Does NOT use `mkdir -p` semantics (does not create missing containers)
+         *
+         * Throws:
+         *  - 400 error if target doesn't end in a / (is not a container)
```
dmitrizagidulin: Actually, this doesn't happen currently (POST doesn't throw a 400 if target path does not end in a /). Since the dispatching between "create container" and "create plain resource" is done via the Link: type rel, the handler is lax about a missing / at the end (in fact, it adds it if it's not there).

lib/api/ldp/api.js
```diff
@@ -0,0 +1,73 @@
...
+   * @throws {Error} 409 If PATCH operation results in conflict
+   *
+   * @throws {Error} 406 If no appropriate representation found (for content type)
+   *
+   * @throws {Error} 405 If HTTP method not allowed / not implemented
+   */
+  async middleware (req, res, next) {
+    try {
+      // parse target, operation, (optionally) body, preferences, store authn
+      const request = LdpRequest.from({req, host: this.host})
```
RubenVerborgh: The extra host parameter feels a bit clumsy; it's also incomplete because the protocol is not considered. Maybe we should more generally have an `overrides` parameter or similar.

lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+
+  static from ({req, host}) {
+    let request
+
+    switch (req.method) {
+      case 'head':
+        /**
+         * Checks to see if target exists
+         */
+        request = LdpHeadRequest.from({req, host})
```
RubenVerborgh: Where do these come from? (I assume to be implemented still?)

Also, I wonder whether it's necessary to do the subclassing at this level, but it is an option. That said, we there's probably some common parsing to be done first.

lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+    this.req = options.req
+    this.authn = options.authn
+    this.parsedBody = options.parsedBody
+  }
+
+  static from ({req, host}) {
+    let request
+
+    switch (req.method) {
+      case 'head':
```
RubenVerborgh: Uppercase?

But also, I don't really like the `switch`. There should just be a `handlers` map somewhere, such that we can simply do `handler = handlers[req.method]`.

lib/api/ldp/ldp-request.js
```diff
@@ -0,0 +1,145 @@
...
+         *   - creates the container if none existed
+         *   - does nothing if container exists, does not delete/clear existing contents
+         *   - has `mkdir -p` semantics (creates missing container hierarchy)
+         *
+         * If target is a resource:
+         *   - writes the resource, always overwriting existing contents if any existed
+         *     (since we don't support conditional `If-None-Match` requests)
+         *   - has `mkdir -p` semantics (creates missing container hierarchy)
+         */
+        request = LdpPutRequest.from({req, host})
```
RubenVerborgh: The above comments seem to belong in `LdpPutRequest`.

lib/api/ldp/ldp-request.js
RubenVerborgh: Then maybe `target` should be parsed first before this decision happens. You should be able to use https://github.com/solid/node-solid-server/blob/v4.0.14/lib/resource-mapper.js for that.

lib/api/ldp/ldp-request.js
RubenVerborgh: Seems like a 405 case to me, that's the most specific one.

lib/api/ldp/api.js
dmitrizagidulin: how about if we call it `serverConfig` instead of `host`?

lib/api/ldp/ldp-request.js
dmitrizagidulin: Yeah, heh, I plan to move 'em as soon as those classes get implemented.

lib/api/ldp/api.js
dmitrizagidulin: I wasn't going to pass them initially. I added them because of the section in the architecture doc that said something like "after performing [a write] operation, check for a read permissions on that resource and return the contents if those are there, otherwise a 204 no content."

lib/api/ldp/api.js
RubenVerborgh: The name makes sense.

But actually, I think you just need to parse `target` first as follows:
- have a `resourceMapper` private field, which is an instance of `https://github.com/solid/node-solid-server/blob/v4.0.14/lib/resource-mapper.js` (this already contains all the overrides)
- set `const target = this.resourceMapper.mapUrlToFile(…)`
- pass `target` to the LDP request constructor.

Alternatively, the `resourceMapper` can be passed to the LDP request constructor.

lib/api/ldp/ldp-request.js
dmitrizagidulin: Aren't Express request methods in lowercase?

I was torn between doing a `switch()` statement and a handlers map. 

I think I went with it for those cases (like with the GET method) where additional logic may need to be done before figuring out a subclass, so that a simple map would not suffice.

Any particular reason for disliking switch, out of curiosity?

lib/api/ldp/api.js
dmitrizagidulin: Re 'private field' - heh heh, so, I don't really believe in private fields. I think Steve Yegge captures it perfectly with his essay: http://steve-yegge.blogspot.com/2010/07/wikileaks-to-leak-5000-open-source-java.html)

But that aside - so here's what's interesting about the `resourceMapper` (which I agree is useful).
It belongs to the *backend*, not to the API request, specifically only to the `LdpFileStore` (and not to an in-memory store or a Quad Store backend).

So the idea here is that the Request objects would only need to call `this.store.exists(resource)` and have the backend (if appropriate) worry about file mapping.

dmitrizagidulin: > Actually, I think that LdpApi should take an ldpProvider as argument.

That's exactly what the `store` argument is, it's an ldp provider.

dmitrizagidulin: Would you like it to be renamed to that?

lib/api/ldp/ldp-request.js
RubenVerborgh: Well, actual operation-level subclasses :smile:

These are request-level subclasses.
I would expect to have one `LdpRequest` which _has_ an operation, and operations can then have different subclasses.

Not much difference I guess, slightly different way of modeling.

dmitrizagidulin: What do they say about naming being the hardest problem in CS? :)
I think it would be clearer if these subclasses are named e.g. `LdpGetOperation`, instead of `LdpGetRequest` -- cause they're modeling operations, not http requests. We leave request land in that `switch()` statement.

lib/api/ldp/api.js
RubenVerborgh: Ah right. I don't mind the name too much, I just think that `host` and  `fetch` should be dropped as parameters of `LdpApi`; they are the responsibility of the store (since all `LdpApi` does is passing them). Law of Demeter.

RubenVerborgh: And there should just be no defaults for `store` and `acl`; the API shouldn't need to know how to construct them. All the API should do is send the five steps the right way.

So if I'm very critical/demanding, I think that AclApi should almost literally reflect https://github.com/solid/solid-architecture/blob/master/server/request-flow.md: get all the right variables, and pass them to other components. That's more or less all it should be doing. Then it's up to those components to get instantiated with what they need (server options etc.) to fulfill their task.

dmitrizagidulin: Ok. Will keep the `store` name.

I wasn't sure about passing in the `host` and `fetch` as separate parameters, either.
But they're not just needed in the store. The `host` and `fetch` are both needed by the ACL Checker. (And I was initially thinking that the `host` may be needed in the url parsing of `target`,  but maybe not).

I suppose I can not pass them in, and just require passing in an already-initialized `acl` checker object (that will already have them)

RubenVerborgh: So something like:

```
const target = this.targetParser.parse(request);
const operation = this.operationParser.parse(request);
const bodyParser = this.bodyParsers.find(request.headers['content-type']);
const body = bodyParser && bodyParser.parse(body);
```

dmitrizagidulin: > And there should just be no defaults for store and acl

Ok, copy that. Will remove :)

And hey, critical & demanding is perfect, that's what I want :) I'm happy to have LdpApi reflect the request-flow document (with some slight exceptions).

dmitrizagidulin: Ok, so let's take those one at a time.

> `const operation = this.operationParser.parse(request)`

Is the `operationParser()` meant to take the place of the `switch()` statement? Cause that's essentially what it is, an operations parsing & factory method. 

> `const target = this.targetParser.parse(request)`

This makes sense, though I'm somewhat unclear on if we need any complicated parsing for the target url. (Which is different from url-to-fs mapping, a domain of the backend).

> `const bodyParser = this.bodyParsers.find(request.headers['content-type'])`

I feel that this belongs in each operation subclass. Like, HEAD ops don't need parsing. Neither do PUTs and POSTS etc. But PATCH does.

lib/api/ldp/ldp-request.js
dmitrizagidulin: > Keeping specialized knowledge in one place. The class implementing the switch currently needs knowledge on the internals of the constructors it creates, spreading out this specific knowledge (especially if adding that additional logic).

So I think that's my point - the `switch()` statement *enables* specialized knowledge (as in the case of the GET operation - though that's not clear from the current code). (Though notice that it's not specialized knowledge of constructors, the subclass `from()` methods hide that knowledge).

Here, let me rename the `Ldp*Request` references to `Ldp*Operation`, to make the discussion clearer. (And take out the extra constructor params to LdpApi, as discussed). And add the illustrative code to the GET section of the `switch()`.

lib/api/ldp/api.js
RubenVerborgh: > Is the operationParser() meant to take the place of the switch() statement?

No:

>> The operation consists of the HTTP method, and a set of flags. Valid operations are GET, HEAD, OPTIONS, POST, PUT, PATCH, DELETE. Based on this method—and, in the case of PATCH, the parsed body—the read, append, write, and/or delete flags are assigned.

It's just parsing, really. (This shows that body should be parsed before operation though, so body can be passed as a param.)

> This makes sense, though I'm somewhat unclear on if we need any complicated parsing for the target url.

Complicated enough that node-solid-server got it wrong in a couple of places :-)

The idea is to have this in a small separate component, so we can test all parsings independently. This includes URLs with spaces, encoded characters, etc. etc.

> I feel that this belongs in each operation subclass.

We need to parse the body to determine the operation, and this is needed as input for the permissions checker. (And the operation subclasses do not [need to] know about permissions.)

lib/api/ldp/ldp-request.js
RubenVerborgh: > So I think that's my point - the switch() statement enables specialized knowledge (as in the case of the GET operation - though that's not clear from the current code).

Yes, but it spreads it out across two places: the `switch` statement and the `Ldp*Request` constructor. The class containing the `switch` has to know about each of the subclasses. That's knowledge it doesn't need, and ties the implementation of this class to each of the subclasses. That's coupling that can be avoided.

dmitrizagidulin: > Yes, but it spreads it out across two places: the switch statement and the Ldp*Request constructor.

No prob. Lemme replace the switch() with a mapper, let's see how it looks.

dmitrizagidulin:
@RubenVerborgh take a look at the latest iteration when you have a chance.

I'm getting stuck on the subject of body parsing (in trying to accommodate a lack of switch() statements :) ).

Specifically, here's the logic for parsers:

PATCH - uses a `text()` body parser for the request stream, then passes it to either a `application/sparql-update` or a `text/n3` parser.

PUT & POST - either uses a multipart parser (such as `busboy` for multiple file uploads), otherwise doesn't use a parser and pipes the request stream straight into a file stream.

All others (head / get / options / delete / copy) - no parsers.

So that suggests to me only two options:

1) Have a `switch()` type statement somewhere like `parseBody()`, that goes through the above logic. or

2) Pass the request read stream into the Operation subclass (and have the relevant operations do parsing as needed)

lib/api/ldp/api.js
```diff
@@ -0,0 +1,70 @@
...
+   * @throws {Error} 400 If invalid parameters (or error parsing request body,
+   *   for cases like PATCH requests)
+   *
+   * @throws {Error} 409 If PATCH operation results in conflict
+   *
+   * @throws {Error} 406 If no appropriate representation found (for content type)
+   *
+   * @throws {Error} 405 If HTTP method not allowed / not implemented
+   */
+  async middleware (req, res, next) {
```
RubenVerborgh: Maybe we can name this `handleRequest`?

lib/api/ldp/api.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const LdpOperation = require('./ldp-operation')
+
+class LdpApi {
```
RubenVerborgh: Naming: perhaps `LdpHttpHandler` or similar?

lib/api/ldp/api.js
```diff
@@ -0,0 +1,70 @@
...
+   * @throws {Error} 409 If PATCH operation results in conflict
+   *
+   * @throws {Error} 406 If no appropriate representation found (for content type)
+   *
+   * @throws {Error} 405 If HTTP method not allowed / not implemented
+   */
+  async middleware (req, res, next) {
+    try {
+      // parse target, operation, (optionally) body, preferences, store authn
+      const operation = LdpOperation.from(req)
```
RubenVerborgh: Given the comment, I'd expect
```
const { target, operation, body, preferences, authentication } = …
```

lib/api/ldp/api.js
```diff
@@ -0,0 +1,70 @@
...
+
+      // check that operation is permitted (throws error if not)
+      const permissions = await this.acl.allow(operation)
```
RubenVerborgh: Should take `target` and `operation`.

lib/api/ldp/api.js
RubenVerborgh: …but that's too much already. LDP shouldn't know about authentication, that seems the task of this class.

lib/api/ldp/ldp-operation.js
```diff
@@ -0,0 +1,143 @@
...
+  constructor (options) {
+    this.authn = options.authn
+    this.target = options.target
+    this.headers = options.headers
+    this.bodyStream = options.bodyStream
+  }
+
+  static async from ({req}) {
+    const { method, headers } = req
+    const authn = { webId: req.session.webId, credentials: {} } // <- bearer tokens go here
```
RubenVerborgh: LDP should definitely not know about this. We should get authentication out of there, and definitely the very specific dependency on a session variable. This kind of state should only be accessed in a single place.

lib/api/ldp/api.js
dmitrizagidulin: Sure, renaming...

lib/api/ldp/api.js
dmitrizagidulin: Sure, we can rename it `handleRequest` - calling it `middleware` just references the common Express terminology.

lib/api/ldp/api.js
dmitrizagidulin: Heh, well, the comment was just referencing the architecture doc :) I removed the comment tho. (It just says `// parse operation` now).

lib/api/ldp/ldp-operation.js
dmitrizagidulin: Sounds good.

dmitrizagidulin:
@RubenVerborgh 

> There's a third option, which is documented in the request flow: a lazy parser.

Ahh, I see, I didn't quite understand that part, initially.

It still feels like the wrong place to set up the parser. It's too early, there's method-specific parser logic later on (which is more complicated than just the content-type).

Do you not think it makes sense to pass in the body stream (instead of a body parser) into the Operation?

lib/api/ldp/api.js
dmitrizagidulin: Why so? The `target` is already contained in the `operation`.

RubenVerborgh:
First of all, let's ensure we use the same terminology.

When I say “operation”, I mean “the HTTP method, and a set of flags”, as defined in https://github.com/solid/solid-architecture/blob/master/server/request-flow.md. Concrete examples are `{ method: 'GET', read: true }` and `{ method: 'PATCH', append: true }`.

When you say “operation”, you seem to mean the entirety of target, body, authentication, preferences and (what I call) operation.

I don't mind which terminology we pick, but we should be clear about that.

> It still feels like the wrong place to set up the parser. It's too early

It is earlier than the current design, but this is on purpose.

The idea is that, in Step 2 of https://github.com/solid/solid-architecture/blob/master/server/request-flow.md, we parse everything necessary to determine the permissions. Then, in Step 3, we check the permissions. And in Step 4, we don't need to think about permissions anymore.

The method name by itself is insufficient to determine the permissions, as there are cases (specifically `PATCH`) where the name does not tell us whether read/write/append/delete permissions are needed.

We _could_ make an exception for `PATCH`, which is what we currently do in the server. However, this specific exception breaks the layering of node-solid-server, i.e., it is specifically because of `PATCH` that the authentication in the current implementation _cannot_ be a layer around LDP. Instead, we have to pass authentication all the way through to LDP, which leads to coupling, hinders testability, etc. It's messy, and we want to avoid that.

Therefore, https://github.com/solid/solid-architecture/blob/master/server/request-flow.md has been designed such that node-solid-server can be designed as a layered system. Which means a separate authentication layer around the LDP store, that this store does not need to know about. Which means that the request has to be parsed earlier. The philosophy is the request should be understood before we go to Step 3, and this does include parsing the body (to accommodate the specific case of `PATCH`—but if we make exceptions, we break the layered design).

And with the lazy design, we only parse when needed anyway, so we don't do too much work.

Bottom line: yes, we are parsing early, but this makes the overall design easier, with fewer dependencies, and hence more testable.

You are definitely free to disagree with that decision, and I am open to discussion as usual, but that discussion should probably happen on https://github.com/solid/solid-architecture/ if the goal of this PR is indeed to implement that document.

lib/api/ldp/api.js
RubenVerborgh: …but the implementation was not in sync with neither that comment nor the architecture doc. Added a clarification about terminology here: https://github.com/solid/node-solid-server/pull/706#issuecomment-396981652, because we seem to indicate different things with “operation”.

lib/api/ldp/api.js
RubenVerborgh: Not in the architecture's doc definition of “operation”, see https://github.com/solid/node-solid-server/pull/706#issuecomment-396981652

dmitrizagidulin:
@RubenVerborgh - ok, I'm in the process of absorbing your points about definition of iteration, and early parser selection.

lib/api/ldp/ldp-operation.js
```diff
@@ -0,0 +1,196 @@
...
+
+    if (glob.hasMagic(targetUrl)) {
+      // this is a Ldp Glob operation, return that
+    }
+
+    // check to see if target is a container (based on Link: type rel)
+    // ...
+    // if it is a container, check to see if index.html exists
+    const indexFile = url.resolve(targetUrl, '/index.html')
+    if (store.exists(indexFile)) {
```
dmitrizagidulin: @RubenVerborgh - the GET request dispatching logic here illustrates why I was thinking it makes sense to include the `store` (backend) param to `Operation.from()`.

lib/api/ldp/ldp-operation.js
```diff
@@ -0,0 +1,196 @@
...
+  /**
+   * Taken from utils.getBaseUri()
+   *
+   * @param req {IncomingRequest}
+   * @param host {SolidHost}
+   *
+   * @returns {URL}
+   */
+  static parseTargetUrl ({req, host}) {
+    const protocol = host.serverUri.replace(/:.*/, '')
```
dmitrizagidulin: @RubenVerborgh re our earlier discussion of whether or not the `host` param is needed - this usage of it while parsing the target is what I was thinking about.

lib/api/ldp/ldp-operation.js
RubenVerborgh: Mmm, I think the `from` is doing at lot here. [Constructors shouldn't do actual work](http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/) (and although a static method, `from` is acting as a constructor here).

I think the constructor should just create/parse the operation, and then we have a separate `operation.perform(store)` method.

The question probably is how much store-agnostic work an Operation can do.

lib/api/ldp/ldp-operation.js
RubenVerborgh: The Operation should not probably not be performing this task; the target URL is also needed by the permission checker. (Also relates to our different definitions of “operation” though, where yours includes target but mine doesn't.)

lib/api/ldp/ldp-operation.js
dmitrizagidulin: Ok, I also believe that constructors shouldn't do actual work. But as far as I know, factory methods that set up and initialize the various things to pass to a constructor, are a valid strategy. So the `from()` method is just a factory method.

Now, as far as whether a store is involved in *defining* the operation - so my point here is that without a store, it's impossible to determine the operation. (Specifically, to decide whether this is a List Container op or a Get Op).

dmitrizagidulin:
@RubenVerborgh 

Ok, so as far as parsing the body (or at least selecting a lazy parser) before performing the operation, that's fine. I still think it's too early, but maybe it'll be fine - I'll go with it, and see if I run into any trouble in implementation.

Re definition of operation - yeah, I think that's exactly it. Our definitions are differing. I think that even if we create a simple operation object (that's just a method and some flags), there will still be a need to have "operation" like subclasses with specific logic. Like, we still need to distinguish between a glob operation, a list container, or a get action. And that's only possible with several things involved (like a store, etc).

lib/api/ldp/ldp-operation.js
RubenVerborgh: > But as far as I know, factory methods that set up and initialize the various things to pass to a constructor, are a valid strategy. So the from() method is just a factory method.

It's never a black-and-white situation, but I perceive the static `from` method more as a de-facto constructor rather than a factory method:

1. I is a static method, and constructors are (kind of) static methods too. The idea of the factory pattern is precisely to not depend on static methods.
2. The idea of a factory is that users of the factory don't need to know how the factory creates its objects. Some other component has instantiated a factory with several constructor arguments, so when we call a factory method, we don't need to pass in lots of things over and over again (like protocol and host etc.). I.e., the nice thing of a factory is that it is an instantiated object, so we don't need to pass all the state on every method call.
3. There is no practical difference (from a caller's perspective) if we move the code currently in  `from` to the constructor.

> Now, as far as whether a store is involved in defining the operation - so my point here is that without a store, it's impossible to determine the operation. 

Depending on the definition of “operation”, you do have a valid point here. Perhaps should be updated in the architecture document.

RubenVerborgh:
> Re definition of operation - yeah, I think that's exactly it. Our definitions are differing. I think that even if we create a simple operation object (that's just a method and some flags),

I think we should rename in the architectural flow document. Perhaps instead of `operation`, should be `method` and `requiredPermissions`? And then another field, depending on whether it is a glob/container/…?

lib/api/ldp/ldp-operation.js
dmitrizagidulin: I wonder if the name `host` is misleading.. It doesn't refer to the url hostname, it's a host server config object instance. Should I rename it `serverConfig`?

RubenVerborgh: Perhaps, yes.

Also node that part of the relevant code and tests are here: https://github.com/solid/node-solid-server/blob/v4.0.14/lib/resource-mapper.js

ResourceMapper currently does two things: re-constructing the intended URL (with host and protocol), and mapping that URL to a file.

Since this first function is shared across all LDP stores, but that second function is unique to the file-based store, it might make sense to move that first code (and associated tests) to here.

dmitrizagidulin: Sounds good, lets move the relevant (url non-fs) functions in resource-mapper.js to solid-host.js ?

RubenVerborgh: Yes!

lib/api/ldp/api.js
```diff
@@ -0,0 +1,106 @@
...
+      operation.writeResponse({res, result})
+    } catch (error) {
+      next(error)
+    }
+  }
+
+  async operationFrom (req) {
+    const options = this.parseOperation(req)
+
+    const Operation = LdpOperation.BY_METHOD[options.method]
```
dmitrizagidulin: @RubenVerborgh ok, so this is the next iteration based on our discussion.

The factory code was move from `Operation.from()` static method, to this instance factory method on `ldp`.

Here's my main problem with this design, though (specifically, getting the operation by HTTP method).

It breaks down when it comes to the GET method. (And PATCH method to a certain extent.) But let's talk about the get.

For most methods, there's a 1:1 mapping of HTTP method to Operation. But for GET, it's more like 1 get method to 3-4 potential operations. Where does the logic dispatching between them belong?

dmitrizagidulin: Somewhere there needs to be a `switch()` statement :)

RubenVerborgh: > But for GET, it's more like 1 get method to 3-4 potential operations.

What are they?

dmitrizagidulin: So really, this line should be more like:

```js
if (method === 'get') {
  Operation = this.parseGetOperation(...)
}  else {
  Operation = LdpOperation.BY_METHOD[method]
}
```

melvincarvalho: Re : cases for GET (off the top of my head)

- Resource
- Container
- Globbing
- Streaming media ie byte ranges
- Data browser
- Conneg

RubenVerborgh: Wow, that's a lot more than I thought. Conneg is orthonogal though, but the rest of them seem individual cases indeed.

So to get back to the point then, I'd suggest `Operation = new LdpOperation.BY_METHOD[method]({ method, target, body, … })` or similar, such that `LdpGetOperation` is responsible for creating its own subtypes.

dmitrizagidulin: >  such that LdpGetOperation is responsible for creating its own subtypes.

Ah, wait tho - that would mean instantiating different subtypes in the constructor of LdpGetOperation.. Which is awkward, no? Maybe a static `from()` would work? :)

dmitrizagidulin: I suppose, part of this depends on whether we want to model "List container" / "Glob operation" / etc as separate LdpOperation subclasses...

RubenVerborgh: > Ah, wait tho - that would mean instantiating different subtypes in the constructor of LdpGetOperation.. Which is awkward, no?

Not necessarily. Can be factory too, with our without `from` :smile: The difference is purely syntactical.

> I suppose, part of this depends on whether we want to model "List container" / "Glob operation" / etc as separate LdpOperation subclasses...

Indeed. It's a manner of how we want to manage complexity. Could be through subclasses, could be through different methods in LdpGetOperation. It's not that big of a decision, I think. Important thing is that the API shouldn't know how to construct different kinds of operations, that knowledge should be with the operations themselves.

dmitrizagidulin: Sounds good :)

lib/api/ldp/ldp-operation.js
```diff
@@ -0,0 +1,196 @@
...
+ *  - "List Container" request, if target is an ldp container
+ *  - However, if Accept: header is HTML, check if index.html exists
+ *  - Need to also support similar use case if index.ttl exists
+ *  - "Glob pattern request" if glob magic chars are detected (ie `*`)
+ *  - Otherwise, plain Get request
+ *
+ *  @return {LdpGlobOperation|LdpListContainerOperation|LdpGetOperation}
+ */
+class LdpGetOperation extends LdpOperation {
+  static from (options) {
```
dmitrizagidulin: @RubenVerborgh ok, so, I'm not *super* thrilled with this factory method, but it at least roughly fleshes out our semantics.

RubenVerborgh: Indeed, it's quite alright!

lib/storage/ldp/ldp-file-store.js
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * @param target {LdpTarget}
+   *
+   * @returns {Promise<LdpFileResource>}
+   */
+  async resource (target) {
+    const filePath = await this.mapper
+      .mapUrlToFile({
+        url: target.url, contentType: target.mediaType(), createIfNotExists: true
```
dmitrizagidulin: @RubenVerborgh Ok, so now we're at the stage of sketching the backend / LdpFileStore.
I suspect we'll need to refactor either ResourceMapper or LegacyResourceMapper.

The current behavior of `mapUrlToFile()` is to either 1) create the filepath using $ (non-legacy behavior) if `createIfNotExists: true`, or 2) Throw a 'File not found' error if `createIfNotExists: false` (which is always true if called via `LegacyResourceMapper`).

I'd like a slightly different semantics - to *not* throw a 'file not found' error, and just return "if the file existed this is what its path would be" in legacy mode.

RubenVerborgh: I hope no major refactoring of `LegacyResourceMapper` is necessary; I tried to design it so it can be plugged in right away.

Let's ignore the new `ResourceMapper` for now and let's check the legacy one, which intends to capture the current behavior in Solid.

> 2) Throw a 'File not found' error if createIfNotExists: false (which is always true if called via LegacyResourceMapper).

You can't run into such an error with `LegacyResourceMapper`. It is a simple functional mapping between a URL and a filename, which, unlike `ResourceMapper`, does not depend on the filesystem. The legacy mapper (aka current behavior) is independent of whether the file already exists, it's a simple 1-on-1 mapping, regardless of disk contents. This changes with the `ResourceMapper`, which follows https://www.w3.org/DesignIssues/HTTPFilenameMapping.html instead.

> I'd like a slightly different semantics - to not throw a 'file not found' error, and just return "if the file existed this is what its path would be" in legacy mode.

We should have that already. If not, you probably found a bug.

dmitrizagidulin: > You can't run into such an error with `LegacyResourceMapper`

Ohhh I see it now. I misread. n/m.

dmitrizagidulin:
Ok, so at this point, I should be able to hook up the new Delete handler to the existing ldp middleware, and it should (theoretically) pass the integration tests. (The delete operation is the easiest one, but still).

dmitrizagidulin:
Oh hey, check this out! Turns out Express.js has implemented [`res.sendFile`](https://expressjs.com/en/api.html#res.sendFile), with acceptRanges and everything. (Means we don't have to roll our own! At least for non-RDF resources).

lib/api/ldp/api.js
```diff
@@ -0,0 +1,137 @@
...
+    return Operation.from(options)
+  }
+
+  /**
+   * @param req {IncomingRequest}
+   *
+   * @returns {object} Operation constructor params object
+   */
+  parseOperation (req) {
+    const { method, headers, session: { authentication } } = req
```
RubenVerborgh: The knowledge about where and how the authentication is stored should probably live in another component.

lib/api/ldp/ldp-operation.js
```diff
@@ -0,0 +1,300 @@
...
+  writeHeaders ({res, result, permissions}) {
+    this.writeCommonHeaders({res, result})
+    // write HEAD-specific headers
+  }
+
+  /**
+   * @param res {ServerResponse}
+   */
+  writeResponse ({res}) {
+    res.sendStatus(200)
```
RubenVerborgh: Is that a useful default implementation? Empty body should be 201.
(We risk creating incorrect documents, for instance, `text/html` cannot be an empty string.)

lib/storage/ldp/ldp-file-store.js
```diff
@@ -0,0 +1,165 @@
...
+   * @param options.suppressDataBrowser {boolean}
+   */
+  constructor (options) {
+    this.fs = options.fs
+    this.mapper = options.mapper
+
+    this.suffixAcl = options.suffixAcl
+    this.suffixMeta = options.suffixMeta
+    this.dataBrowserPath = options.dataBrowserPath
+    this.suppressDataBrowser = options.suppressDataBrowser
```
RubenVerborgh: Maybe "databrowser" is too specific, rather something like a viewer?
(An if no viewer path is set, viewer is disabled.)

lib/storage/ldp/ldp-file-store.js
```diff
@@ -0,0 +1,165 @@
...
+   */
+  async resource (target) {
+    const { path, contentType: mediaType } = await this.mapper.mapUrlToFile(target)
+
+    let exists = true
+    let fsStats
+
+    // Try and load file metadata
+    try {
+      fsStats = await this.fs.stat(path)
```
RubenVerborgh: Is this a promise?

RubenVerborgh:
>> Re definition of operation - yeah, I think that's exactly it. Our definitions are differing. I think that even if we create a simple operation object (that's just a method and some flags),
> 
> I think we should rename in the architectural flow document.

In the architectural document, “operation” has now been split into “method” and “required permissions”: https://github.com/solid/solid-architecture/commit/38e85ba66d470142e6b89e9faedaa59624fb412c

lib/api/ldp/api.js
dmitrizagidulin: Heh, sure thing. I was only including it here cause it was mentioned in the architecture flow doc.

lib/api/ldp/ldp-operation.js
dmitrizagidulin: Ok, so this is something I struggled with for a bit. (Whether to return a 200 or a 204 for a HEAD response.) And the consensus out there seems to be a 200. (I'm not sure incorrect document is a threat here, as HEAD requests by definition don't include a response body.

lib/storage/ldp/ldp-file-store.js
dmitrizagidulin: Oh, agreed all the way. These are simply the existing flags - we'll need to refactor/rename them separately.

lib/storage/ldp/ldp-file-store.js
dmitrizagidulin: It is, yeah. See the Promise return type at https://nodejs.org/api/fs.html#fs_filehandle_stat_options

lib/api/ldp/ldp-operation.js
RubenVerborgh: No explicit mention of the status code in the RFC, but

> The HEAD method is identical to GET except that the server MUST NOT
   send a message body in the response (i.e., the response terminates at
   the end of the header section). 
— https://tools.ietf.org/html/rfc7231#section-4.3.2

So I tend to use the same status code as if it were a GET.

dmitrizagidulin: Heh, yeah, the lack of explicit mention of the status code in the RFC is part of the trouble here.

> So I tend to use the same status code as if it were a GET.

Meaning what? a 200?

RubenVerborgh: 200 / 204 / 404 / 410 / 500, whatever the GET would return :-)

dmitrizagidulin: Ok. So I think we're on the same page. So, going back to the code, it seems that a 200 code is reasonable as a default (and that errors like 404 / 500 etc will interrupt the flow and override).

dmitrizagidulin: As far as returning a 204 on a get, I agree with [this S.O. answer](https://stackoverflow.com/a/25510864) that while it's technically allowed, it's weird and somewhat nonsensical. And I'd rather avoid it. (Besides, I'm not sure how a 204 response on a GET is even possible on Solid server. So, non-issue.)

dmitrizagidulin:
(rebased on latest `develop`)

rimmartin:
Hi
Would Direct and Indirect container support be implemented better after this PR is merged or could these containers be implemented independent from this PR?

dmitrizagidulin:
@rimmartin I'd recommend that it's implemented after this PR.

That said, there is no plan (that I know of) to support either Direct or Indirect LDP container, in Solid. 

You might be best served by opening an issue on https://github.com/solid/solid and state your proposal (what would the semantics of Direct containers be, in the context of Solid? Solid is a web-ification of the file system, so what does it mean for a file to be stored in a non-Basic container? and so on.)

rimmartin:
Thank you @dmitrizagidulin .  I have a bit of a learning curve till this gets merged and LDP makes more sense to me

rimmartin:
Hi @dmitrizagidulin, I suppose v5.0.0 getting to release is priority over LDP API dev.

but I'm needing to simulate/store symlinks; at this point it is the main priority for my project https://github.com/DataDriven-CAM/solid-fs which is implementing node fs for isomorphic-git and any other javascript project needing it.

This PR is based on master? Do you think it is yet compatible with the 5.0.0 branch?

dmitrizagidulin:
@rimmartin It's not based on master currently, it was based on `develop` from several months ago (October). Let me see what it would take to base it on v5.

kjetilk:
So, I suppose we can close this now, and you'd rather have it as a module, @dmitrizagidulin ?

jaxoncreed:
@dmitrizagidulin Is this still being actively worked on?

dmitrizagidulin:
@jaxoncreed Yes, but based on our previous conversation with you and the team, I'm going to move this PR to a separate external repo. So, it's ok to close. Thanks!

dmitrizagidulin:
@kjetilk Ooof, my apologies, I somehow missed your previous comment! :)

jaxoncreed:
@dmitrizagidulin Sounds good

Issue #708, opened 2018-06-14
Support LetsEncrypt webroot plugin


kjetilk:
It seems like we can simplify the LetsEncrypt domain validation by accommodating for the webroot plugin. It [seems](https://certbot.eff.org/docs/using.html#webroot) just making sure files inside ```/.well-known/acme-challenge/``` is served will do this.

RubenVerborgh:
All that requires is a `/.well-known/` at the server root (or, in case of single user, in that user's data folder), and this folder should have a public ACL.

We can provide such a folder by default as a convenience to the user.

PR #715, opened 2018-06-15
Make per-origin .well-known folder

kjetilk:
This should add a .well-known folder with a corresponding .acl to allow the public to read. 

This should address Issue #708 . Was this what you had in mind  @RubenVerborgh .

Kjetil

### DIFF
--- a/default-templates/new-account/.well-known/.acl
+++ b/default-templates/new-account/.well-known/.acl
@@ -1,0 +1,19 @@
+# ACL resource for the well-known folder
+@prefix acl: <http://www.w3.org/ns/auth/acl#>.
+@prefix foaf: <http://xmlns.com/foaf/0.1/>.
+
+# The owner has all permissions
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
This set a `.well-known` folder for every user / subdomain (or main domain in the case of a single-user instance). I think this actually needs to be in the `server` folder, or we need to have one there too.

melvincarvalho:
Each origin would benefit from a well known folder, imho.  It allows bootstraps and nice P2P things like did: URIs and DHTs.

kjetilk:
Ah, I just looked in ```new-account```, but it makes sense it is for per-user stuff. 

I have no strong opinions on whether it makes sense to have a per-origin ```.well-known/```, but since I have done the work now, we might as well consider it. :-) 

I'll retitle this accordingly and make a new PR for one in ```server```.

PR #717, opened 2018-06-21
Set up SonarJS static analysis

RubenVerborgh:


RubenVerborgh:
Making it run is just the package and Travis changes. Stopping it from complaining is the other changes.

kjetilk:
OK, I would have preferred smaller, more atomic commits, but I can live with this. :-)

RubenVerborgh:
Yeah, my rule is that the build should work on each individual commit, so couldn't split up this one :-)

kjetilk:
Oh, OK! 

Actually, we should talk about that. My policy is just the build should succeed when I push, so I tend to have lots of commits between each push.

RubenVerborgh:
Indeed, we should have a document about policies and procedures.

kjetilk:
Yeah, lets discuss on gitter

RubenVerborgh:
I had to drop Sonar because it has issues with symlinks. We were using the wrong version apparently: https://github.com/SonarSource/SonarJS/issues/1057

Issue #720, opened 2018-06-30
My WebID isn't working on foreign site


kjetilk:
When trying to log into a remote site to a non-public document that my WebID https://solid.kjernsmo.net/profile/card#me has been given access to, I end up getting redirected to the remote site's login page. 

There has been some debug activity on gitter, but I figured I might as well make an issue on it, and attach a HAR file with a record of a session.

kjetilk:
Renamed the file to add a .log suffix. It is just JSON:
[Archive 18-07-01 00-01-56.har.log](https://github.com/solid/node-solid-server/files/2152242/Archive.18-07-01.00-01-56.har.log)

dmitrizagidulin:
What’s the error, though? Getting redirected to the remote site for login is the expected behavior.

kjetilk:
Huh? But surely, I shouldn't need an account on the remote site to access the documents that my own identity has been authorized to access?

dmitrizagidulin:
From conversation on Gitter, it sounds like the steps were:

1. Access a write-protected resource on `inrupt.net`, get a 401 Unauthorized response, get redirected to Select Provider screen.
2. Enter an external Web ID (hosted on `solid.kjernsmo.net`), get redirected to `kjernsmo.net/authorize` endpoint.
3. Log in at `solid.kjernsmo.net` via username & password.

**Expected:** to be redirected back to the `/shared/` protected resource on `inrupt.net`

**Actual:** got redirected to `inrupt.net/login` instead.

From the HAR log above, note that during the initial redirect to `kjernsmo.net/authorize`, the `redirect_uri=https://inrupt.net/api/oidc/rp/https%3A%2F%2Fsolid.kjernsmo.net` param is present (inside the `request` JWT), as expected.

And after login via password, the user is redirected to `https://inrupt.net/api/oidc/rp/https%3A%2F%2Fsolid.kjernsmo.net`, with the `code` and `state` params, as expected. So far so good.

So it sounds like the main issue is that the `redirectUrl` mechanism, after the callback request, is not working (and resorting to the default `/login` value). 

This was recently fixed in PR https://github.com/solid/node-solid-server/pull/648, so this is either a regression, or maybe `inrupt.net` running an older solid-server version?

dmitrizagidulin:
Ok, so it sounds like inrupt.net is running version `4.0.14`, which is after pr #648.

dmitrizagidulin:
[...]

PR #721, opened 2018-07-02
Update oidc-auth-manager to 0.16.1

dmitrizagidulin:
Closes issue #720

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,27 +5,12 @@
   "requires": true,
   "dependencies": {
     "@trust/jose": {
-      "version": "0.1.7",
-      "resolved": "https://registry.npmjs.org/@trust/jose/-/jose-0.1.7.tgz",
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
@RubenVerborgh I think there's been a change in how the lockfile gets generated, from (8 or 10?) to 10.5. Like, if I generate a lockfile on 8.9 (that's the one that's running on 8 & 10 but not 10.5), and compare it with a lockfile generated on 10.5, the version syntax that they're using changes. 

Sooo.. let's see if I generate a lockfile on 10.5 and commit it, if 8 and 10 can run the newer syntax. (If not.. we may need to maintain separate branches for node versions. Or git ignore the package lock)

RubenVerborgh:
10 should be fine, and we need to commit the lock indeed 🙂

dmitrizagidulin:
Ok whew, that seems to have worked :)

PR #723, opened 2018-07-10
DevOps changes to integrate snyk.io and email notifications

joshdcollins:
Based on previous conversations the below changes have been made to add security visibility into the build process:
- package.json updated to include snyk dependency.  This enables `snyk test` and `snyk wizard` commands; further the test hook runs `snyk test`
- .snyk added which contains the snyk policy.  All currently identified items are marked as snoozed for next 30 days so as to not immediately break builds
- travis.yml updated to add encrypted snyk.io API token (encrypted using travis cli) as well as email notifications

Note:  `snyk wizard` can be re-run to identify whether issues should be upgraded, ignored, or snoozed.


kjetilk:
I'm just wondering if it should run for every push...? Historically, I have sometimes pushed to see if the tests pass in a different environment, or if it is just my dev environment that is set up so that tests pass. In that case, this would be a distraction.

Given that Github also warns about possible security problems, perhaps it is possible to only run snyk for the master branch, or perhaps also develop?

joshdcollins:
@kjetilk I'd maybe challenge that a bit.  The best time to catch these items is at the time of injection (the feature branch) as opposed to waiting for a merge into develop/master.  My concern about doing it at that point is requiring someone from the core team to have to clean up an issue introduced by the PR (but not known until after the PR is merged).  To me, part of the goal IS for it to be a distraction so you're forced to resolve a potential issue as close to its point of entry as possible before it becomes baked in more difficult to regression test.

Ultimately, I'll go along with the configuration the two of you are comfortable with, but we generally try to shift as many of these checks to the feature branch level.

RubenVerborgh:
I follow @kjetilk's argument that build failures for a reason unrelated to the commit might be annoying.

How about having snyk.io as a cron job instead? https://docs.travis-ci.com/user/cron-jobs/

kjetilk:
I think you're right, @joshdcollins that it is important that the author has immediate feedback if they are introducing a *new* dependency. If they do, and there are many security issues with it, they should be told before they make a lot of code towards it, so that they can possibly opt to do it differently.

However, I assume that this is a relatively rare occasion, and I also suppose that checking snyk  to see if there are many security problems before introducing a new dep is something we could do as a matter of policy.

Wouldn't most red flags raised by snyk be unrelated to the commits you are doing, and therefore present an undesired distraction?

joshdcollins:
That's fair, ultimately there are two ways for a new issue from Snyk to break the build:

1. In the event of a dependency add/change that is made explicitly part of a PR
2. Being an unlucky contributor to be the first build after a new vulnerability is disclosed and identified

I don't see a means for Travis conditionally run build step based on something as granular as a specific file (package.json) changing.  

If we move it to only run on the develop/master branch that means that someone on the core team would be responsible for correcting any found issues -- is that a tenable option?  If so, I'll change the config to only run on develop/master.

To @RubenVerborgh 's point about a cron job, the actual snyk web console runs daily and I can add certain members as recipients of notifications on any new items found (wouldn't have to be executed out of Travis).

kjetilk:
OK, good point, @joshdcollins that we would have to deal with any breakage in the core team.... I don't know what the best approach is, but perhaps we should start with just master/develop, and then change it if it becomes a problem? What do you think, @justinwb ?

kjetilk:
Why the on-hold, label, @RubenVerborgh ? I'd like to get this in ASAP, but we just need to agree on the details. What do you make of the issue on whether to enable it on just master/develop or globally, @justinwb ?

RubenVerborgh:
Unsure, you take it over 🙂

justinwb:
@kjetilk i think a good middle group is to run on develop/master. Lets do that and see how it goes.

kjetilk:
OK, good! @joshdcollins can you please update the PR, to run on develop and master and fix the conflicts?

PR #725, opened 2018-07-14
Update README.md (corrected user folder location)

jacobpdq:
User folder locations are not under `./accounts` but `./data`


melvincarvalho:
Wonderful!  Thanks so much.  However, the data dir should possibly be consistent with the command line argument.  ie change that too?

What do you think?

jacobpdq:
I think that's what you mean @melvincarvalho:) thank you

Issue #685, opened 2018-05-28
Ensure solid-server works when installed globally with npm


RubenVerborgh:
There are reports of the solid-server executable not working when installed globally. It's rather important that it does, however, or we essentially have a broken npm package.

While this is hard to test on Travis, Solid developers could always run the global executable:
```
npm ls
solid-server start # rather than bin/solid-server
```

akuckartz:
Are there any tests which are run automatically ?

RubenVerborgh:
Yes, but only library tests, no tests that cover the executable.

justinwb:
Running into an issue when doing a clean global install on ubuntu.
npm = 5.6
node = 8.11.3

$ sudo npm install -g solid-server
npm WARN deprecated nodemailer@3.1.8: All versions below 4.0.1 of Nodemailer are deprecated. See https://nodemailer.com/status/
npm ERR! code 1
npm ERR! Command failed: /usr/bin/git clone -q git://github.com/solid/oidc-op.git /home/ubuntu/.npm/_cacache/tmp/git-clone-1fab4c56
npm ERR! /home/ubuntu/.npm/_cacache/tmp/git-clone-1fab4c56/.git: Permission denied
npm ERR! 

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/ubuntu/.npm/_logs/2018-07-09T17_16_24_769Z-debug.log

RubenVerborgh:
Seems like that specific issue would be solved in develop, as that version no longer depends on a package from GitHub (and it shouldn't).

@kjetilk Shall we release a new patch version?

kjetilk:
Yeah, lets do a patch version! Lets find a time for you to walk me through the release process.

RubenVerborgh:
Okay, ping me tomorrow afternoon!

justinwb:
let me know once it's pushed and i'll confirm the fix 👍

RubenVerborgh:
@justinwb Published, please test :-)

There might be other blockers ahead, but not the git ones.

justinwb:
This is still happening to me, but i believe that i tracked down the root cause, identified in this issue: https://github.com/zkat/pacote/issues/105

I can confirm that when i run the global install as sudo, i get the error error time. However, if I su to root rather than 'sudo -s' or running the command as sudo, the install works correctly.

RubenVerborgh:
> This is still happening to me

The exact same error? Because the dependency on a git repository should be gone.

[...]

PR #726, opened 2018-07-16
Update package.json

justinwb:
Updated solid-permissions to remove the git feature branch reference and point to the most recently published version (0.6.0) which corresponded with the group acl changes.


RubenVerborgh:
Regenerated package-lock.json, but I noticed there are still GitHub dependencies in there :-( Not good for #685.

PR #737, opened 2018-08-06
Fix logging in with client certificate

RubenVerborgh:
As described in https://github.com/solid/solid-auth-client/issues/50, there is a bug that results in an error “Cannot issue PoPToken - missing session key” for people logging in with a client certificate (in OIDC mode).

The root cause of this bug is a too tightly coupled design without clear interface boundaries, wherein several components assume knowledge about each other's internals. If those internals change, other things break—as has happened here.

The concrete problems in this case are:
- in a futile attempt to authenticate, node-solid-server was writing to the internal localStorage of solid-auth-client, the structure of which changed in solid-auth-client v2
- a variety of mechanisms are used to pass various parts of state from the login form to the OIDC libraries and back

This PR adds the following:
- let the server handle certificate authentication (32f9e7d3eb6eaaf853c65a08ac17c364ec29e268)
- create an additional mechanism to pass the return URL to the server (08aa6d3e4b34645776a0fce508d223fdb6a5bb5e and 20752d7f9409695b7919e945ca9893280a6949ea)

This is not a clean solution, but the fastest way to fix this issue.

### DIFF
--- a/default-views/auth/login-tls.hbs
+++ b/default-views/auth/login-tls.hbs
@@ -8,33 +8,3 @@
     {{> auth/auth-hidden-fields}}
   </div>
 </form>
-
-<script type="text/javascript">
-  const button = document.getElementById('login-tls')
-  button.addEventListener('click', function(event) {
-    event.preventDefault()
-    fetch('/login/tls', { method: 'POST', credentials: 'include' })
-      .then(function(response) {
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
When testing this, don't forget to update your views:

```bash
$ cp -r default-views/ config/views/
```

Issue #636, opened 2018-02-19
Incorrect Vary header


RubenVerborgh:
The server should at least have `Accept` and `Origin` as values for the `Vary` header in responses.

dmitrizagidulin:
@RubenVerborgh happy to take a crack at this, unless you're already fixing?

RubenVerborgh:
All yours 😉

dmitrizagidulin:
@RubenVerborgh Why `Origin` for the `Vary` header? (I agree with the `Accept` part)

RubenVerborgh:
We use the `cors` library, which sets the `Access-Control-Allow-Origin` header based on the `Origin` header of the request.

RubenVerborgh:
That said, `cors` likely sets its own `Vary` header, and I've seen that indeed. We should be careful not to overwrite it.

timbl:
If you don't have a Vary: Origin header then Chrome for example will cache a version from a different origin, and then reject it because the ACAO header doesn't match.

dmitrizagidulin:
Ok, so, just to clarify, we _do_ currently return a `Vary: Origin` header. 
So is the issue about also adding a `Vary: Accept`? Or can we close it?

RubenVerborgh:
Still need to fix Accept.

RubenVerborgh:
Ooh Authorization too

dmitrizagidulin:
@RubenVerborgh sure thing. added.

kjetilk:
Great stuff! Looks good to me, one quick review from one of you commenters on the bug, and I'll merge.

kjetilk:
OK, good, seems like the library prevents duplicate headers anyway, so we're good.

PR #740, opened 2018-08-07
Add Vary: Accept header

dmitrizagidulin:
Addresses #636

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -120,8 +120,8 @@ function initHeaders (app) {
   app.use(corsSettings)
 
   app.use((req, res, next) => {
-    // Set X-Powered-By
     res.set('X-Powered-By', 'solid-server')
+    res.set('Vary', 'Accept') // Cors lib adds Vary: Origin automatically
     // Set default Allow methods
     res.set('Allow', 'OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE')
     next()
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -122,8 +122,8 @@ function initHeaders (app) {
   app.use((req, res, next) => {
     res.set('X-Powered-By', 'solid-server')
 
-    // Cors lib adds Vary: Origin automatically
-    res.set('Vary', 'Accept, Authorization')
+    // Cors lib adds Vary: Origin automatically, but inreliably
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
I'll check in a bit.

RubenVerborgh:
Debug log for the above case is 

```
solid:server Error page because of: { Error: Could not verify Web ID from token claims
    at oidc.webIdFromClaims.then.catch.err (/Users/ruben/Documents/UGent/MIT/Solid/node-solid-server/lib/api/authn/webid-oidc.js:48:21)
    at process._tickCallback (internal/process/next_tick.js:178:7)
  statusCode: 401,
  cause: Error: Preferred provider for Web ID https://ruben.verborgh.org/profile/#me does not match token issuer https://localhost:8443
    at discoverProviderFor.then.preferredProvider (/Users/ruben/Documents/UGent/MIT/Solid/oidc-auth-manager/src/oidc-manager.js:388:15)
    at process._tickCallback (internal/process/next_tick.js:178:7) } +8s
```

dmitrizagidulin:
@kjetilk no worries there. the CORS lib automatically adds it for us.

RubenVerborgh:
> the CORS lib automatically adds it for us.

Except that it doesn't in at least one case, see above :wink:

dmitrizagidulin:
Oh I see. Hmmm. Lets try it?

dmitrizagidulin:
Er, wait a minute.

dmitrizagidulin:
No, I think that behavior is correct. The error log above has no Origin because it’s a server-side fetch, not from a browser

RubenVerborgh:
That doesn't matter; `Vary` still needs to be correct for proxies etc.

dmitrizagidulin:
Ah, ok. Well, lets set it ourselves then

kjetilk:
You'd think that a HTTP library somewhere would prevent duplicates from being sent to the client...? Should we just set it and see what happens?

RubenVerborgh:
Sure. A duplicate wouldn't be _that_ bad.

kjetilk:
I can't see more than one ```Vary``` header, and no duplicated ```Origin```. 

If there are any cases where you would expect a duplicate, could you please check, @RubenVerborgh ?

Issue #747, opened 2018-08-10
Provide one UI for login (instead of separate server/client ones)


RubenVerborgh:
Right now, there are two UIs for logging in:

1. one provided by node-solid-server
2. one provided by solid-auth-client

This is confusing, and it's hard to keep them in sync. (Another issue is that 1 only authenticates the user to the server; 2 also authenticates the user for AJAX requests: #736)

Since the solid-auth-client UI offers a better identity provider picker, I propose to migrate to that UI.

RubenVerborgh:
Fixed by #751.

Issue #736, opened 2018-08-06
Unified mechanism for maintaining login state across client and server


RubenVerborgh:
Right now, users need to log in once on the server (session cookie) and once on the client (so AJAX requests can be signed through OIDC).

We should evidently change this into one login. Maybe having establishing server-side authentication state (in the form of a [session] cookie) isn't necessary at all.

RubenVerborgh:
> but more importantly, to identify the user to the /authorize endpoint, as we discussed during the previous meeting.

Would it be an option to call the `/authorize` endpoint with an AJAX call that has the `Authorization` header set?

melvincarvalho:
@RubenVerborgh do you have an example of what would be set in the `Authorization` header?

RubenVerborgh:
The regular OIDC bearer token as for scripted OIDC requests.

```
Authorization: Bearer eyJhbGciOiJSUzI1NiJ9.ey…7iPAbNZlx-xccMCIfMu4oNJuLMWaw
```

melvincarvalho:
@RubenVerborgh thanks!

I agree with @dmitrizagidulin here.  The modular nature of solid requires, both scenarios.

In the longer term you could imagine multiple different user agents (browser, mobile, desktop, native app) with multiple login strategies to a server.  I dont think the server login can go away, in the short term (ie for 5.0.0).

I may have missed some points as I'm not as close to the identity / authn / authz side as you guys.  But I welcome proposals or extra detail.

RubenVerborgh:
> The modular nature of solid requires, both scenarios.

Hi @melvincarvalho, can you be very specific here? There is a difference between “we require server-side login state management” versus “we require browser-based access”.

> In the longer term you could imagine multiple different user agents (browser, mobile, desktop, native app) with multiple login strategies to a server.

The only user agent that uses “server login” (to be understood as “cookie-based session management”) are non-scripted browser interactions; none of the others require a server-based login whatsoever, nor will they ever.

melvincarvalho:
> There is a difference between “we require server-side login state management” versus “we require browser-based access”

Solid is a very modular system, with commodity servers, and a range of user agents.  These can be browsers, servers, robots, native apps and more.  Each user agent will have different capabilities and restrictions.  e.g. an electron app might be able to have better key management than a mainstream browser (for example safe browser, which has [implemented](https://youtu.be/eO6pYYWZBs8?t=03h55m18s) webid switching, including per tab webids)

[...]

PR #751, opened 2018-08-12
Always log in through the client

RubenVerborgh:
Currently, when accessing a protected resource in the browser, node-solid-server would provide you with its own UI for selecting an IDP and logging in. This would only create a server session, but not a client login: regular browser requests would work, but AJAX requests would fail. You would need to log in a second time with solid-auth-client to make Databrowser and others work.

To address that, this pull request:
- removes the node-solid-server IDP selection UI
- replaces it with a solid-auth-client login (fixing #747)

Not only does this provide an identical user experience for server- and client-initiated logins, it also synchronizes client-side and server-side login (to the extent possible; more work to be done in #736).


justinwb:
This seems to be working well. The only problem that I ran into was the symlinks not getting created correctly when I did the install, which I did in the inrupt dev environment by running 'npm install -g git://github.com/solid/node-solid-server.git#feature/client-login'. Once I created them by hand, everything worked well.

RubenVerborgh:
Thanks for testing. Is the location of the module different in a global install? It might be at ../solid-auth-client instead of ./node_modules/solid-auth-client .

If so, I need to account for this.

kjetilk:
Sounds good! Seems like we can get this into 4.1.0 indeed, if we can get that final issue settled.

justinwb:
It's still at ./node_modules/solid-auth-client even in the global install (absolute path of /usr/lib/node_modules/solid-server/node_modules/solid-auth-client on this ubuntu 16.04 install). I did some very quick digging and i think it may actually be a problem with npm's support of symlinks. Check out https://github.com/npm/npm/issues/13050

PR #769, opened 2018-08-25
Use latest solid-auth-client

RubenVerborgh:
In addition to the latest fixes, this also allows us to considerably simplify our client-side scripts. (Note that the existing scripts continue working though.)

There's currently one drawback, and that is that `localStorage` needs to be cleared because the popup location changed. Will try to address that in solid-auth-client.

### DIFF
--- a/default-templates/new-account/index.html
+++ b/default-templates/new-account/index.html
@@ -65,65 +65,18 @@
     </div>
   </section>
 </div>
-</body>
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
The current test setup can unfortunately not really cover this; they are client-side scripts, and we do not have any such client-level testing.

PR #770, opened 2018-08-30
Communicate an originsAllowed from config

kjetilk:
This should communicate an `originsAllowed` option all the way from the default config and up to the ACLchecker, which again passes it to the `PermissionSet`. Also adds some debugging.

Obviously, the actual trusted origin has to change once we have registered and agreed on domain for it.

We should perhaps have an integration test, but I figured I'd let you poke at it first. :-)

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -12,6 +12,7 @@ module.exports = {
   'serverUri': 'https://localhost:8443',
   'webid': true,
   'strictOrigin': true,
+  'originsAllowed': ['https://apps.solid.invalid'],
   'dataBrowserPath': 'default'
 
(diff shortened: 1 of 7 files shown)


kjetilk:
Bug description is in https://github.com/solid/solid-permissions/issues/26

kjetilk:
I'm thinking perhaps that some of the strange behaviour that I've seen is actually caused by wrong configs, i.e. #776 .

RubenVerborgh:
Can you look at merging #776 then? I'll rebase.

RubenVerborgh:
#776 has been merged, you might want/need to rebase.

PR #774, opened 2018-09-04
Merge master into develop

kjetilk:
Develop is now rather far behind master, we should merge it into develop, I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@
 
 ### Install
 
-You can install and run the server either using Node.js directly or using 
-[Docker](https://www.docker.com/).  This and the following sections describe the 
+You can install and run the server either using Node.js directly or using
+[Docker](https://www.docker.com/).  This and the following sections describe the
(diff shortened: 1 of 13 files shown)


kjetilk:
Arrrghh! @justinwb could you please have a look at this too...?

justinwb:
This looks good to me @kjetilk. The CI build should pass once the merge is done and package/package-lock.json files are updated

RubenVerborgh:
Oh right, these should have been in develop in the first place. Let's perhaps lock down master to prevent this from happening.

RubenVerborgh:
> The CI build should pass once the merge is done

Hmm, there isn't really a difference between the commit in the PR and the commit after merge. If the PR fails, the merged result should also be expected to fail.

I'll set up status checks before merging to develop is allowed (admins can bypass).

RubenVerborgh:
Build expected fixed in 241c15f66e4fc290d085af7c3c24fef7abb0f282.

PR #776, opened 2018-09-05
Ensure that tests use test config

RubenVerborgh:
Tests would fail if contents in `./config` differed.

To test, delete ` ./config/views/*`; should fail without this commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ install:
 
 script:
   - npm run standard
-  - npm run sonar
   - npm run nyc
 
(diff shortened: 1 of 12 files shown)


RubenVerborgh:
This removes Sonar, but we were using the wrong thing anyway (https://github.com/SonarSource/SonarJS/issues/1057).

RubenVerborgh:
@kjetilk Rebased.

kjetilk:
I noticed when trying to resolve conflicts :-)

PR #781, opened 2018-09-11
Merge develop into v5.0.0

kjetilk:
We should merge develop into v5.0.0 so that they don't diverge too much. OK?


kjetilk:
We can, but I don't quite see why, we'd like to have the rest in 5.0.0 too, and branches diverging are more likely to cause conflicts that result in extra work. 

Also, it would require another PR, wouldn't it?

RubenVerborgh:
It doesn't matter _that_ much I suppose; I'd just like to have a clear merging policy, as to avoid too many random incoming merges from develop.

I'd say that, every time we see a need to merge develop into v5, we probably also have a need to release a new patch version—and vice-versa.

So maybe we release a patch version from develop, and merge that in?

A new PR would probably not be needed.

PR #783, opened 2018-09-11
Fix code for Login redirect

cblakeley:
A response_type of 'code' needs to be recognized for relying parties using authorization flow.
OSDB uses the solid/oidc-rp library in this way.

### DIFF
--- a/lib/requests/login-request.js
+++ b/lib/requests/login-request.js
@@ -158,7 +158,8 @@ class LoginRequest extends AuthRequest {
    */
   postLoginUrl (validUser) {
     // Login request is part of an app's auth flow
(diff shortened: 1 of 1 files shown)


kjetilk:
Thanks for the PR! 

Could you please make sure tests pass, and also please PR against the `develop` branch when making PRs on NSS. Right now, they are the same, but they tend to diverge between releases, so it is better to be consistent about it.

kjetilk:
Sounds good! @dmitrizagidulin , care to have a look?

smalinin:
Fixed issue with pass tests

dmitrizagidulin:
@kjetilk sure. Looks good to me. :+1:

PR #792, opened 2018-09-26
Use local Mashlib

### LINKED ISSUE #773: Use local mashlib version
The embedded Databrowser currently points to a public mashlib build. We probably want to point to a local copy, such that we can control the version number (as we do with solid-auth-client).

RubenVerborgh:
Closes #773.


kjetilk:
I don't know too much about this code, so I'm inclined to trust your judgment and Travis on this. :-)

PR #795, opened 2018-09-29
Upgrade to oidc-op-0.4.0

RubenVerborgh:
Should fix https://github.com/solid/oidc-op/issues/10


RubenVerborgh:
@kidehen Deployed!

Issue #796, opened 2018-09-29
Git dependency on global install causes install failures with sudo


justinwb:
In the most recent path release 4.1.4, a user reported being unable to do a global install of solid-server when sudoed as root. Looks like a git dependency has creeped back in. Also we should ensure travis is catching cases like this as part of the resolution.

npm ERR! Command failed: /usr/bin/git clone --depth=1 -q -b gh-pages git://github.com/linkeddata/better-simple-slideshow.git /root/.npm/_cacache/tmp/git-clone-3bcf071d
npm ERR! fatal: could not create leading directories of '/root/.npm/_cacache/tmp/git-clone-3bcf071d': Permission denied

timbl:
Original gitter char here: https://gitter.im/solid/node-solid-server?at=5baf952c271506518d904424   .. @RubenVerborgh poinst out probably mashlib.
We should then have a test at many levels or at least mashlib that git dependencies havent crept in n

timbl:
```
$ cd mashlib/
$ grep git package-lock.json 
      "version": "github:linkeddata/better-simple-slideshow#gh-pages",
      "from": "better-simple-slideshow@github:linkeddata/better-simple-slideshow#gh-pages"
        "better-simple-slideshow": "github:linkeddata/better-simple-slideshow#gh-pages",
$ 

```

RubenVerborgh:
Working on this in https://github.com/solid/node-solid-server/pull/797.

In the meantime there is https://github.com/linkeddata/mashlib/issues/18 on Mashlib. Can we use an npm version of better-simple-slideshow, @timbl?

RubenVerborgh:
Travis now complains if there are GitHub dependencies: https://travis-ci.org/solid/node-solid-server/jobs/435068777

Let's:
- [ ] get Mashlib fixed in linkeddata/mashlib#18
- [ ] update the Mashlib version in #797
- [ ] merge #797

PR #797, opened 2018-09-29
Test and fix global package installation

RubenVerborgh:
Ensure there are no GitHub dependencies (#796) or other things that might break a global installation,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ addons:
     - nicola.localhost
 
 before_install:
+  - echo "No GitHub dependencies allowed" &&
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Since Travis uses nvm, global installation doesn't require sudo and just works (CC @justinwb).

However, trying global installation might catch other problems, and I'm adding an explicit check for GitHub packages in package-lock.json.

RubenVerborgh:
Awaiting Mashlib fix in https://github.com/linkeddata/mashlib/issues/18

RubenVerborgh:
(Accidentally merged into master, but fixed.)

PR #802, opened 2018-09-30
Fix Local Windows Development

mohamedmansour:
When it was reading from disk index.html, it added the wrong
slash, use url.resolve to fix this.

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -1,5 +1,5 @@
 module.exports = allow
-const url  = require('url')
+const url = require('url')
 const ACL = require('../acl-checker')
 const $rdf = require('rdflib')
 const utils = require('../utils')
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -1,5 +1,5 @@
 module.exports = allow
-
+const url  = require('url')
 const ACL = require('../acl-checker')
 const $rdf = require('rdflib')
 const utils = require('../utils')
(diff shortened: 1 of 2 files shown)


mohamedmansour:
@dmitrizagidulin this fixes the ACL to match, tested on windows 10

dmitrizagidulin:
:+1: looks reasonable.

mohamedmansour:
@RubenVerborgh  how would you otherwise do this when `reqPath` is `\index.html` Manual string manipulation.  This is how express handles it since it is coming from the windows filepath. Are you certain URL.resolve is bad for security? We always use it to make sure URLs are joined correctly is cross plat.

mohamedmansour:
@RubenVerborgh  maybe the way we are reading the ACL? `<./index.html>` Node resolves it as `\index.html` in Windows? Let me understand how ACLs are being read now.

RubenVerborgh:
> maybe the way we are reading the ACL? `<./index.html>` Node resolves it as `\index.html` in Windows?

It's not Node resolving this, it must be our code instruction Node to do so. But ACLs should be resolved as URLs, not as files. So there is something wrong with our ACL resolver code if filename conventions (like backslashes) are creeping in.

kjetilk:
@RubenVerborgh Perhaps raise a separate issue about ACL resolving?

RubenVerborgh:
Yes. @mohamedmansour Could you come up with an exact description of what you encounter under what circumstances? I.e., a bug report? Then we can investigate if this PR is the correct fix for that bug.

PR #804, opened 2018-09-30
Send errors as plain text

RubenVerborgh:


RubenVerborgh:
@kjetilk Nothing in the stack should. If they do, we want to double check that anyway.

RubenVerborgh:
That said, I haven't seen any code throw HTML errors (thankfully).

PR #806, opened 2018-09-30
Disable CORS proxy by default

RubenVerborgh:


RubenVerborgh:
Thanks, only new installs indeed.

PR #809, opened 2018-09-30
Use default popup URI

RubenVerborgh:
solid-auth-client supports a .well-known popup URI as a default, so it is not necessary to specify this popup.

Moreover, if we do specify this popup URI, we risk running into https://github.com/solid/solid-auth-client/issues/54

Also see https://github.com/solid/solid-ui/pull/43


RubenVerborgh:
develop is fine, I want to release a bunch of such simple fixes with a patch tomorrow.

PR #817, opened 2018-10-01
Feature/eslint

megoth:
An initial config for ESLint with /* eslint disable */ all files in lib/.

The idea is to remove /* eslint disable */ when you touch a file and you want to tidy it up. If we never do that, fine, but now we have the possibility.

I've also added husky to help with pre-commit; basically, if you try to commit a file that isn't linted correctly you'll get an error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "dirty-chai": "^1.2.2",
     "eslint": "^5.6.1",
     "eslint-config-node": "^4.0.0",
+    "husky": "^1.0.1",
     "localstorage-memory": "^1.0.2",
     "mocha": "^5.1.1",
     "nock": "^9.0.14",
@@ -102,8 +103,7 @@
     "nyc": "NODE_TLS_REJECT_UNAUTHORIZED=0 nyc --reporter=text-summary mocha",
     "mocha": "NODE_TLS_REJECT_UNAUTHORIZED=0 mocha",
     "test": "npm run standard && npm run nyc",
-    "lint": "eslint lib/**",
-    "precommit": "npm run lint"
+    "lint": "eslint lib/**"
   },
   "nyc": {
     "reporter": [
@@ -111,6 +111,11 @@
       "text-summary"
     ],
     "cache": true
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint"
+    }
   },
   "standard": {
     "globals": [
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "chai-as-promised": "^6.0.0",
     "dirty-chai": "^1.2.2",
     "eslint": "^5.6.1",
+    "eslint-config-node": "^4.0.0",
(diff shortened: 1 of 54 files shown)


RubenVerborgh:
@megoth Just curious to hear about what linting rules we will adopt (if not standard).
Not that I'm a fan of standard, but there must have been some reason for adopting it.

megoth:
Ah, yes, I simply took the config rules for https://www.npmjs.com/package/eslint-config-node; it seemed nice and somewhat updated, so I was thinking it suited our need. But I should've talked with the rest of the team first, in case others had suggestions on which config to start with.

Apart from that @kjetilk and I set up the initial config together, so should reflect the basic code style there (spaces, no enforcements of semi-colons).

This can also be changed as we make use of it, in case we discover it is a mismatch with our needs.

RubenVerborgh:
But `standard` uses eslint internally, so simply setting up eslint with `standard` as base rules should do it?

dmitrizagidulin:
Wait, this repo (as well as most of the other repos in the Solid ecosystem) already use Standard.js for linting. (And it's enforced by a pre-commit hook.)

If the team has decided that Standard no longer suffices, I'd love to at least see a discussion / see some reasons given? :)

megoth:
We could revert to standard and set ecmaVersion to 2015, and I think that should cover the basics for us.

package.json
```diff
@@ -99,7 +102,8 @@
     "standard": "standard '{bin,examples,lib,test}/**/*.js'",
```
RubenVerborgh: This should go.

package.json
```diff
@@ -99,7 +102,8 @@
...
     "nyc": "NODE_TLS_REJECT_UNAUTHORIZED=0 nyc --reporter=text-summary mocha",
     "mocha": "NODE_TLS_REJECT_UNAUTHORIZED=0 mocha",
-    "test": "npm run standard && npm run nyc"
+    "test": "npm run standard && npm run nyc",
```
RubenVerborgh: standard should go or be replaced with lint

dmitrizagidulin: Why, though?

megoth:
I'm sorry, I didn't know we used standard (I haven't used it, so didn't reckognize it as a linter) =/ 

I jumped the gun on this one, my bad.

I think we could just drop this PR, as there is no need for an additional linter. I'll also make sure to check in with rest of the team before I suggest something beside the issues we're working on ^_^ (I tried to do this as a learning case to get to know the code base better)

melvincarvalho:
We're trying to use standard in all our repos, not just NSS.

It was @nicola that suggested it, and I've come to like it alot.

megoth:
I've closed this, and we'll continue using standard ^_^

package.json
RubenVerborgh: _If_ this PR replaces standard by lint, then this should happen. (It would crash now because the dependency was removed.)

melvincarvalho: Which dependency was removed in which branch?

PR #818, opened 2018-10-01
Validate usernames before creating accounts.

### LINKED ISSUE #810: Check usernames for invalid characters at character creation
Currently on https://inrupt.net/ you can register with a username of  (for example) "stuart yeates" which then generates a "Welcome to Solid" email with a bad link of https://stuart/profile/card#me

Either the entire system needs to properly handle and encode non-alphabetic characters  (probably fairly hard) or the input form needs to check and warn when non-alphabetic characters are present.

fourcube:
Fixes #810.

### DIFF
--- a/lib/requests/create-account-request.js
+++ b/lib/requests/create-account-request.js
@@ -124,6 +124,7 @@ class CreateAccountRequest extends AuthRequest {
     let accountManager = this.accountManager
 
     return Promise.resolve(userAccount)
+      .then(this.cancelIfUsernameInvalid.bind(this))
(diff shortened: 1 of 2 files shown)


kjetilk:
Great stuff, thanks a lot for the PR! :-)

a small nitpick, since you reported the problem as one where a period caused the problem, perhaps it would be a good idea to add e.g. `foo.bar` as an entry to the list of `invalidUsernames`.

kjetilk:
That was fast! Thanks a lot, @fourcube !

fourcube:
My pleasure, thanks for the fast review!

Issue #662, opened 2018-05-03
Wire up ResourceMapper


RubenVerborgh:
The `ResourceMapper` introduced in #643 should be wired up, such that the new URL/filename correspondence is used.

The steps are:
1. [x] wire up the `LegacyResourceMapper` first (#661)
2. [x] replace `LegacyResourceMapper` by `ResourceMapper` and adjust tests (#662)

rubensworks:
Done in #959

Issue #661, opened 2018-05-03
Wire up LegacyResourceMapper


RubenVerborgh:
The `LegacyResourceMapper` introduced in #660 should replace all inline URL to filename (and vice-versa) mapping, so we:
- provide consistent mapping behavior all over, and eliminate bugs in existing mapping code (or at least confine those bugs to one place)
- simplify the transition to the new `ResourceMapper` (#662)

RubenVerborgh:
We possibly might want to do this by changing the interface of `ldp`, such that it accepts URLs rather than filenames. Then, `ldp` itself calls `LegacyResourceMapper` instead of requiring all of its callers to do that.

rubensworks:
Ok, I'm working on this issue now.

I'm still relatively new to the codebase, so I could use some input on where I could find calls that should be changed.
So far, I see the following:
* [`util.js/uriToFileName()`](https://github.com/rubensworks/node-solid-server/blob/feature/wire-legacy-resource-mapper-2/lib/utils.js#L70)
* [`util.js/uriToRelativeFilename() `](https://github.com/rubensworks/node-solid-server/blob/feature/wire-legacy-resource-mapper-2/lib/utils.js#L87)
* [`util.js/getBaseUri() `](https://github.com/rubensworks/node-solid-server/blob/feature/wire-legacy-resource-mapper-2/lib/utils.js#L93)
* [`util.js/getFullUri() `](https://github.com/rubensworks/node-solid-server/blob/feature/wire-legacy-resource-mapper-2/lib/utils.js#L117)
* [`util.js/pathBasename()`](https://github.com/rubensworks/node-solid-server/blob/feature/wire-legacy-resource-mapper-2/lib/utils.js#L121)

I assume there may be others as well?

RubenVerborgh:
My heuristic is: everywhere where multiuser is checked.

megoth:
@rubensworks if you're going to edit `lib/utils.js`, maybe it's a good time those functions to a utility-file in `lib/common`? Trying to make methods a bit more reusable by putting them in more explicitly named utility-files.

rubensworks:
@megoth The plan is actually to remove as much as possible from `lib/utils.js` :-) In favor of the new methods of `ResourceMapper`.

But the remaining functions could indeed be moved to `lib/common` afterwards.

rubensworks:
I assume #911 will be merged before this refactor is done, so I'll have to rebase in any case.

rubensworks:
[...]

Issue #454, opened 2016-10-26
HEAD doesn't set right Content-Type  or right Content-Length headers


elf-pavlik:
for example GET request to foo.png includes correct header
`Content-Type: image/png`
but HEAD request to foo.png responds with **incorrect** header
`Content-Type: text/plain; charset=utf-8`

dmitrizagidulin:
Interesting.. thanks for reporting this!

dmitrizagidulin:
Also address setting the correct `Content-Length` header on HEAD requests.

timbl:
About time this serious breakage of the HTTP protocol was fixed

timbl:
Grumble.  The bad content-length makes checking that JS library files are the right ones  more difficult.   It is basic HTTP that headers on HEAD *MUST* be exactly the ones you would get with GET 
timbl added the priority  label on Feb 21, 2017

RubenVerborgh:
Might be fixed with #662.

megoth:
@elf-pavlik hey, I've been having trouble reproducing the bug with Content-Type; could you maybe send me the PNG file you had trouble with?

The bug with Content-Length I've been able to reproduce though (as there is no property with that), so I'll continue working on that for now ^_^

megoth:
@kjetilk and I are realizing that this might not be best handled in the current code; Content-Length should be handled automatically as part of http-handling in general, and there is probably some middleware out there that we could use.

We're suggesting that this get labelled with “test suite” (or something like it), indicating that we will write tests that handles the various use cases where Content-Length should be set, and that this test-suite will be used by us and others wanting to implement a Solid-compliant server.

We're also suggesting that this issue be labelled with revisit, as this issue is not critical.

kjetilk:
Hmmm, or it should be a bug on that `HEAD` isn't identical to `GET` headers, which is also just really weird that the framework doesn't do right out of the box.

megoth:
Yes, once again I didn't quite get the task =P I've been able to reproduce the bug now, and will continue on it.

RubenVerborgh:
If you ask me, we can just have the exact same logic, except we don't write a body (so catch all 'res.end/write` calls or `pipe`s to it). That should do for now; optimizations always possible later.

kjetilk:
[...]

PR #831, opened 2018-10-02
Setting contentType for HEAD requests

megoth:
res.sendStatus(200) didn't pass the correct Content-Type, so this fix should handle that.

I think this could be handled better: we should not need to set Content-Type manually like this, that should be handled by the external http-handlers we use. But since get.js is kinda invasive in order to handle RDF-files, I think this is the "best" solution (for now).

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -17,6 +17,7 @@ const translate = require('../utils.js').translate
 const error = require('../http-error')
 
 const RDFs = require('../ldp').RDF_MIME_TYPES
+const LegacyResourceMapper = require('../legacy-resource-mapper')
 
 function handler (req, res, next) {
   const ldp = req.app.locals.ldp
@@ -76,7 +77,13 @@ function handler (req, res, next) {
     // Till here it must exist
     if (!includeBody) {
       debug('HEAD only')
-      contentType = mime.contentType(_path.extname(path))
+      const rootUrl = utils.getBaseUri(req)
+      const mapper = new LegacyResourceMapper({
+        rootUrl,
+        rootPath: ldp.root,
+        includeHost: ldp.multiuser
+      })
+      contentType = mapper._getContentTypeByExtension(path)
       res.setHeader('Content-Type', contentType)
       res.status(200).send('OK')
       return next()
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -76,7 +76,9 @@ function handler (req, res, next) {
     // Till here it must exist
     if (!includeBody) {
       debug('HEAD only')
-      res.sendStatus(200)
+      contentType = mime.contentType(_path.extname(path))
+      res.setHeader('Content-Type', contentType)
+      res.status(200).send('OK')
       return next()
     }
 
@@ -85,8 +87,8 @@ function handler (req, res, next) {
       let mimeTypeByExt = mime.lookup(_path.basename(path))
       let isHtmlResource = mimeTypeByExt && mimeTypeByExt.includes('html')
       let useDataBrowser = ldp.dataBrowserPath && (
-          container ||
-          RDFs.includes(contentType) && !isHtmlResource && !ldp.suppressDataBrowser)
+        container ||
+        RDFs.includes(contentType) && !isHtmlResource && !ldp.suppressDataBrowser)
 
       if (useDataBrowser) {
         res.set('Content-Type', 'text/html')
@@ -160,7 +162,7 @@ function globHandler (req, res, next) {
     debugGlob('found matches ' + matches)
     Promise.all(matches.map(match => new Promise((resolve, reject) => {
       const baseUri = reqOrigin + '/' + match.substr(root.length)
-      fs.readFile(match, {encoding: 'utf8'}, function (err, fileData) {
+      fs.readFile(match, { encoding: 'utf8' }, function (err, fileData) {
         if (err) {
           debugGlob('error ' + err)
           return resolve()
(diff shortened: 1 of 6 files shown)


RubenVerborgh:
> we should not need to set Content-Type manually like this

Note in this context #643 and related ones.

megoth:
This is a fix for #454 btw (forgot to link it earlier)

lib/handlers/get.js
```diff
@@ -76,7 +76,9 @@ function handler (req, res, next) {
     // Till here it must exist
     if (!includeBody) {
       debug('HEAD only')
-      res.sendStatus(200)
+      contentType = mime.contentType(_path.extname(path))
```
RubenVerborgh: Not complete, since the current server assumes that extensionless paths are Turtle.

I would suggest to use an instance of LegacyResourceMapper here. It's overkill for now, but will give you the correct answer, and can in the future easily be replaced by ResourceMapper (which has a different MIME type detection system). See #661 and #662.

You can see an example instantiation in `lib/handlers/allow.js`.

megoth:
I've added a new test to handle extensionLess URLs, and made use of LegacyResourceMapper.

megoth:
I've stumbled upon a problem with extensionless files and NamedNode in rdflib. I've spoken with @kjetilk and we've decided that I'll put this a bit on hold (and do more research later on, hopefully with the help of @RubenVerborgh or @timbl ) so that we can focus on #177.

lib/handlers/get.js
```diff
@@ -76,7 +77,15 @@ function handler (req, res, next) {
     // Till here it must exist
     if (!includeBody) {
       debug('HEAD only')
-      res.sendStatus(200)
+      const rootUrl = utils.getBaseUri(req)
+      const mapper = new LegacyResourceMapper({
```
RubenVerborgh: Just wondering if you can initialize this earlier and reuse; fine otherwise.

lib/handlers/get.js
```diff
@@ -76,7 +77,13 @@ function handler (req, res, next) {
...
     if (!includeBody) {
       debug('HEAD only')
-      contentType = mime.contentType(_path.extname(path))
+      const rootUrl = utils.getBaseUri(req)
+      const mapper = new LegacyResourceMapper({
+        rootUrl,
+        rootPath: ldp.root,
+        includeHost: ldp.multiuser
+      })
+      contentType = mapper._getContentTypeByExtension(path)
```
RubenVerborgh: This is a private method though (underscore); you probably want `mapper.mapFileToUrl({ path }).contentType`.

lib/handlers/get.js
megoth: I could probably do that, this is more or less a hack to get it handled properly

lib/handlers/get.js
megoth: Ok, I'll do that ^_^

megoth:
@timbl @RubenVerborgh it's a little difficult for me to explain the problem, but I'll try.

The reason that the test breaks now is that `test/integration/ldp.js` -> `listContainer` -> `should ldp:contains the same files in dir` lists `sampleContainerblank` for the file `blank` (no extensions) instead of `blank`. 

I might be misunderstanding how it's used, but I think I can track the problem to an object (in a triple) that is set as NamedNode, and that doesn't parse the `memberUri` correctly. This is code in question (line ~60 or so in `lib/ldp-container.js`):

```
resourceGraph.add(
      resourceGraph.sym(containerUri),
      ns.ldp('contains'),
      resourceGraph.sym(memberUri))
```

`memberUri` in the case of the blank file is `https://server.tld/resources/sampleContainer/blank`, and that is not parsed correctly. The statement above becomes (I think):

```
Statement {
  subject:
   NamedNode {
     termType: 'NamedNode',
     value: 'https://server.tld/sampleContainer' },
  predicate:
   NamedNode {
     termType: 'NamedNode',
     value: 'http://www.w3.org/ns/ldp#contains' },
  object:
   NamedNode {
     termType: 'NamedNode',
     value: 'https://server.tld/sampleContainerblank' },
  why:
   NamedNode {
     termType: 'NamedNode',
     value: 'https://server.tld/sampleContainer' } }
```

So we see that the value of object is not what we expect.

Do I maybe need to call another method resourceGraph.sym in the case of an extensionless file? (I'll try this route next.)

timbl:
It looks to me as though memberUri is wrong

timbl:
Look at the code which generates the listing of the directory?  "https://server.tld/resources/sampleContainer/blank"  is at some point constructed from the URL the folder "https://server.tld/resources/sampleContainer/" and the filename "blank" -- or it is constructed from the URI of the foldder less the slash "https://server.tld/resources/sampleContainer" and "/" and "blank".   
Sounds like confusion as to wether the URI of the folder includes the slash
... in general it should so the code which generated the URI of the folder is wrong, or may
be this is a complex case where the client has been redirected from folder without slash to folder with slash

timbl:
Yes containerUri is wrong it should have a slash

megoth:
Finally got my head around back to this, and thanks to @timbl I fixed it ^_^ (so annoying when I get so focused on the wrong parts that I don't see the actual problem =P)

megoth:
Just to check; any reason to why `ResourceMapper.mapFileToUrl` was asynchronous? None of the functions it called was async, so I removed it and modified the places that used it accordingly.

Apart from that everything should be good for mergin now, @kjetilk ^_^

RubenVerborgh:
> Just to check; any reason to why `ResourceMapper.mapFileToUrl` was asynchronous?

Forward compatibility.

megoth:
Ok, then I'll make it asynchronous again ^_^

megoth:
New commits in place, that doesn't remove the asynchronous part of `mapFileToUrl`.

Looks good @RubenVerborgh ?

RubenVerborgh:
Fine with me!

PR #834, opened 2018-10-02
Allow cookies from all subdomains

RubenVerborgh:
It seems that applying #793 has enabled a security mechanism (#526) that is too strong. (It is unclear why this security mechanism didn't kick in earlier.)

#526 rejects cookies as authentication mechanism from other hosts, requiring other hosts to always send a Bearer header.

However, its definition of "other hosts" did not include subdomains. For example:
- https://alice.solid.community would not be trusted to send a login cookie for https://my.self/#me by https://solid.community

This pull request changes things such that
- https://*.solid.community is trusted to send a login cookie for https://my.self/#me by https://solid.community
- https://my.self is trusted to send a login cookie for https://my.self/#me by https://solid.community

This addresses breakage in 4.1.4 and 4.1.5 caused by #793 as noted by @kidehen in https://github.com/solid/node-solid-server/pull/793#issuecomment-426341408

### DIFF
--- a/lib/models/solid-host.js
+++ b/lib/models/solid-host.js
@@ -72,12 +72,14 @@ class SolidHost {
   allowsSessionFor (userId, origin) {
     // Allow no user or an empty origin
     if (!userId || !origin) return true
-    // Allow the server's main domain
-    if (origin === this.serverUri) return true
-    // Allow the user's subdomain
-    const userIdHost = userId.replace(/([^:/])\/.*/, '$1')
-    if (origin === userIdHost) return true
-    // Disallow everything else
+    // Allow the server and subdomains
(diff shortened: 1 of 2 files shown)


lib/models/solid-host.js
```diff
@@ -109,4 +111,9 @@ class SolidHost {
   }
 }
 
+function getHostName (origin) {
```
kjetilk: Doesn't using `origin` confuse that there may be other URIs that the hostname is extracted from than the `Origin`? Or am I confused for other reasons?

RubenVerborgh: You are right, I will change the parameter name.

PR #835, opened 2018-10-02
Block cookie auth from non-subdomains, but allow all other auth from everywhere

RubenVerborgh:
#834 overreached, since it applied the cross-domain cookie ban from #793 to all auth methods, not just the cookie-base auth method.

The implementation currently contains a level-breaking hack, where the general module needs to know about the OIDC logout paths. A proper fix would require a more intense refactoring, where cookie-based auth is seen as a third method in addition to WebID-TLS and WebID-OIDC, not as part of either.

Also, we should note that this fix (as did #834) allows _any_ third party app to log the user out without their explicit permission if they happen to know the user's pod.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -168,23 +168,27 @@ function initWebId (argv, app, ldp) {
   // (for same-domain browsing by people only)
   const useSecureCookies = !!argv.sslKey // use secure cookies when over HTTPS
   const sessionHandler = session(sessionSettings(useSecureCookies, argv.host))
+  app.use(sessionHandler)
+  // Reject cookies from third-party applications.
+  // Otherwise, when a user is logged in to their Solid server,
+  // any third-party application could perform authenticated requests
+  // without permission by including the credentials set by the Solid server.
   app.use((req, res, next) => {
-    sessionHandler(req, res, () => {
-      // Reject cookies from third-party applications.
-      // Otherwise, when a user is logged in to their Solid server,
-      // any third-party application could perform authenticated requests
-      // without permission by including the credentials set by the Solid server.
-      const origin = req.headers.origin
-      const userId = req.session.userId
-      if (!argv.host.allowsSessionFor(userId, origin)) {
-        debug(`Rejecting session for ${userId} from ${origin}`)
-        // Destroy session data
-        delete req.session.userId
-        // Ensure this modified session is not saved
-        req.session.save = (done) => done()
-      }
-      next()
-    })
+    const origin = req.headers.origin
+    const userId = req.session.userId
+    // Exception: allow logout requests from all third-party apps
+    // such that OIDC client can log out via cookie auth
+    // TODO: remove this exception when OIDC clients
+    // use Bearer token to authenticate instead of cookie
+    // (https://github.com/solid/node-solid-server/pull/835#issuecomment-426429003)
+    if (!argv.host.allowsSessionFor(userId, origin) && !isLogoutRequest(req)) {
+      debug(`Rejecting session for ${userId} from ${origin}`)
+      // Destroy session data
+      delete req.session.userId
+      // Ensure this modified session is not saved
+      req.session.save = (done) => done()
+    }
+    next()
   })
 
   let accountManager = AccountManager.from({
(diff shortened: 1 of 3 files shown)


lib/create-app.js
```diff
@@ -169,6 +169,23 @@ function initWebId (argv, app, ldp) {
...
   app.use(sessionHandler)
+  // Reject cookies from third-party applications.
+  // Otherwise, when a user is logged in to their Solid server,
+  // any third-party application could perform authenticated requests
+  // without permission by including the credentials set by the Solid server.
+  app.use((req, res, next) => {
+    const origin = req.headers.origin
+    const userId = req.session.userId
+    // Exception: allow logout requests from all third-party apps
+    if (!argv.host.allowsSessionFor(userId, origin) && !isLogoutRequest(req)) {
```
kjetilk: This was a bit hard to review. Is this actually the only change here?

kidehen:
@RubenVerborgh ,

Re 

> Also, we should note that this fix (as did #834) allows any third party app to log the user out without their explicit permission if they happen to know the user's pod.

I assume that only applies to WebID-OIDC rather than both WebID-OIDC and  WebID-TLS. Basically, if one deployed a TLS-mode pod said issue wouldn't exist, which is what I would expect.

lib/create-app.js
RubenVerborgh: This thing is the only real change.

RubenVerborgh:
> I assume that only applies to WebID-OIDC rather than both WebID-OIDC and WebID-TLS.

Yes.

RubenVerborgh:
Actually, @kidehen, I realize that this issue is not a server-side issue but an OIDC client issue.

At the moment, the logout request is being authenticated through a cookie. And we normally block cross-origin cookies for security purposes, but now explicitly make an exception for logout requests.

However, why don't we authenticate the logout request with an OIDC bearer token? Then none of this is necessary. Or is this against OIDC spec?

lib/create-app.js
```diff
@@ -187,30 +204,20 @@ function initWebId (argv, app, ldp) {
...
   }
 }
 
+/**
+ * Determines whether the given request is a logout request
+ */
+function isLogoutRequest (req) {
+  // TODO: this is a hack that hard-codes OIDC paths,
+  // this code should live in the OIDC module
+  return req.path === '/logout' || req.path === '/goodbye'
```
RubenVerborgh: @kidehen I copied this code from @smalinin, but I actually don't think that goodbye is necessary in this exception list (although it doesn't really hurt). The user needs to be authenticated for logout, but not for goodbye, right?

dmitrizagidulin:
Authenticating with a cookie is a temporary hack. The spec requires the token to be used. This is implemented in oidc-op, btw.

RubenVerborgh:
Thanks @dmitrizagidulin, this is what I suspected. Anything stopping solid-auth-client from using that oidc-op code? We use `logout` on oidc-rp right now.

dmitrizagidulin:
Yeah, solid-auth-client just requires refactoring to use oidc-rp’s logoutRequest() method instead of logout(). (and send a POST to the resulting url)

kidehen:
@dmitrizagidulin,

We are now back to oidc-op, which @cblakeley has been working on. Thus, can we triangulate back to the issue and PR combo that put an end to this issue. 

/cc @RubenVerborgh @smalinin @kjetilk

dmitrizagidulin:
@kidehen Not quite, we’re not back to oidc-op. I was just saying that the correct fix has been already added to oidc-op and -rp. And the last step required is the refactor to auth client.

lib/create-app.js
kidehen: Hmm..
I would suggest the workaround as is for now, we know it works (and it doesn't hurt  in the worst case).  

/cc @smalinin @cblakeley

dmitrizagidulin:
@kidehen 

> Where is the correct fix bearing in mind the oidc-op and -rp fixes you mention?

I'm referring to PRs https://github.com/solid/oidc-op/pull/7 and https://github.com/solid/oidc-rp/pull/6

> so our workaround stays until that work is completed, right?

That seems reasonable to me. We can roll back the cookie changes in this PR afterwards, when the auth client refactor is merged.

kjetilk:
OK, sounds good. @RubenVerborgh please merge at your discretion, we can then roll a release ASAP.

RubenVerborgh:
Thanks all!

RubenVerborgh:
solid-auth-client fix so that cookies are no longer used in https://github.com/solid/solid-auth-client/pull/70

Issue #825, opened 2018-10-02
Provide a more information on the index page for the server itself


justinwb:
Currently if you go to the server (e.g. https://solid.community), you get a completely vanilla index page, with no identifiable detail about the server in question. We need to expose the following (which should be unique per solid server and (probably) pulled from the config.json.

The lack of this information, especially related to who to contact for problems or issues, is leading to many people coming to the general community chat looking for help.

**Server Name** - Friendly name for the server - e.g. (Solid Community Provider)
**Server Description** - Short description of what the  server is about - e.g. (A free provider for the Solid community, by the Solid community).
**Server Logo** - Logo file to present (if available).
**Link to Project Status** - Link to a page at https://solid.inrupt.com that provides a brief status of where major elements of the ecosystem and project currently stand. For example, current state of documentation against the total projected work to be done. Current state of user experience. Current state of security model. Etc.
**Support Contact** - What e-mail address to reach out to for help, problems, issues

megoth:
So a lot of this is quite easy to fix, but I think we should have a discussion on what we want to do with the front page, if anything; @justinwb and the team is working on a very good MVP for a new interface for the front page, so we might want to just wait until then. Or we might want to patch the current front page a bit, to make it a bit more useful for newcomers. 

But I suggest not doing to much, as it is going to replaced at some point anyway.

megoth:
@justinwb Btw, where should we get the friendly name of the server, description, and logo? And which URLs should we use for for project status and support? (I'll try to suggest some as soon as I've worked more on this issue.)

Should it be config-properties that POD providers can fill in manually (i.e. not as part of the `solid init` script) perhaps?

@timbl (to include your requests on #844):
- **More explanation:** Do we have a good text we can use here? And what do you mean with IDS? (Perhaps Identity server? Shouldn't it decentralized identity server, in that case?)
- **LINKS to help:** Which URLs should we include? Perhaps https://solid.inrupt.com/docs ?
- **Links to solid:** I'm guessing we should link to https://solid.mit.edu/ ?
- **Main material which the server operator can make with background pictures and styling and an explanation of what this particular community is, serves, etc, who runs it, where it hosted etc.** This ties into a lot of what @justinwb suggested.

dmitrizagidulin:
@megoth The server index page is created in [`ensureWelcomePage()`](https://github.com/solid/node-solid-server/blob/master/lib/server-config.js#L39) (which is called from `create-app.js`).

[...]

PR #836, opened 2018-10-02
Add prototype warning

kjetilk:
@melvincarvalho suggested a wording to say it is a prototype on the landing page. Seems like a good idea to me, so I made three commits, we may take them all, or any of them. Please review.

### DIFF
--- a/default-templates/server/index.html
+++ b/default-templates/server/index.html
@@ -14,7 +14,7 @@
   <div class="row">
     <div class="col-md-12">
       <p style="margin-top: 3em; margin-bottom: 3em;">
-	This is a prototype implementation of a Solid server.
+	This is a prototype implementation of a <a href="https://solid.inrupt.com/">Solid</a> server.
 
 	It is a fully functional server, but the security and
 	stability of the server cannot be guaranteed.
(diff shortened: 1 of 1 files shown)


default-templates/server/index.html
```diff
@@ -8,13 +8,18 @@
...
 </div>
 <div class="container">
   <div class="row">
     <div class="col-md-12">
       <p style="margin-top: 3em; margin-bottom: 3em;">
-        If you have not already done so, please create an account.
+	This is a prototype implementation of a <a href="https://solid.inrupt.com/">Solid</a> server.
+
+	It is a fully functional server, but the security and
+	stability of the server cannot be guaranteed.
```
RubenVerborgh: I think this is too strongly worded; perhaps "but there are no security or stability guarantees"?

melvincarvalho: Either text looks fine.  

Just to communicate it's alpha or beta (like gmail).  

I dont think there should be a hyperlink in there.

Each pod provider can customize their message.

kjetilk: OK, fine, I think we can save something bigger for #825.

PR #837, opened 2018-10-03
Reset api implementation

gagandeepp:


kjetilk:
Thanks a lot for the PR! Some team members are in transit these days, so it may take some time to get an answer, just so you know. :-)

lib/requests/reset-storage.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
megoth: Please follow the naming convention from before, i.e. postfix the filename with `-request`.

lib/api/authn/webid-oidc.js
```diff
@@ -23,7 +23,7 @@ const {
  * @param app {Object} Express.js app instance
  * @param argv {Object} Config options hashmap
  */
-function initialize (app, argv) {
+function initialize(app, argv) {
```
megoth: Please remove minor changes, such as removing spaces, as they make merging more difficult; We'll have a cleanup-issue at some point to make sure that all files follow a certain format. (We already have linting, but it's not very strict on this.)

lib/requests/reset-storage.js
```diff
@@ -0,0 +1,37 @@
...
+    static reset(req, res) {
+        debug('Reset Cookies & Local Storage from the application')
+        const cookies = req.cookies;
+        for (let prop in cookies) {
+            if (!cookies.hasOwnProperty(prop)) {
+                continue;
+            }
+            res.cookie(prop, '', { expires: new Date(0) })
+        }
+        // if (typeof (localStorage) !== undefined) {
```
megoth: localStorage.clear would have to be run on front-end code, not back-end code such as this.

In that regard, cookies could also be removed via the front-end, but this should work fine as well.

megoth:
@gagandeepp could you link the issue that you're fixing this PR with? There are so many issues to follow at this point, so a little confused =P

lib/requests/reset-storage.js
gagandeepp: fixed (y)

lib/api/authn/webid-oidc.js
gagandeepp: @megoth which linter are you using Jslint/eslint? If not ,can you please create another feature to implement eslint in this code repository,and assign it to me?

dmitrizagidulin:
@gagandeepp Thank you for this PR!

Couple of notes:

#### Linting and tests
We're using [Standard.js](https://standardjs.com/) for linting (it uses eslint under the hood). It's hooked up to the test suite, so make sure that `npm test` passes on your computer, before submitting the PR, that should catch any style issues.

Speaking of tests, we definitely prefer unit and/or integration tests, with PRs like this.

#### Build Fail Log
> can you please provide the build fail log,I am unable to investigate this isue

Anyone can look at the Travis CI failures - just expand the failed checks, click on the Detail, and click on a particular build - you can see why it failed.

#### Clearing cookies
```js
        for (let prop in cookies) {
            if (!cookies.hasOwnProperty(prop)) {
                continue;
            }
            res.cookie(prop, '', { expires: new Date(0) })
        }
```
^ this is not an efficient way to clear a cookie, there's no need to iterate through all the properties. Just use the existing [`res.clearCookie`](http://expressjs.com/en/api.html#res.clearCookie) call.

#### Purpose of PR

All of that said, lets step back and ask - what's the purpose of this PR? It seems to me to just duplicate Logout functionality (the session cookie is cleared when you log out). It doesn't do anything for front-end applications (like clearing localStorage).

dmitrizagidulin:
@megoth 

> We already have linting, but it's not very strict on this

On the contrary, our linting is very strict about this! :) It's the reason the PR is failing on Travis.

RubenVerborgh:
We should have the linting in a pre-commit hook (not the tests though, too slow).

PR #839, opened 2018-10-03
Adding an extra password field to require users to repeat passwords

megoth:
This is a fix for #822 

Handles backend validation and adds a scripts that validates on frontend as well.

PR #842, opened 2018-10-04
Nicer UI and corrections on markup

### LINKED ISSUE #833: Inconsistent use of bootstrap classes
I'm working on register-form.hbs and noticing some strange use of Bootstrap classes, specifically not using `form-group` for group form elements. This has been done in other files. 

I don't want to change more than I have in my commits, hence this issue. Should we make the views consistent, or simply make sure that this is done correctly in later work?

he8us:
Hello,

So I was fiddling a bit in the code of NSS when I discovered the views. Since the code was a bit heavy and IMHO the UI was a bit messy, I decided to rework it.

What I did:
* Changed markup to use bootstrap modules
* Removed unnecessary tags
* Downgraded inline JS to support more browsers
   * Replaced `let` and `const` by `var`
   * Converted `() => {}` to `function(){}`
   * Put code in a closure, to avoid polluting the global scope
* Refactor the views to be a bit more user friendly.
    * Buttons are now colored to me more meaningful
    * Unified page titles
    * Made use of responsive classes of bootstrap to have the login page adapt to the screensize
    * ...

@kjetilk I might have been a bit over excited  and changed much more than what was strictly necessary to be consistent with the classes of bootstrap. In my defense, I had nearly finished when I saw #833. I can wait for yours to be merged, then fix the conflicts.

Fixes #833

### DIFF
--- a/default-views/auth/change-password.hbs
+++ b/default-views/auth/change-password.hbs
@@ -5,61 +5,35 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Change Password</title>
   <link rel="stylesheet" href="/common/css/bootstrap.min.css">
+  <link rel="stylesheet" href="/common/css/server.css">
 </head>
 <body>
 <div class="container">
-  <h4>Change Password</h4>
-</div>
-<div class="container">
-  <form method="post" action="/account/password/change">
-    {{#if error}}
-      <div class="form-group">
-        <div class="row">
-          <div class="col-md-12">
-            <p class="text-danger"><strong>{{error}}</strong></p>
-          </div>
-        </div>
-      </div>
-    {{/if}}
 
-    {{#if validToken}}
-      <div class="form-group">
-        <div class="row">
-          <div class="col-md-12">
-            <label for="newPassword">New Password:</label>
+  <div class="page-header">
+    <h1>Change Password</h1>
+  </div>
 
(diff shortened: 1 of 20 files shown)


he8us:
Waiting for #839 to be merged, then fix conflicts

kjetilk:
Hehe, cool! Thanks a lot for jumping in!

So, the frontend stuff is not really my field, so I'll let the others review! :-)

kjetilk:
Oh, BTW, I guess I didn't say, please base on `develop` for the short term stuff, and `release/v5.0.0` for things that should wait for a new feature release. :-)

default-views/auth/change-password.hbs
```diff
@@ -5,61 +5,35 @@
...
+        <input type="password" class="form-control" name="newPassword"/>
       </div>
-    {{/if}}
 
-    {{#if validToken}}
-      <div class="form-group">
-        <div class="row">
-          <div class="col-md-12">
-            <label for="newPassword">New Password:</label>
+      <button type="submit" class="btn btn-primary">Change Password</button>
```
megoth: Not directly related, but more of a note to self: We've added repeat password on another page, we probably should do that here as well.

But for this case, we need to find a good markup of the password component. I've used two divs with `col-md-3`in the previous design, so might revert something a bit here, unless you have a better suggestion?

RubenVerborgh:
Introduced bug #862.

PR #845, opened 2018-10-05
Password Validator

mintunitish:
Validate the password before registering a user: 

- Checks for OWASP specification based tests.
- Checks for pwned passwords.

Resolves [163](https://github.com/solid/solid/issues/163l)

### DIFF
--- a/default-views/account/register.hbs
+++ b/default-views/account/register.hbs
@@ -27,14 +27,56 @@
     {{> account/register-form}}
   {{/if}}
 </div>
-<script src="https://raw.githubusercontent.com/nowsecure/owasp-password-strength-test/master/owasp-password-strength-test.js" defer></script>
+<script src="https://raw.githubusercontent.com/nowsecure/owasp-password-strength-test/master/owasp-password-strength-test.js"
+        defer></script>
 <script>
-    function validatePassword () {
+    function validatePasswordBeforeSubmit (e) {
+        e.preventDefault();
+        const pwdErrorDiv = document.getElementById('passwordHelp');
         let pw = document.getElementById('password').value;
-        let owaspCheck = owaspPasswordStrengthTest.test(pw);
-        //check for errors and show
-        // call pwned api and check
-        //submit the form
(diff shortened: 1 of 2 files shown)


default-views/account/register.hbs
```diff
@@ -21,5 +21,57 @@
...
 </div>
+<script src="https://raw.githubusercontent.com/nowsecure/owasp-password-strength-test/master/owasp-password-strength-test.js"
+        defer></script>
+<script>
+    function validatePasswordBeforeSubmit (e) {
+        e.preventDefault();
+        const pwdErrorDiv = document.getElementById('passwordHelp');
+        let pw = document.getElementById('password').value;
+        let owaspCheck = owaspPasswordStrengthTest.test(pw)
+        if (owaspCheck.strong === true) {
```
megoth: Is this necessary? `owaspCheck.strong` will always be truish in the case that it is valid, no?

default-views/account/register.hbs
```diff
@@ -21,5 +21,57 @@
     {{> account/register-form}}
   {{/if}}
 </div>
+<script src="https://raw.githubusercontent.com/nowsecure/owasp-password-strength-test/master/owasp-password-strength-test.js"
+        defer></script>
+<script>
+    function validatePasswordBeforeSubmit (e) {
+        e.preventDefault();
+        const pwdErrorDiv = document.getElementById('passwordHelp');
```
megoth: This is a not an ID that is present in the markup yet? (I cannot find it, at least)

mintunitish: Looks like I forgot to add a file to the PR. Will do it asap.

default-views/account/register.hbs
mintunitish: It won't be always true. If the owasp tests failed it will return false

mintunitish:
Everything should be working now.

mintunitish:
There's some image file that failing the tests. I think the tests need to be altered too?

PR #848, opened 2018-10-06
Add logout button to access denied page

RubenVerborgh:
Unfortunately it doesn't seem to work because server-side logout does not seem to work (WebID is not deleted from session).

### DIFF
--- a/default-views/auth/no-permission.hbs
+++ b/default-views/auth/no-permission.hbs
@@ -17,8 +17,18 @@
       You are currently logged in as <code>{{webId}}</code>,
       but do not have permission to access <code>{{currentUrl}}</code>.
     </p>
+    <p>
+      <button class="btn btn-danger" onclick="logout()">Log out</button>
+    </p>
   </div>
 </div>
 </div>
+<script src="/common/js/solid-auth-client.bundle.js"></script>
+<script>
+  async function logout () {
+    await solid.auth.logout()
+    location.reload()
+  }
+</script>
 </body>
 </html>
--- a/default-views/auth/no-permission.hbs
+++ b/default-views/auth/no-permission.hbs
@@ -18,17 +18,12 @@
       but do not have permission to access <code>{{currentUrl}}</code>.
     </p>
     <p>
-      <button class="btn btn-danger" onclick="logout()">Log out</button>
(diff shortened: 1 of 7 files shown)


RubenVerborgh:
@kidehen Is logout working for you?

RubenVerborgh:
Blocked by https://github.com/solid/oidc-rp/issues/21

default-views/auth/no-permission.hbs
```diff
@@ -17,8 +17,18 @@
       You are currently logged in as <code>{{webId}}</code>,
       but do not have permission to access <code>{{currentUrl}}</code>.
     </p>
+    <p>
+      <button class="btn btn-primary" onclick="logout()">Log out</button>
```
rubensworks: Perhaps we can use `btn-danger` here instead, for consistency with the other logout buttons.

kjetilk: Consistency is a good thing. :-)

RubenVerborgh:
Unblocked, new reviews welcome.

RubenVerborgh:
Improvement coming 🙂

RubenVerborgh:
Now with d7a1c0bb6bdfd0e52fcd4e74e9a74d8efb0c1ceb, which ensures that all pages uses the same login/logout code. This a) removes duplication b) ensures the latest code (since we cannot change home pages once people have created them) and c) ensures that logging in on the homepage is materialized into a cookie and thus carried over to other pages.

megoth:
> @megoth And now even better with the latest commit, removing `onclick` as remarked by @csarven.

Yeah, I'm a bit on the fence there, tbh; I'm kinda enjoying onclick as it makes it much easier to debug later on. But it does require adding methods to the global scope, which is A Bad Thing [tm] in general (for various reasons we don't need to get into here). In this case it's a small application, so it's probably better to use addEventListener (for bigger applications I recommend having a convention to handlers addEventListener makes use of, e.g. prefix `js-` for CSS classes).

But this won't be a problem with applications like React, so I'll drop my rant here =P

RubenVerborgh:
> But it does require adding methods to the global scope

And also, it would tie homepages to a specific version of the auth-button script, which would prevent us from updating that script in the future. (Related: #936.)

PR #854, opened 2018-10-10
Adding script clean and adding it to solid script

megoth:
For my own sanity - I keep forgetting to remove config files if I want to get updates from default-templates and default-views.

Ignore this if it's to invasive

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "solid": "npm run clean && node ./bin/solid",
+    "solid": "node ./bin/solid",
     "standard": "standard '{bin,examples,lib,test}/**/*.js'",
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
@megoth the script looks good. My only question is - does it make sense to have the `clean` script automatically run whenever you launch the server? (Not that I expect a lot of people use the `npm run solid` script instead of launching it directly via node).

kjetilk:
Yeah, that was my question too, perhaps only do it if launched explicitly?

dmitrizagidulin:
I think what I'm trying to say is - maybe it's just enough to add a `clean` script. so that testers and devs can explicitly launch with `npm run clean && npm run solid`? Without hitching it to the run script?

megoth:
Yeah, I'm ok with it not being part of the `solid`-script; it was merely a convenience for my own sake ^_^ I'll update the PR very soon

megoth:
@kjetilk @dmitrizagidulin I've updated the PR ^_^

PR #859, opened 2018-10-12
Add a repeat password field and give a visual feedback on the strength

he8us:
This PR is meant to replace #839 which fixes the conflicts created with the merge of #842 and add a password strength meter + repeat password field

Based on the code of @megoth for the second password field and @mintunitish for the password validation functions.

Their code were refactored to have a consistent codebase.

I also had to made a local copy of the owasp-password-strength-test.js because chrome didn't wanted to execute the code since the file is served as 'text/plain'.

I added the glyphicons to be able to show a quick feedback icon on the fields.

I also externalized the JS code in it's own file to be able to compress it if needed later on.

### DIFF
--- a/common/css/server.css
+++ b/common/css/server.css
@@ -1,4 +1,0 @@
-.panel-login-tls,
-.panel-already-registered{
-    text-align: center;
-}
--- a/common/js/solid.js
+++ b/common/js/solid.js
@@ -1,8 +1,8 @@
-/* global owaspPasswordStrengthTest, TextEncoder, crypto */
+/* global owaspPasswordStrengthTest, TextEncoder, crypto, fetch */
 (function () {
   'use strict'
 
-  var PasswordValidator = function (passwordField, repeatedPasswordField) {
+  const PasswordValidator = function (passwordField, repeatedPasswordField) {
     if (
       passwordField === null || passwordField === undefined ||
       repeatedPasswordField === null || repeatedPasswordField === undefined
@@ -20,23 +20,23 @@
     this.errors = []
   }
 
-  PasswordValidator.prototype.FEEDBACK_SUCCESS = 'success'
-  PasswordValidator.prototype.FEEDBACK_WARNING = 'warning'
-  PasswordValidator.prototype.FEEDBACK_ERROR = 'error'
+  const FEEDBACK_SUCCESS = 'success'
+  const FEEDBACK_WARNING = 'warning'
+  const FEEDBACK_ERROR = 'error'
 
-  PasswordValidator.prototype.ICON_SUCCESS = 'glyphicon-ok'
-  PasswordValidator.prototype.ICON_WARNING = 'glyphicon-warning-sign'
-  PasswordValidator.prototype.ICON_ERROR = 'glyphicon-remove'
+  const ICON_SUCCESS = 'glyphicon-ok'
+  const ICON_WARNING = 'glyphicon-warning-sign'
+  const ICON_ERROR = 'glyphicon-remove'
 
-  PasswordValidator.prototype.VALIDATION_SUCCESS = 'has-success'
-  PasswordValidator.prototype.VALIDATION_WARNING = 'has-warning'
-  PasswordValidator.prototype.VALIDATION_ERROR = 'has-error'
+  const VALIDATION_SUCCESS = 'has-success'
+  const VALIDATION_WARNING = 'has-warning'
+  const VALIDATION_ERROR = 'has-error'
 
-  PasswordValidator.prototype.STRENGTH_PROGRESS_0 = 'progress-bar-danger level-0'
-  PasswordValidator.prototype.STRENGTH_PROGRESS_1 = 'progress-bar-danger level-1'
-  PasswordValidator.prototype.STRENGTH_PROGRESS_2 = 'progress-bar-warning level-2'
-  PasswordValidator.prototype.STRENGTH_PROGRESS_3 = 'progress-bar-success level-3'
-  PasswordValidator.prototype.STRENGTH_PROGRESS_4 = 'progress-bar-success level-4'
+  const STRENGTH_PROGRESS_0 = 'progress-bar-danger level-0'
+  const STRENGTH_PROGRESS_1 = 'progress-bar-danger level-1'
+  const STRENGTH_PROGRESS_2 = 'progress-bar-warning level-2'
+  const STRENGTH_PROGRESS_3 = 'progress-bar-success level-3'
+  const STRENGTH_PROGRESS_4 = 'progress-bar-success level-4'
 
   /**
    * Prefetch all dom nodes at initialisation in order to gain time at execution since DOM manipulations
@@ -77,8 +77,8 @@
    * Validate password on the fly to provide the user a visual strength meter
    */
   PasswordValidator.prototype.instantFeedbackForPassword = function () {
-    var passwordStrength = this.getPasswordStrength(this.passwordField.value)
-    var strengthLevel = this.getStrengthLevel(passwordStrength)
+    const passwordStrength = this.getPasswordStrength(this.passwordField.value)
+    const strengthLevel = this.getStrengthLevel(passwordStrength)
 
     if (this.currentStrengthLevel === strengthLevel) {
       return
@@ -98,8 +98,8 @@
    */
   PasswordValidator.prototype.validatePassword = function () {
     this.errors = []
-    var password = this.passwordField.value
-    var passwordStrength = this.getPasswordStrength(password)
+    const password = this.passwordField.value
+    const passwordStrength = this.getPasswordStrength(password)
     this.currentStrengthLevel = this.getStrengthLevel(passwordStrength)
 
     if (passwordStrength.errors) {
@@ -135,7 +135,7 @@
   }
 
   PasswordValidator.prototype.setPasswordFeedback = function () {
-    var feedback = this.getFeedbackFromLevel()
+    const feedback = this.getFeedbackFromLevel()
     this.updateStrengthMeter()
     this.displayPasswordErrors()
     this.setFeedbackForField(feedback, this.passwordField)
@@ -145,7 +145,7 @@
    * Update the repeated password feedback icon and color
    */
   PasswordValidator.prototype.updateRepeatedPasswordFeedback = function () {
-    var feedback = this.checkPasswordFieldsEquality() ? this.FEEDBACK_SUCCESS : this.FEEDBACK_ERROR
+    const feedback = this.checkPasswordFieldsEquality() ? FEEDBACK_SUCCESS : FEEDBACK_ERROR
     this.setFeedbackForField(feedback, this.repeatedPasswordField)
   }
 
@@ -155,8 +155,8 @@
    * @param {HTMLElement} field
    */
   PasswordValidator.prototype.setFeedbackForField = function (feedback, field) {
-    var formGroup = this.getFormGroupElementForField(field)
-    var visualFeedback = this.getFeedbackElementForField(field)
+    const formGroup = this.getFormGroupElementForField(field)
+    const visualFeedback = this.getFeedbackElementForField(field)
 
     this.resetValidation(formGroup)
     this.resetFeedbackIcon(visualFeedback)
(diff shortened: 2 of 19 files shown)


he8us:
This PR implementation is not finished yet because I have 2 questions

1. I didn't block the form submission in case of the password is weak, should I? If yes, at which level?
2. Which are the browser we need to support? Because, we use the [Crypto Web API](https://developer.mozilla.org/en-US/docs/Web/API/Crypto) which is not available in Internet Explorer.

mintunitish:
I think blocking the form submission is handled by calling the onsubmit event, the behavior should be same. 
And for the crypto API we can probably work with a polyfill for IE.

mintunitish:
The optimal behavior should be to assist the user to create a strong password rather than restricting to use only strong passwords, after all at the end it's the user's account.

Displaying a password strength meter or suggestion message will be good and if still the user wants to go ahead with weak password then let it be.

kjetilk:
> The optimal behavior should be to assist the user to create a strong password rather than restricting to use only strong passwords, after all at the end it's the user's account.

Yeah, especially since password meters are not necessarily all that good. However, I can see a case for rejecting passwords that occur above a certain frequency in the leaked passwords list. Any match in there should result in a red warning saying that this password has been leaked, with a link to the source.

common/js/server.js
```diff
@@ -0,0 +1,462 @@
+/* global owaspPasswordStrengthTest, TextEncoder */
```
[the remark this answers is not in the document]
he8us: Well since it's a file I created and written based on JS code I extracted from `default-views/account/register-form.hbs` I don't know if we need to put the license since we don't have other file with a license header. If so, should I copy/paste the content of `LICENSE`?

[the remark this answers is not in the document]
RubenVerborgh: Ah this file is your work entirely? Wasn't sure if it was a generic password tester.

common/js/server.js
```diff
@@ -0,0 +1,462 @@
...
+      case 0:
+      case 1:
+        return this.FEEDBACK_ERROR
+      case 2:
+        return this.FEEDBACK_WARNING
+      case 3:
+      case 4:
+        return this.FEEDBACK_SUCCESS
+    }
+  }
```
RubenVerborgh: The above code can be replaced by a hash/object.

common/js/server.js
```diff
@@ -0,0 +1,462 @@
...
+      case 1:
+        return this.STRENGTH_PROGRESS_1
+      case 2:
+        return this.STRENGTH_PROGRESS_2
+      case 3:
+        return this.STRENGTH_PROGRESS_3
+      case 4:
+        return this.STRENGTH_PROGRESS_4
+    }
+  }
```
RubenVerborgh: The above code can be replaced by a hash/object.

common/js/server.js
```diff
@@ -0,0 +1,462 @@
...
+  PasswordValidator.prototype.getFeedbackIconClass = function (feedback) {
+    switch (feedback) {
+      case this.FEEDBACK_SUCCESS:
+        return this.ICON_SUCCESS
+      case this.FEEDBACK_WARNING:
+        return this.ICON_WARNING
+      case this.FEEDBACK_ERROR:
+        return this.ICON_ERROR
+    }
+  }
```
RubenVerborgh: The above code can be replaced by a hash/object.

common/js/server.js
```diff
@@ -0,0 +1,462 @@
...
+  PasswordValidator.prototype.getValidationClass = function (feedback) {
+    switch (feedback) {
+      case this.FEEDBACK_SUCCESS:
+        return this.VALIDATION_SUCCESS
+      case this.FEEDBACK_WARNING:
+        return this.VALIDATION_WARNING
+      case this.FEEDBACK_ERROR:
+        return this.VALIDATION_ERROR
+    }
+  }
```
RubenVerborgh: The above code can be replaced by a hash/object.

common/css/server.css
```diff
@@ -2,3 +2,39 @@
 .panel-already-registered{
     text-align: center;
 }
+
```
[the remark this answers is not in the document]
he8us: In `common/css/bootstrap.min.css:5:3066`, they are part of bootstrap but were not installed with it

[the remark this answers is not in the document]
RubenVerborgh: Makes sense, thanks!

he8us:
As stated in [my first comment](https://github.com/solid/node-solid-server/pull/859#issuecomment-429343264) this PR is not yet ready to review code wise but to raise comments and reflection.

In all cases I think it's not save to rely only on front-end code to make security assumptions, so if we want to block the form submission, we also need to validate the password server-side.

During my development I also noticed that we don't use any package manager for the front end code: we only have boostrap CSS and not the icon font delivered with it. So I added them but now we have license issues. That's why I created #863

common/js/polyfills.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
he8us: This polyfill script is still a question mark. Because now it's polyfilling the TextEncoder used to check is the password was leaked. It was introduced by #845 

TextEncoder is not available in Edge, IE nor Safari so I included the polyfill from MDN: https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#Browser_compatibility

Now that #863 is merged, I was thinking to load the polyfill from NPM.

* Can I Use table for the TextEncoder: https://caniuse.com/#feat=textencoder

That also rose the issue of which are the browser we need to support. I'm pretty sure we need to support Edge and Safari but I don't know about IE11? (At work, since we do B2B, we still need it). If we need to support IE11, I still need to figure out how to generate the SHA1 and to replace fetch to XMLHttpRequest.

* Can I Use table for the web cryptography: https://caniuse.com/#feat=cryptography
* Can I use table for fetch https://caniuse.com/#feat=fetch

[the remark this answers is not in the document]
RubenVerborgh: Let's indeed use npm!

This is my viewpoint regarding browser compatibility: https://github.com/solid/solid-auth-client/blob/v2.2.8/.browserslistr, resulting in https://browserl.ist/?q=node+8%2C+%3E+1%25%2C+not+ie+%3C%3D+11%2C+not+operamini+all

he8us:
@RubenVerborgh Changes applied

common/js/solid.js
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    return 4
+  }
+
+  PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP = []
+  PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP[0] = PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP[1] = PasswordValidator.prototype.FEEDBACK_ERROR
+  PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP[2] = PasswordValidator.prototype.FEEDBACK_WARNING
+  PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP[3] = PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP[4] = PasswordValidator.prototype.FEEDBACK_SUCCESS
+
```
RubenVerborgh: Just this would do:

```
PasswordValidator.prototype.LEVEL_TO_FEEDBACK_MAP = [
   FEEDBACK_ERROR, FEEDBACK_ERROR, FEEDBACK_WARNING, FEEDBACK_SUCCESS
]
```

if we additionally expose those as variables. Maybe they don't even need to be on the prototype.

common/js/solid.js
```diff
@@ -0,0 +1,440 @@
...
+  PasswordValidator.prototype.getFeedbackFromLevel = function () {
+    return this.LEVEL_TO_FEEDBACK_MAP[this.currentStrengthLevel]
+  }
+
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP = []
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP[0] = PasswordValidator.prototype.STRENGTH_PROGRESS_0
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP[1] = PasswordValidator.prototype.STRENGTH_PROGRESS_1
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP[2] = PasswordValidator.prototype.STRENGTH_PROGRESS_2
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP[3] = PasswordValidator.prototype.STRENGTH_PROGRESS_3
+  PasswordValidator.prototype.LEVEL_TO_PROGRESS_MAP[4] = PasswordValidator.prototype.STRENGTH_PROGRESS_4
```
RubenVerborgh: Idem as above.

common/js/solid.js
```diff
@@ -0,0 +1,440 @@
...
+  PasswordValidator.prototype.addPasswordError = function (error) {
+    if (Array.isArray(error)) {
+      for (var i = 0, ln = error.length; i < ln; i++) {
+        this.addPasswordError(error[i])
+      }
+      return
+    }
+
+    this.errors.push(error)
+  }
```
RubenVerborgh: the above can be replaced by

```
this.errors.push(...(Array.isArray(error) ? error : [error]))
```

or just simply eliminating the call and doing the above inline

common/js/solid.js
```diff
@@ -0,0 +1,440 @@
...
+
+  PasswordValidator.prototype.displayPasswordErrors = function () {
+    this.passwordHelpText.innerHTML = '<p>' + this.errors.join('</p><p>') + '</p>'
```
RubenVerborgh: SECURITY: html injection in error

common/js/solid.js
```diff
@@ -0,0 +1,440 @@
...
+
+  PasswordValidator.prototype.VALIDATION_SUCCESS = 'has-success'
+  PasswordValidator.prototype.VALIDATION_WARNING = 'has-warning'
+  PasswordValidator.prototype.VALIDATION_ERROR = 'has-error'
+
+  PasswordValidator.prototype.STRENGTH_PROGRESS_0 = 'progress-bar-danger level-0'
+  PasswordValidator.prototype.STRENGTH_PROGRESS_1 = 'progress-bar-danger level-1'
+  PasswordValidator.prototype.STRENGTH_PROGRESS_2 = 'progress-bar-warning level-2'
+  PasswordValidator.prototype.STRENGTH_PROGRESS_3 = 'progress-bar-success level-3'
+  PasswordValidator.prototype.STRENGTH_PROGRESS_4 = 'progress-bar-success level-4'
```
RubenVerborgh: Lett's just have the above as `const`, not on the prototype.

RubenVerborgh:
Maybe just some general feedback: overall, I think there a lot of lines of code are used to achieve things. Since the number of bugs grows proportionally with the number of lines of code (not number of ideas expressed), it is good to strive for efficiency whenever meaningful.

common/js/solid.js
he8us: An whole new world opens to me without the need to support IE11 :D

RubenVerborgh: Ah right, I am used to transpiling, so I don't know what browsers need what. However, since no transpiling is in effect here, we might need to be more careful.

he8us:
@RubenVerborgh I do agree, for the next PR I will try not to. I went this way because I don't know the level of javascript-fu of the maintainers. I usually write more concise code but it might quickly become non understandable for other team members with other background than me: at work I'm the only JS developer, the other 7 are PHP or Java devs.

common/js/solid.js
he8us: fixed

common/js/solid.js
he8us: Fixed

common/js/solid.js
he8us: Fixed

common/js/solid.js
he8us: Fixed

common/js/solid.js
he8us: Fixed

he8us:
Just rebased develop

Issue #823, opened 2018-10-02
Explain how username is leveraged on registration page


justinwb:
We've received numerous reports of users not realizing that their username would be used publicly. It isn't obvious to them that it becomes the public sub-domain associated with their account, and forms the basis of their public webid. We need to provide a bit of explanation here, and I think we do it in two parts...

1. As users are typing in their username, have a bit of text updating as they type with:
Your public Solid POD URL will be: http://[username].hostname.com
Your public Solid WebID will be: https://[username].hostname.com/profile/card#me

2. Then provide a bit of explanation on what these things actually are...

(Something like... but feel free to tweak)

Your POD URL is like the homepage for your solid pod. By default, it is readable by the public, but you can always change that if you like by changing the access control.

Your Solid WebID is your globally unique identity that you can use to identify and authenticate yourself with other PODs across the world.

kjetilk:
Right, but this isn't universally true, it is just for multi-user installs, so I think we should only display this message for multi-user installs.

kjetilk:
Issues such as should also go in here: https://github.com/solid/solid/issues/205

melvincarvalho:
Good point.  It's worth noting that nothing in solid says that anything in your pod needs to be public.  It's just the node solid server defaults.  We can explain this to users.  

The slight confusion occurs because registration does two things

- creates a pod and public folder
- creates a public webid for those that dont have one (99% of users)

Longer term when signup is an app, we'll have better ways to communicate these things, I suspect!

kjetilk:
Not sure while this is still open, I think it is fixed and released, so probably I just forgot to close it.

PR #864, opened 2018-10-16
Improve explanations of usernames

kjetilk:
This should implement #823 and also address (at least partially) the bug in https://github.com/solid/solid/issues/205 by downcasing usernames on submition.

### DIFF
--- a/default-views/account/register-form.hbs
+++ b/default-views/account/register-form.hbs
@@ -10,12 +10,12 @@
             <input type="text" class="form-control" name="username" id="username" placeholder="alice"
                    required/>
 
-	    {{#if multiuser}}
+
 	    <p>Your username should be a lower-case word with only
 	    letters a-z and numbers 0-9 and without periods.</p>
-	    <p>Your public Solid POD URL will be: <tt>https://[username].hostname.com</tt></p>
+	    <p>Your public Solid POD URL will be: <tt>https://<span class="editable-username">alice</span>.hostname.com</tt></p>
 	    <p>Your public Solid WebID will be:
-	    <tt>https://[username].hostname.com/profile/card#me</tt></p>
+	    <tt>https://<span class="editable-username">alice</span>.hostname.com/profile/card#me</tt></p>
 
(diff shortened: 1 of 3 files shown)


default-views/account/register-form.hbs
```diff
@@ -9,7 +9,30 @@
...
                    required/>
-          </div>
+
+	    {{#if multiuser}}
+	    <p>Your username should be a lower-case word with only
+	    letters a-z and numbers 0-9 and without periods.</p>
+	    <p>Your public Solid POD URL will be:
+	    <tt>https://<span class="editable-username">alice</span>.<script type="text/javascript">
+              document.write(window.location.host)
+	    </script></tt></p>
```
RubenVerborgh: We could probably do this with server-side templating (passing in the hostname from `renderForm` in `create-account-request.js` if not already passed in). But this works too.

kjetilk: Yeah, I looked into that, but it seems we'd have to hook in the `SolidHost` class to do it, and I figured it is not worth it.

kjetilk:
Right. It is just that stuff seems to happen every time I rebase, and I end up having to do a merge anyway...

RubenVerborgh:
@kjetilk Ping me if I can help.

kjetilk:
Yup, I will @RubenVerborgh , just need to investigate a bit more myself, to see if I understand the pain point.

BTW, the Snyk failure, is there something we could cherry-pick for that?

RubenVerborgh:
@kjetilk Just a rebase on develop will do. The following should work:
```
git fetch
git checkout fix-username-docs
git rebase -i origin/develop
```

Then pick only the commits you need.

kjetilk:
OK, so perhaps this is what happened: I merged the 5.0.0 branch into it...? If so, that was certainly an accident. I probably meant to merge develop into the 5.0.0 at that point.

Anyway, let me see if I can rebase it successfully with that understanding.

kjetilk:
OK, so how about now?

kjetilk:
OK, I usually look at commits as the gory details, but let me do that to experience the difference :-)

PR #871, opened 2018-10-18
Align landing page login experience

RubenVerborgh:
The login button of the server landing page (https://solid.community/) currently only logs you in to the server, not the client.

This is different from the login experience of a pod page (https://ruben.solid.community/), which was upgraded in #751.

This PR fixes this, such that logging in on https://solid.community/ has the same effect as logging in on https://ruben.solid.community/.

### DIFF
--- a/default-templates/server/index.html
+++ b/default-templates/server/index.html
@@ -11,13 +11,9 @@
 <div class="container">
   <div class="page-header">
     <div class="pull-right">
-      <a href="/register" class="btn btn-primary">
-        Register
-      </a>
+      <a href="/register" class="btn btn-primary">Register</a>
 
-      <a href="/login" class="btn btn-success">
-        Login
-      </a>
+      <a href="/login" onclick="login()" class="btn btn-success">Login</a>
     </div>
 
     <h1>Welcome to the Solid Prototype</h1>
@@ -30,6 +26,41 @@
 
     If you have not already done so, please create an account.
   </p>
+
+  <p class="lead hidden" id="loggedIn">
+    You are logged in as
+    <a href="#" id="profileLink"></a>.
+  </p>
 </div>
+<script src="/common/js/solid-auth-client.bundle.js"></script>
+<script>
+  const elements = {};
+  ['loggedIn', 'profileLink'].forEach(id => {
+    elements[id] = document.getElementById(id)
+  })
+
+  async function login () {
+    const session = await solid.auth.popupLogin()
+    if (session) {
+      // Make authenticated request to the server to establish a session cookie
+      const {status} = await solid.auth.fetch(location)
+      if (status === 401) {
+        alert(`Invalid login.`)
+        await solid.auth.logout()
+      }
+    }
+  }
+
+  solid.auth.trackSession(async session => {
+    if (!session) {
+      elements.loggedIn.classList.add('hidden')
+    }
+    else {
+      elements.loggedIn.classList.remove('hidden')
+      elements.profileLink.innerText = session.webId
+      elements.profileLink.href = session.webId
+    }
+  })
+</script>
 </body>
 </html>


default-templates/server/index.html
```diff
@@ -30,6 +26,41 @@ <h1>Welcome to the Solid Prototype</h1>
...
   </p>
+
+  <p class="lead hidden" id="loggedIn">
+    You are logged in as
+    <a href="#" id="profileLink"></a>.
+  </p>
 </div>
+<script src="/common/js/solid-auth-client.bundle.js"></script>
+<script>
+  const elements = {};
```
megoth: We should be vary of using syntax such as const and let in the client, as these are not supported by all browsers: https://caniuse.com/#search=const

Which begs the question: Which browsers do we officially support?

kjetilk:
Weird, Travis reports a syntax error, but only for Node 8. Could you please check that?

default-templates/server/index.html
```diff
@@ -30,6 +26,41 @@ <h1>Welcome to the Solid Prototype</h1>
...
+  ['loggedIn', 'profileLink'].forEach(id => {
```
megoth: Good support for `Array.prototype.foreach`, but same question as above: https://caniuse.com/#search=forEach

default-templates/server/index.html
```diff
@@ -30,6 +26,41 @@ <h1>Welcome to the Solid Prototype</h1>
...
+    elements[id] = document.getElementById(id)
+  })
+
+  async function login () {
```
megoth: Same as earlier: https://caniuse.com/#search=async

default-templates/server/index.html
```diff
@@ -11,13 +11,9 @@
...
     <div class="pull-right">
-      <a href="/register" class="btn btn-primary">
-        Register
-      </a>
+      <a href="/register" class="btn btn-primary">Register</a>
 
-      <a href="/login" class="btn btn-success">
-        Login
-      </a>
+      <a onclick="login()" class="btn btn-success">Login</a>
```
megoth: Also, this should be changed to a button-element; links with no href has no semantic meaning

default-templates/server/index.html
RubenVerborgh: This PR copies over functionality from `default-views/auth/login-required.hbs`, which uses the same syntax. So I agree with the comment, but this decision has already been made elswhere, i.e., this PR is not breaking things that were working before.

default-templates/server/index.html
RubenVerborgh: That's green everywhere :-)

default-templates/server/index.html
RubenVerborgh: Already in `default-views/auth/login-required.hbs`

default-templates/server/index.html
RubenVerborgh: Myeah they should've been buttons in the first place I suppose. I'll just keep the href.

default-templates/server/index.html
megoth: Ok, do we know where that decision is documented?

default-templates/server/index.html
megoth: Yeah, but not for older versions of IE...

RubenVerborgh:
> As I'm not sure which browsers we officially support, I'm leaving this as "Request changes", to make sure that we agree on which syntax to use.

This PR does not use syntax that is not used elsewhere already; in particular, the same login functionality is already implemented in this syntax.

> Should we maybe introduce a transpiler that builds our front-end resources, so that we can use more modern syntax but make sure that it's transpiled to syntax that is better supported by browsers?

We should, but it's not going to be easy with the current code structure.

default-templates/server/index.html
megoth: I'm sorry that I haven't raised concerns on this earlier, but I wasn't part of the review-process for the code that you refer to.

default-templates/server/index.html
RubenVerborgh: `forEach` is the least of our problems there, many more things are broken. Furthermore, `solid-auth-client` itself (loaded on this page) doesn't support IE: https://github.com/solid/solid-auth-client/blob/v2.2.8/.browserslistrc So it will not work regardless of what syntax I use here.

default-templates/server/index.html
RubenVerborgh: Your concerns are definitely valid. It's just that other browsers are already broken by other pages, so this PR will not make a difference. Even if I'd make this page compatible with (let's say) IE9, the very next page they land on will not be.

megoth:
> We should, but it's not going to be easy with the current code structure.

I agree.

I'll create a separate issue for this, and see where the ball falls on whether or not this should be done.

RubenVerborgh:
Okay. I would propose merging this one (if no other objections), since it follows the same compatibility of other scripts; but we should have a discussion then on what compatibility we want.

kjetilk:
Are you good, @megoth ?

PR #875, opened 2018-10-19
Quick-fix to get more semantically correct markup

megoth:
Just a minor quick-fix

### DIFF
--- a/default-templates/server/index.html
+++ b/default-templates/server/index.html
@@ -13,7 +13,7 @@
     <div class="pull-right">
       <a href="/register" class="btn btn-primary">Register</a>
 
-      <a onclick="login()" class="btn btn-success">Login</a>
+      <button onclick="login()" class="btn btn-success" type="button">Login</button>
     </div>
 
     <h1>Welcome to the Solid Prototype</h1>


RubenVerborgh:
@megoth I had amended a link to #871 as per your comment there, but noticed then that this broke things, so left it off for speed reasons (we had quite the rush yesterday with the release).

megoth:
> @megoth I had amended a link to #871 as per your comment there, but noticed then that this broke things, so left it off for speed reasons (we had quite the rush yesterday with the release).

Ah ok, I was unsure whether or not the comment had come through, so pushed this quick commit to tackle it. Didn't register that it broke myself though, what happened?

> Agree of course, but can we be consistent and make the other one a button too?

If it is a link (i.e. having a href-attribute set), it should be semantically marked as a link. We can use the classes to make them look the same. (I'm accepting distinguishing between semantics and presentation in this regard.)

> Is the type=button really needed?

The default value it type="submit", which might trigger a submit-event if element is parented by a form-element. In this case it is not, but semantically I would argue that it doesn't make sense to mark the button as a submit-button anyway.

In general I agree that the level of scrutiny for this work is low, as there are more important work to be done. But the mixing of a- and button-elements are kind of a pet peeve for me (so, so many don't care about the difference, but it hurts accessibility if we don't treat our markup well), so I just thought I would throw something in there; but of course, now we've spent quite some time discussing this, so wasn't well thought through on my part =P

RubenVerborgh:
> Didn't register that it broke myself though, what happened?

I tackled your comment by putting back the href attribute, so it was an actual link again. But I forgot to cancel the link action when the JavaScript kicked in.

megoth:
> For instance, this login button? Should just be https://github.com/solid/react-components/blob/master/src/components/LoginButton.jsx. That code, in contrast to this, is actually tested, valid HTML, no onClick, etc.

I think this looks good! Easy to read and comprehend, and simple, which is an important feature ^_^

(As a sidenote: I'm personally liking onclick, as eventListeners bound elsewhere might be difficult to track down sometimes. But in plain JS it does require functionality to be exposed on the global scope, which I'm not happy about. So using frameworks like React is really good for this.)

megoth:
I don't think this change is needed anymore, so I'll close the PR.

RubenVerborgh:
The suggestions by @megoth and @csarven have both been implemented in #848.

PR #869, opened 2018-10-17
Feature/server-info

megoth:
This PR solves #825 .

- Adds options to init script, that allows those setting up servers to give some human readable info to their visitors
- Makes use of that data to populate a section for the default server index page

In addition:

- It includes a bug fix (didn't copy all of the content in `default-templates/server` to the `data/[server dir]`)
- Some refactoring to reuse existing functionality
- Suggesting where to put util-methods (common methods that are reusable across the system)

PR #879, opened 2018-10-20
Allow configuration via environment variables

m14t:
**What does this PR do?**

Checks for environment variables of config options, and if present overrides any config file or command-line option with that value.

Environment variable names will be the SNAKE_CASE version of an option's name, prefixed with `SOLID_`.

For example `--db-path` would be overridden by `SOLID_DB_PATH` and  `--serverUri` would be overridden by `SOLID_SERVER_URI`.

Config file values are overridden by CLI flags, which are overridden by Environment variables.

**Links to relevant issues:**

- Implements https://github.com/solid/node-solid-server/pull/869


bin/lib/start.js
```diff
@@ -38,6 +38,25 @@ module.exports = function (program, server) {
...
+
+          return { configName, envName }
+        })
+        .forEach(({ configName, envName }) => {
+          if (process.env[envName]) {
+            try {
+              argv[configName] = JSON.parse(process.env[envName])
+            } catch (e) {
+              argv[configName] = process.env[envName]
+            }
```
RubenVerborgh: This implies that, if a parsing error occurs, we just set the malformed JSON without error. That can't be good.

Let's check if it is JSON indeed (braces?) and only then parse, and not eat errors.

That said, can we avoid JSON at all?

README.md
```diff
@@ -194,9 +195,50 @@ $ solid start --help
     --email-auth-pass [value]     Password of your email service
     --useApiApps                  Do you want to load your default apps on /api/apps?
     --api-apps [value]            Path to the folder to mount on /api/apps
+    --redirect-http-from [value]  HTTP port or ','-separated ports to redirect to the solid server port (e.g. "80,8080").
     -v, --verbose                 Print the logs to console
+    -h, --help                    output usage information
  ```
 
+Instead of using flags, these same options can also be configured via the following environment variables:
```
RubenVerborgh: What takes precedence?

RubenVerborgh: Also, there's now 3 options: config, flags, env. Explain why 3 and why one would prefer any of these in what situations.

bin/lib/start.js
```diff
@@ -38,6 +38,25 @@ module.exports = function (program, server) {
...
 
+      // Read in ENV variables
+      options
+        .map(option => {
+          const configName = option.name.replace(/-([a-z])/g, (_, letter) => letter.toUpperCase())
+          const snakeCaseName = configName.replace(/([A-Z])/g, '_$1')
+          const envName = `SOLID_${snakeCaseName.toUpperCase()}`
+
+          return { configName, envName }
+        })
```
RubenVerborgh: Above can simply be

```JavaScript
      options.map(({ name }) => ({
        configName: name.replace(/-([a-z])/g, (_, l) => l.toUpperCase()),
        envName: `SOLID_${name.replace(/-/g, '_')}`
      }))
```

bin/lib/start.js
m14t: The case is actually a little different.

The `JSON.parse` allows `SOLID_SOME_VAR=true` to get a boolean `true` or `SOLID_PORT=1234` to get a numeric `1234` (and not a string of `"1234"`). It also does allow a user to specify `SOLID_SOME_VAR='{"key": "value"}'` to set an object, which I'm not sure is needed, but the option is available.

However if you try set `SOLID_DB_PATH=./my.db`, the string `
./my.db"` is not valid json, which causes the `JSON.parse` to throw, and so in that case we set this value to the string as it was passed in.

I'd be happy to make any changes, or add any documentation you'd like to see.

RubenVerborgh:
```JavaScript
const raw = process.env[envName]
argv[configName] = /^(d+|true|false)$/.test(raw) ? JSON.parse(raw) : raw
```

RubenVerborgh: But actually, the port should be a string, so

```JavaScript
const raw = process.env[envName]
argv[configName] = /^(true|false)$/.test(raw) ? raw === 'true' : raw
```

bin/lib/start.js
m14t: I believe your example is missing a `. toUpperCase()` on the `envName` line:
```
envName: `SOLID_${name.replace(/-/g, '_').toUpperCase()}`
```

This is a possible solution, though it would map `useCorsProxy` to `SOLID_USECORSPROXY`, where the solution that I submitted would map it to `SOLID_USE_CORS_PROXY` which I thought was better.

Let me know what you prefer.

RubenVerborgh: > I believe your example is missing a `. toUpperCase()`

It was; fixed.

> though it would map useCorsProxy to SOLID_USECORSPROXY, 

It would not; note the `g` after `/-/g`.

m14t: I think there may be an oversight here. [`useCorsProxy`](https://github.com/solid/node-solid-server/blob/master/bin/lib/options.js#L149) does not contain any `-`s in it.

The examples I gave are the output from running the two proposed options.

RubenVerborgh: You're right; I wonder whether that's a mistake in that file then, having both properties with dashes and in camel case. But I guess that in your code, you'd want to change the dashes into underscores as well?

m14t: The way this method is currently written supports the `name`s in `./options` to be in CamelCase or kebab-case, and so I think leaving that as is, might be best for the time being.

image

I agree that having both camel and kebab case `name`s in `./options` is inconsistent, but I would suggest deferring that change for another pull request.

Changing those values will be a breaking change. For example if we were to change `name: 'serverUri',` to `name: 'server-uri',`, anyone currently using the `--serverUri` flag would need to change it to `--server-uri`.

RubenVerborgh: I think we should change them and apply the `configName: name.replace(/-([a-z])/g, (_, l) => l.toUpperCase()),` transformation from above. Or maybe the other way round, having them in camel case in the config (since it is JS) and then changing them to hyphenated.

RubenVerborgh: Ah wait, never mind… it should've been --server-uri, not --serverUri. Sigh.

m14t:
@RubenVerborgh are there any other changes that I can make here?  I've created https://github.com/solid/node-solid-server/pull/885 and have a commit ready to push here with the simplified method if that is merged.

kjetilk:
While awaiting @RubenVerborgh 's feedback, could you please rebase on `develop`, @m14t ?

RubenVerborgh:
Ooh I had approved this, apparently didn't get through. Seeing the timestamp of the commit, looks like that was during last Monday's GitHub outage.

m14t:
Thanks @RubenVerborgh.

Before merging this, one last discussion to have is if the priorities are correct here.

I think that the config file should take the lowest priority, but now that I think about this more, am unsure if environment variables should override the CLI Flag.  Since an env variable is set across multiple runs of a program, and a flag could be changed on every run, I am thinking that it should go (the larger the number wins):
1. config file
2. env variable
3. cli flag

Thoughts?

RubenVerborgh:
Agreed on your order, my thoughts exactly.

kjetilk:
Great stuff! Please rebase on `develop`, I'll merge just after that!

m14t:
@kjetilk I believe this is updated off of `develop` (which currently appears to be the same as `master` and `v4.2.0`)

I just pushed the change to the CLI Flag order. @RubenVerborgh may want to re-review but it was _mostly_ shifting the same code to a different spot.

**Only Config file --> Config file wins**
screen shot 2018-10-24 at 1 57 59 pm

**Config file + ENV Var --> ENV Wins**
screen shot 2018-10-24 at 1 58 41 pm

**Config file + ENV Var + CLI Flag --> CLI Flag Wins**
screen shot 2018-10-24 at 1 58 27 pm

kjetilk:
OK, with #869 now merged, could you please rebase it on `develop`? Then, we'll figured out if there are any conflicts.

m14t:
Rebased off of `97814d4` on `upstream/develop`.

There were no conflicts, but I did update `c0329fb` to include the new flags.

PR #885, opened 2018-10-24
Kebab case CLI Flags

m14t:
**This is a breaking change**

While working on https://github.com/solid/node-solid-server/pull/879 there was discussion about the `name`s used in `./bin/lib/options.js` and the inconsistency between `camelCase` and `kebab-case`.

These variations cause the parsing needed to come up with the proper environment variable name to be slightly more complicated.

**Note:** Once this is merged, anyone using any of these CLI flags will need to start using the new ones.  Config files do not need to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,10 @@ $ solid start --help
     --server-name [value]         A name for your server (not required, but will be presented on your server's frontpage)
     --server-description [value]  A description of your server (not required)
     --server-logo [value]         A logo that represents you, your brand, or your server (not required)
-    -q, --quiet                  Do not print the logs to console
-    -v, --verbose                 DEPRECATED: Print the logs to console.  
+    --enforce-toc                 Do you want to enforce Terms & Conditions for your service?
+    --toc-uri [value]             URI to your Terms & Conditions
+    --support-email [value]       The support email you provide for your users (not required)
+    -q, --quiet                   Do not print the logs to console
     -h, --help                    output usage information
  ```
 
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ $ solid start --help
 
     --root [value]                Root folder to serve (default: './data')
     --port [value]                SSL port to use
-    --serverUri [value]           Solid server uri (default: 'https://localhost:8443')
+    --server-uri [value]          Solid server uri (default: 'https://localhost:8443')
     --webid                       Enable WebID authentication and access control (uses HTTPS)
     --mount [value]               Serve on a specific URL path (default: '/')
     --config-path [value]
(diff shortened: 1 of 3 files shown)


kjetilk:
OK, thanks! Since it is a breaking change, it would have to wait for v5.0.0, so could you please rebase it on the `release/v5.0.0` branch?

melvincarvalho:
I think 5.0.0 is too big a release to have another breaking change.  Let's consider this after 5.0.0.

RubenVerborgh:
From a different perspective: we now have an opportunity window for breaking changes.

melvincarvalho:
@RubenVerborgh indeed.  But this is bike shedding.  It can wait.

kjetilk:
I have no strong opinions, lets see if we want to take it in for 5 when we get there, but it will nevertheless have to be rebased to be considered.

kjetilk:
> @RubenVerborgh indeed. But this is bike shedding. It can wait.

Fair enough, but @m14t is contributing his time, and it doesn't cost us a lot to merge, it costs some users that use CLI something, but they are not many. 

Anyway, how about this: If you are willing to write the patch, @m14t , so that it doesn't break,  but instead indicates deprecation of the old CLI options, it is a step towards removing the old options, it might be more easily digestable. :-) How does that sound?

melvincarvalho:
Much appreciate the contribution.  And it passes all tests, which is awesome!

This is more a question of timing.  5.0.0 is already late, and there is a will to get it out the door.  Once that is done we have more breathing space for other changes and clean up.

We should not be afraid to release 6.0.0 soon after if there are compelling changes.  Working code is more important than numerical labels.  A longer release cycle cadence can happen after we have stability.

RubenVerborgh:
Breakage is fine across major versions, if you just error with a clear message when people use an old option. I wouldn't do compat, too much maintenance. People upgrade to 5, and that needs a change. Fine, if they are warned.

I'm willing to champion this PR; since it's a change at the edge, it will not ripple through.

kjetilk:
OK, I won't argue either way. :-) Lets see when we are ready to work on v5 again.

m14t:
Thank you everyone for the consideration. I defer to your judgment on when the best time/branch to merge this is.

This branch has been rebased off of `release/5.0.0` and the PR target updated.

As this PR is now, if you try to use an old flag, like `--serverUri` you see the following error message:
image

kjetilk:
OK, lets see. If it is just a click of the "Merge pull request" button away from 5.0.0, then it won't delay the release, but if it does delay the release, we can wait. But we are also quite some time from the release, I'll just wait and see before merging.

m14t:
To confirm that the options are indeed arriving further down, I applied the following patch locally:
```diff
diff --git a/bin/lib/start.js b/bin/lib/start.js
index 0762cc7..4d8fbef 100644
--- a/bin/lib/start.js
+++ b/bin/lib/start.js
@@ -44,6 +44,12 @@ module.exports = function (program, server) {
 }
 
 function bin (argv, server) {
+  console.log({
+    serverUri: argv.serverUri,
+    corsProxy: argv.corsProxy,
+    useEmail: argv.useEmail,
+    useApiApps: argv.useApiApps,
+  })
   if (!argv.email) {
     argv.email = {
       host: argv['emailHost'],
```

And then ran:
```bash
npm run solid start -- --server-uri https://somehost:1443 --cors-proxy http://cors.tld --use-api-apps --use-email
```

**Screenshot**
image

RubenVerborgh:
:+1+

kjetilk:
Could you please resolve the conflict, @m14t ? I'll see if I can merge this, 5.0.0 is due out RSN.

m14t:
@kjetilk done!

PR #888, opened 2018-10-27
Add private folder to template

RubenVerborgh:
I watched a new Solid user figure out how to upload stuff to a place no one could see.

Turns out it's hard/unintuitive to create this in the UI:
- creating a child folder of `/` is hard, because there is an HTML page there, not Databrowser;
- when you do create a child folder somewhere else, the "Everyone" icon is listed under Viewers, and I couldn't find how to "drag away" the icon, i.e., not give everyone any permissions.

While this PR doesnm't mitigate the above two issues, at least it provide a private folder by default, which is probably a sensible thing to do.

### DIFF
--- a/default-templates/new-account/index.html
+++ b/default-templates/new-account/index.html
@@ -42,6 +42,10 @@
         <span class="lead">Public Folder</span>
         <span class="badge">public</span>
       </a>
+      <a href="/private/" target="_blank" class="list-group-item">
+        <span class="lead">Private Folder</span>
+        <span class="badge">private</span>
+      </a>
     </div>
   </section>
 
--- a/default-templates/new-account/private/.acl
+++ b/default-templates/new-account/private/.acl
@@ -1,0 +1,10 @@
+# ACL resource for the private folder
+@prefix acl: <http://www.w3.org/ns/auth/acl#>.
+
+# The owner has all permissions
+<#owner>
+    a acl:Authorization;
+    acl:agent <{{webId}}>;
+    acl:accessTo <./>;
+    acl:defaultForNew <./>;
+    acl:mode acl:Read, acl:Write, acl:Control.


default-templates/new-account/index.html
```diff
@@ -42,6 +42,10 @@ <h1>Data</h1>
         <span class="lead">Public Folder</span>
         <span class="badge">public</span>
       </a>
+      <a href="/public/" target="_blank" class="list-group-item">
```
kjetilk: Cut'n'pastefail there?

RubenVerborgh: And this, this is why we do code reviews 😳

default-templates/new-account/index.html
```diff
@@ -42,6 +42,10 @@ <h1>Data</h1>
         <span class="lead">Public Folder</span>
         <span class="badge">public</span>
       </a>
+      <a href="/private/" target="_blank" class="list-group-item">
```
megoth: Is it necessary with `target="_blank"`? It's usually an annoyance for most use cases.

megoth: I don't see a "Good reason" (ref https://css-tricks.com/use-target_blank/) for doing this, but maybe there's something I'm missing?

RubenVerborgh: Copy/paste from what was above.

But a good reason would be that the click brings them to Databrowser, which might be very overwhelming at first. They can just close and be back on the homepage.

megoth: Ok, let's continue that practice for now then

kjetilk:
This fell off my radar... We're go to merge? In which case, it should be on the ASAP board.

megoth:
Yes, this could be merged; want me to rebase it on master?

megoth:
Should be rebased and ready for merging now.

RubenVerborgh:
Seems like a merge, not a rebase. (And a rebase isn't strictly needed, but
can be done.)

megoth:
Gah, my git skills are not the best, especially rebase is something I haven't quite understood yet.

But hope the result looks good nevertheless.

RubenVerborgh:
Don't merge yet, will fix tonight.

Ryuno-Ki:
Regarding `target="_blank"` … came you across https://mathiasbynens.github.io/rel-noopener/ in the past?

RubenVerborgh:
Yes, but this is a local redirect so not an issue.

RubenVerborgh:
@megoth For reference, here are the steps I took:
```bash
git checkout master
git pull
git checkout feature/private-folder # my local copy was still un-rebased
git rebase master
git push -f
```

Note how there is only 1 commit in this PR now.

kjetilk:
OK, great! We've deprecated `develop`, so normal merges will go to `master`, consistent with other repos. And then we have a `released` branch too.

Issue #824, opened 2018-10-02
Cleanup usernames created with invalid characters in community providers


justinwb:
We'll need to clean up usernames that were created with invalid characters (e.g. periods, etc.) on inrupt.net and solid.community after we push the validation fix in https://github.com/solid/node-solid-server/issues/818.

I would propose that we do a quick search, pull the associated e-mail, and shoot them an e-mail letting them know that we need to change their username, or render it inactive unless we hear back from them with an update to it.

mintunitish:
Is this done? If not can I grab this issue?

kjetilk:
@mintunitish Thanks for volunteering, but I think this is something we have to do on the backend, so we would need to do it internally.

megoth:
I realize now that I misread this issue a bit: I was thinking it should include blacklisted usernames as well.

With that said, should this issue include blacklisted usernames as well? (I think it should.)

And in any case; what process do we want to set up for this?

1) Should it be a script that simply removes invalid usernames?
2) Should it communicate to the owner that it is problematic, and given a set time to change it?
2a) What if the owner hasn't given an email; should we add a bit of markup in the index as a notification?

melvincarvalho:
@megoth there's no real way to fix it.  What I think should be communicated is that our systems had an error during account creation.  Certainly not tho only approach, but I think a reasonable one.  Happy for other ideas.

kjetilk:
Yeah, I think we need to give a notice in advance and apologize for an error during account creation....

The blacklist should be involved, as well as the sanitization that we now have in place. We should send an email out first, and then allow a couple of weeks before we remove. So, this is basically the work of a couple of one-off scripts.

megoth:
Ok, so this is the process that I'm considering.

1. Adding command `blacklist` to `bin/solid`: by default it would give you a list of usernames that are in violation of the current blacklist
2. Adding option `--notify`: this will add a note to the top of a user's index.html and send an email, if possible, that their username is in violation of our blacklist and will be deleted within two weeks unless they send an email to the POD provider requesting a new WebID
3. Adding option `--delete`: this will delete the usernames

Does this look ok? I could automate step 3, but think it's better with a manual step for now

Ryuno-Ki:
What about renaming those accounts? Say replacing periods with dashes?
Then you would „just” have to inform them that the name changed.
(I had not looked into implementation details here).

[...]

Issue #880, opened 2018-10-20
Ban certain usernames that could cause confusion or annoyances


RubenVerborgh:
Names like www, admin, and profanity.

elf-pavlik:
Would *profanity* list come from a configuration of each deployment? Also managing i18n list of profanities might come as challenge.

RubenVerborgh:
The entire list, I guess.

megoth:
I guess this issue would also entail removing existing accounts that fall within the ban as well? How should we handle that =/

RubenVerborgh:
Only if they are problematic.

rimmartin:
Another tough thing to parse/recognize is context and meaning.  Such as Mr Coon or Senator Coon

kjetilk:
Hmmmm, but that could could result in a lot of questions about why this and not that? I think we should mainly consider things that could be used in social engineering.

megoth:
Maybe the blacklist could contain the username not allowed and a reason for why? So that if people try to use that name, they would get a reason behind it?

The list would of course have to be maintained by the POD providers, and they would have different reasons for which usernames they allow and disallow. The base file that we provide with the installation/git repo would contain the names that is disallowed due to obvious misunderstandings (such as www and admin, and perhaps password and user?).

steven-tomlinson:
IMO banning common names that could be used for social engineering seems fairly straightforward and reasonable. 

The profanity issue is a whole other can of worms. 
I am inclined to ignore that sort of thing unless I find it personally offensive. In which case, I would probably consider that to be a troll account and block it on the existing social media platforms.

Maybe we need some kind of Username Alias, Proxy, or Masking mechanism? 
It could substitute an undesirable string in a displayed name with asterisks or a completely different name (VulgarPerson12345) based on a POD Providers blacklist?

timbl:
We should ban *www* because Google tends to represent www.foo.bar the same as foo.abr in future, which would be a security issue:  Someone grabing the www username would be able to pose as the whole site, for example probpting for new user names and passwords. Adding security tag,
https://github.com/linkeddata/rdflib.js/blob/master/src/fetcher.js#L643

[...]

PR #893, opened 2018-10-31
A simple service for handling blacklisting of usernames

megoth:
This is depending on the list in [The Big Username Blacklist](https://github.com/marteinn/The-Big-Username-Blacklist). I wrote it as a service to enable testing, but also to make it easy to reuse for https://github.com/solid/node-solid-server/issues/824


megoth:
This is a fix for https://github.com/solid/node-solid-server/issues/880

megoth:
@dmitrizagidulin yeah, that makes sense; Should I also move `lib/models/email-service.js`?

megoth:
> @dmitrizagidulin yeah, that makes sense; Should I also move `lib/models/email-service.js`?

I assumed it was ok, so I moved it as well ^_^ BlacklistService, EmailService, and TokenService now all resides in `lib/services`

dmitrizagidulin:
@megoth perfect, yeah, I forgot about that one.

lib/requests/create-account-request.js
```diff
@@ -115,30 +116,28 @@ class CreateAccountRequest extends AuthRequest {
...
-          debug('Sending Welcome email')
-          accountManager.sendWelcomeEmail(userAccount)
-        }
-      })
-      .then(() => {
-        return userAccount
-      })
+    this.cancelIfUsernameInvalid(userAccount)
+    this.cancelIfBlacklistedUsername(userAccount)
+    await this.cancelIfAccountExists(userAccount)
```
Ryuno-Ki: Is that an [information leak](http://projects.webappsec.org/w/page/13246936/Information%20Leakage)?
That is, do you have a [protection against brute force the usernames](https://www.owasp.org/index.php/Blocking_Brute_Force_Attacks) in place?

lib/requests/create-account-request.js
```diff
@@ -115,30 +116,28 @@ class CreateAccountRequest extends AuthRequest {
...
+    await this.createAccountStorage(userAccount)
+    await this.saveCredentialsFor(userAccount)
+    await this.sendResponse(userAccount)
+
+    // 'return' not used deliberately, no need to block and wait for email
+    if (userAccount && userAccount.email) {
```
Ryuno-Ki: What about looking up the email in the [Have I Been Pwned? database](https://www.npmjs.com/search?q=keywords:haveibeenpwned)?

test/unit/blacklist-service-test.js
```diff
@@ -0,0 +1,49 @@
...
+    it('is a singleton', () => {
+      const instanceA = blacklistService
+      blacklistService.reset({ customBlacklistedUsernames: ['foo'] })
+      expect(instanceA.validate('foo')).to.equal(blacklistService.validate('foo'))
+    })
+  })
+
+  describe('validate', () => {
+    it('validates given a default list of blacklisted usernames', () => {
+      const validWords = blacklist.reduce((memo, word) => memo + (blacklistService.validate(word) ? 1 : 0), 0)
```
Ryuno-Ki: I find this line hard to read. Especially because of the ternary operator and the closing bracket at the end (have to backtrace the matching opening one).

Could be just me, though :-)

lib/requests/create-account-request.js
```diff
@@ -196,12 +195,33 @@ class CreateAccountRequest extends AuthRequest {
...
    *
-   * @return {Promise<UserAccount>} Chainable
+   * @return {UserAccount} Chainable
    */
   cancelIfUsernameInvalid (userAccount) {
     if (!userAccount.username || !/^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(userAccount.username)) {
       debug('Invalid username ' + userAccount.username)
-      const error = new Error('Invalid username')
+      const error = new Error('Invalid username (contains invalid characters)')
+      error.status = 400
```
Ryuno-Ki: Personally, I reach to [http-status](https://www.npmjs.com/package/http-status) to have the speaking code for humans to read.

lib/requests/create-account-request.js
megoth: You're thinking that people could brute force to find out which usernames are taken? But that is already publicly available as it is now... i.e. WebIDs are not publicly listed in an index, but you could brute force through a list of username given a specific WebID-pattern to find out which usernames have WebIDs on a given server as all PODs have public "frontpage" as of now - that might change in the future.

I'm not familiar with how we handle brute force attempts at logging in with a given username though, if we handle it at all right now... (maybe @kjetilk, @RubenVerborgh, or @dmitrizagidulin know?) Definitely something we need to have in place before version 6.0.0 at least, and we should probably create a separate issue for it (I'll wait for response by the people tagged first).

lib/requests/create-account-request.js
megoth: We do this on the front-end side after https://github.com/solid/node-solid-server/pull/859; we could do it on the back-end as well, but I would suggest creating a separate issue for that.

lib/requests/create-account-request.js
dmitrizagidulin: We do not currently handle brute force attacks to find out what usernames are taken. Like you said, it's publicly available information (you can check to see if an account is taken via an API call).

As far as login brute force attacks, `bcrypt` password library provides a decent protection against those (it may make sense to increase our bcrypt params to take slightly longer).

lib/requests/create-account-request.js
dmitrizagidulin: I think that's orthogonal/unrelated to what this PR is trying to do.

lib/requests/create-account-request.js
```diff
@@ -196,12 +195,33 @@ class CreateAccountRequest extends AuthRequest {
...
    *   username.
    *
-   * @return {Promise<UserAccount>} Chainable
+   * @return {UserAccount} Chainable
    */
   cancelIfUsernameInvalid (userAccount) {
     if (!userAccount.username || !/^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(userAccount.username)) {
       debug('Invalid username ' + userAccount.username)
-      const error = new Error('Invalid username')
+      const error = new Error('Invalid username (contains invalid characters)')
```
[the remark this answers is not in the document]
megoth: This is also handled in https://github.com/solid/node-solid-server/pull/859.

Apart from that I agree that we need to make the feedback better. But as we do have some handling for it front-end side I'm leaning toward making a separate issue for this, if at all.

lib/services/blacklist-service.js
```diff
@@ -0,0 +1,33 @@
...
+  addWord (word) {
+    this.list.push(BlacklistService._prepareWord(word))
+  }
+
+  reset (config) {
+    this.list = BlacklistService._initList(config)
+  }
+
+  validate (word) {
+    return this.list.indexOf(BlacklistService._prepareWord(word)) === -1
```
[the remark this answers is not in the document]
megoth: Yeah, that is easier to read, and is supported in Node as well. I don't want to update this PR with it though, as it is already merged, but I'll keep it in mind for later code.

test/unit/blacklist-service-test.js
megoth: You're probably not alone in this, and I'll keep it in mind for later code.

lib/requests/create-account-request.js
megoth: That is a good tip, thanks! I'll try it out in later code iterations.

megoth:
@Ryuno-Ki thanks for a lot of good feedback! As this PR is already merged, I didn't want to do anything more in this PR, but we might create some issues based on your feedback. I'll leave those conversations unresolved for others in the team to revise.

Ryuno-Ki:
Cool! Then let me go ahead and create some tickets :-)

PR #909, opened 2018-11-06
Make --verbose default. Add --silent option.

### LINKED ISSUE #906: Make --verbose flag the default behavior
Too many new users are confused when they launch solid-server and there's no diagnostics. We should make `-v` the default behavior, and deprecate the flag.

ninachaubal:
Makes verbose the default behavior. Adds a `--silent` option to disable logs. Resolves #906

### DIFF
--- a/bin/lib/start.js
+++ b/bin/lib/start.js
@@ -32,7 +32,7 @@ module.exports = function (program, server) {
       }
     })
 
-  start.option('-v, --verbose', 'Print the logs to console')
+  start.option('-s, --silent', 'Do not print the logs to console')
 
   start.action((opts) => {
     let argv = extend({}, opts, { version: program.version() })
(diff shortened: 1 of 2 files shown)


kjetilk:
BTW, could you please rebase on `release/v5.0.0`? Should be easy right now, since we just merged. :-)

ninachaubal:
@dmitrizagidulin done :)

bin/lib/start.js
```diff
@@ -76,7 +76,7 @@ function bin (argv, server) {
   argv.live = !argv.noLive
 
   // Set up debug environment
-  if (argv.verbose) {
+  if (!argv.silent) {
```
kjetilk:
```suggestion
  if (!argv.quiet) {
```

ninachaubal:
@kjetilk, Thanks for catching that! I've pulled your changes in.

PR #911, opened 2018-11-07
Code to check quota on backend 

kjetilk:
So, here's the hack that I managed to pull together. When writing files to the file system, it will check if the user has already exceeded their quota, and if so, responds with a `413`. Other error codes may also be appropriate.

The main problem here is that it breaks existing tests. However, I have tested it in the browser, and there it works, so I have attributed that to a unrepresentative test environment. It is an awful situation to be in, but it seems to take a lot of time to fix these tests (I've spent the whole day on it already). We should discuss what to do about it.


default-templates/new-account/settings/serverSide.ttl
```diff
@@ -0,0 +1,13 @@
+@prefix dct: <http://purl.org/dc/terms/>.
+@prefix pim: <http://www.w3.org/ns/pim/space#>.
+@prefix solid: <http://www.w3.org/ns/solid/terms#>.
+@prefix unit: <http://www.w3.invalid/ns#>.
+
+<>
+  a pim:ConfigurationFile;
+
+  dct:description "Administrative settings for the server that are only readable to the user." .
```
RubenVerborgh: Not the server, the pod?

Also I think with "only readable to the user" you mean "that the user can only read"?

lib/ldp.js
```diff
@@ -10,6 +10,7 @@ const utils = require('./utils')
 const error = require('./http-error')
 const stringToStream = require('./utils').stringToStream
 const serialize = require('./utils').serialize
+const overQuota = require('./utils').overQuota
```
RubenVerborgh: Note for vNext: we should probably be able to implement this with a decorator pattern only.

lib/debug.js
```diff
@@ -13,3 +13,4 @@ exports.container = debug('solid:container')
 exports.accounts = debug('solid:accounts')
 exports.email = debug('solid:email')
 exports.ldp = debug('solid:ldp')
+exports.other = debug('solid:other')
```
RubenVerborgh: `quota` or `fs` perhaps?

lib/utils.js
```diff
@@ -239,3 +245,79 @@ function routeResolvedFile (router, path, file, appendFileName = true) {
...
+  const storageUri = serverUri + '/'
+  try {
+    $rdf.parse(prefs, graph, storageUri, 'text/turtle')
+  } catch (error) {
+    throw new Error('Failed to parse serverSide.ttl, got ' + error)
+  }
+  const lit = graph.each($rdf.sym(storageUri), ns.solid('storageQuota'), undefined)[0]
+  if (lit) {
+    quota = lit.value
+    const unitUri = 'http://www.w3.invalid/ns#'
```
RubenVerborgh: TODO

lib/utils.js
```diff
@@ -239,3 +245,79 @@ function routeResolvedFile (router, path, file, appendFileName = true) {
...
+ */
+
+async function overQuota (root, serverUri) {
+  let quota = await getQuota(root, serverUri)
+  if (quota === Infinity) {
+    return false
+  }
+  // TODO: cache this value?
+  var size = await actualSize(root)
+  console.log(size, quota)
```
RubenVerborgh: TODO: clarify

test/integration/ldp-test.js
```diff
@@ -4,6 +4,7 @@ var ns = require('solid-namespace')($rdf)
 var LDP = require('../../lib/ldp')
 var path = require('path')
 var stringToStream = require('../../lib/utils').stringToStream
+// var randomBytes = require('randombytes')
```
RubenVerborgh: fix

test/integration/ldp-test.js
```diff
@@ -15,6 +16,7 @@ var fs = require('fs')
 describe('LDP', function () {
   var ldp = new LDP({
     root: path.join(__dirname, '..'),
+//    serverUri: 'https://localhost',
```
RubenVerborgh: fix

test/integration/ldp-test.js
```diff
@@ -133,6 +135,23 @@ describe('LDP', function () {
...
+      })
+    })
+    it('should fail if a over quota', function (done) {
+      var hellostream = stringToStream('hello world')
+      ldp.put('localhost', '/resources/testOverQuota.txt', hellostream, function (err) {
+        console.log(err)
+        assert.equal(err.status, 413)
+        done()
+      })
+    }) */
```
RubenVerborgh: fix

test/integration/quota-test.js
```diff
@@ -0,0 +1,46 @@
...
+const getQuota = require('../../lib/utils').getQuota
+const overQuota = require('../../lib/utils').overQuota
+const path = require('path')
+const read = require('../utils').read
+const root = 'accounts-acl/config/templates/new-account/'
+// const $rdf = require('rdflib')
+
+describe('Get Quota', function () {
+  var prefs = read(path.join(root, 'settings/serverSide.ttl'))
+  it('Check that the file is readable and has predicate', function () {
```
RubenVerborgh: typical style would be lowercase and conjugated:
```
it('checks…
```
such that it reads "getQuota checks that the file…".

default-templates/new-account/settings/serverSide.ttl
kjetilk: yeah

lib/debug.js
kjetilk: `fs` is OK

test/integration/quota-test.js
kjetilk: good point.

lib/ldp.js
kjetilk: Sounds good. For example with traits?

lib/utils.js
kjetilk: Since `actualSize` is a heavy operation on large PODs (which the limit makes them not be for now), and it currently happens on every write, it is a DoS opportunity. Caching the value for a bit could alleviate that, but I don't think it is necessary yet. We need to do much better in v.next.

kjetilk:
OK, since we resolved to just do the enforcing as the highest priority, are there any issues that has to be fixed with this? 

Someone[tm] then need to test it manually :-)

justinwb:
Worked well for me. Note that I had to manually copy serverSide.ttl and serverSide.ttl.acl into my pod's settings folder - but did my best to make sure conditions were same as if it was a new install.

- Test that I can upload files under the quota without error - PASS
- Test than I cannot upload any more when my pod is above the stated quota - PASS
- Test that I cannot upload more than my quota ever - FAIL

Based on the way the code is setup I expected this last one to fail, since we are doing the quota check before we start writing the new file, as opposed to "while" we're writing the file. As long as I am under my quota, I can exceed my quota with a file of arbitrary size successfully. After that point I won't be able to do any more, but I could still upload something pretty huge. 

If we were able to count the bytes written as chunks are read and then written to the file stream, i think we would be able to do an inline halt. For example - this is super old and maybe not correct anymore - but highlights what I'm thinking - https://github.com/brycebaril/node-stream-meter

All that said, for me this is working, and it is a big improvement over what we have now - so I'd be okay with going forward as is, but we might want to flag this as something to come back to for stricter enforcement to cover up that last test case.

kjetilk:
Yeah, it would be better to do it while we're writing, so that we could prevent people from writing a huge last file. But I'm thinking about this as a very temporary thing for 5.0.0. It won't scale at all since it also traverses the tree in real time... 

v.next will need an account management feature, but I am hoping we can reuse something.

kjetilk:
I have updated the PR with the simplifications that were discussed in the sprint planning meeting.

I have not changed the filesystem lookup, but I did time the lookup a bit. Launching `get-folder-size` from command line, the JS code is slower than `du`, which determines the size of a ~35MB in about 10-15 ms. The JS code appears to need 50-100 ms, depending on the shape of the tree. How much of this is `nodejs` starting, and how much is the actual tree traverse, I don't know.

What mainly made this painful was that I know of only two places to persist things, in the user's POD and in the oidc-auth-manager's key-value store. Persisting in the POD is nice, but parsing seems to take about the same time as finding the size of the file system, and so, to save time, the quota and the current size would need to be in the same file, or we would need to persist the quotas in RAM. The latter would be the best solution in this case, I think. But I couldn't find an obvious way to do that, but that's just perhaps me not knowing Node well enough. 

I don't know if we should merge this now, or wait for @megoth to find a better solution, but I'm of the opinion that we should rather make use cases like this, delete account, etc, simple to do in v.next rather than trying to save some milliseconds now.

kjetilk:
This is the final issue holding up a 4.4 release.

Should we merge and release 4.4?

kjetilk:
Thank you! Indeed, I think the speedup we could expect to get in the short term would be on the order of some 10s of milliseconds, I think it makes sense to get this in.

PR #912, opened 2018-11-07
Found a bug in how value of toc-checkbox is handled

megoth:
Values on the checkbox might differ as it is now - this should handle it better


Ryuno-Ki:
`!!something` casts `something` to a boolean.
`something === 'true'` compares `something` with the string (!) `true`.
`true === 'true'` yields `false` ;-)

See also https://github.com/denysdovhan/wtfjs#true-is-false

cory2067:
Can confirm this fixes issues I was having with registration on `master`

Issue #916, opened 2018-11-08
Validate email when users create new account


megoth:
The email field in user creation form is optional, but we should validate it if a value is given.

Ryuno-Ki:
Validating against what? Whether it is a valid email? Whether the email is already in use? (hint: possible information leakage) Whether the email is associated with a known spammer?

megoth:
I was only thinking that it should be a valid email.

TallTed:
If this must be done by any means other than simply sending a "confirm your address" message, don't reinvent the wheel.  Use the relevant snippet from the well tested [Email Address Regular Expression That 99.99% Works](https://emailregex.com/).

Ryuno-Ki:
Um, depending on how strict you want to be [that regex could be huge](http://www.ex-parrot.com/~pdw/Mail-RFC822-Address.html) ;-)

megoth:
I was hoping to reuse the `isEmail` in https://www.npmjs.com/package/validator.

dmitrizagidulin:
I would go even farther and point to: https://davidcel.is/posts/stop-validating-email-addresses-with-regex/
(Which says, don't do it, or if you must, just use `/@/ ` as the regular expression)

dmitrizagidulin:
Btw, as far as that `validator` npm package, check out the `options` docstring for that `isEmail()` call. Doesn't that seem like the benefit-to-complexity ration is a bit low?

megoth:
> If this must be done by any means other than simply sending a "confirm your address" message, don't reinvent the wheel. Use the relevant snippet from the well tested [Email Address Regular Expression That 99.99% Works](https://emailregex.com/).

We already have a confirmation email, so we could drop this issue because of that. 

I was thinking of this merely as a helpful addition for those inserting an invalid email.

TallTed:
@Ryuno-Ki -- Note that [your cited regex page](http://www.ex-parrot.com/~pdw/Mail-RFC822-Address.html) is based on [RFC 822](https://tools.ietf.org/html/rfc822), which was obsoleted by [RFC 2822](https://tools.ietf.org/html/rfc2822), and in turn by [RFC 5322](https://tools.ietf.org/html/rfc5322), on which [my cited regex page](https://emailregex.com/) is based.

[...]

PR #919, opened 2018-11-08
Invalidusernames command

megoth:
This is my attempt at fixing https://github.com/solid/node-solid-server/issues/824.

A lot of things happening here, but it boils down to a command with two options:

- `[solid] invalidusernames`: Will list all invalid usernames that exists on the server today
- `[solid] invalidusernames --notify`: Will go through all invalid accounts with invalid usernames and
  1. Make a backup of their existing index.html (move it to index.backup.html)
  2. Create a new index.html that contains info that this account will be deleted in two weeks (based on the new template `default-views/account/invalid-username.hbs`)
  3. Send an email (if available) with the same notification (based on the template new `default-templates/emails/invalid-username.js`)
- `[solid] invalidusernames --delete`: Will delete all accounts with invalid usernames

I've also added a new option in the `init` script that prompts for a support-email for POD-providers (this will only show if the multiuser-property is set to true). As part of this I made use of the `isEmail`-method in https://www.npmjs.com/package/validator, which I understand from https://github.com/solid/node-solid-server/issues/916 might not've been the most well-thought through decision.

I've tried writing some tests, but find it difficult to write tests for a lot of this =/

Any and all feedback are (as always) appreciated, and I hope I'm not too far off with this solution.


RubenVerborgh:
Why is there both a invalid-username.hbs and .js?

megoth:
- `invalid-username.hbs` is for generating the index.html that replaces the (backed up) index.html in a user's POD
- `invalid-username.js` is for generating the email that is sent (if email address is available)

bin/lib/common.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs')
+const extend = require('extend')
+const { cyan, bold } = require('colorette')
+const util = require('util')
+
+module.exports = {}
```
Ryuno-Ki: I could be wrong, but isn't this the default behaviour in CommonJS?

bin/lib/common.js
```diff
@@ -0,0 +1,27 @@
...
+    const file = await util.promisify(fs.readFile)(configFile)
+
+    // Use flags with priority over config file
+    const config = JSON.parse(file)
+    Object.keys(config).forEach((option) => {
+      argv[option] = argv[option] || config[option]
+    })
+  } catch (err) {
+    // No file exists, not a problem
+    console.log(cyan(bold('TIP')), 'create a config.json: `$ solid init`')
```
Ryuno-Ki: Think about using a logger library (likely a new issue). My favourite is [loglevel](https://www.npmjs.com/package/loglevel) (because it's simple and supports plugins).

[winston](https://www.npmjs.com/package/winston) is popular also.

This way you can fine-tune the logging.

bin/lib/common.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = {}
+module.exports.loadConfig = loadConfig
+
+async function loadConfig (program, options) {
+  let argv = extend({}, options, { version: program.version() })
+  let configFile = argv['configFile'] || './config.json'
+
+  try {
+    const file = await util.promisify(fs.readFile)(configFile)
```
Ryuno-Ki: Is that the only occurence of `fs`?
If so, would `const { readFile } = require('fs')` work also?

Ryuno-Ki: Plus `const { promisify } = require('util')`.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+const fileExists = util.promisify(fs.exists)
+const fileRename = util.promisify(fs.rename)
+
+module.exports = function (program) {
+  program
+    .command('invalidusernames')
+    .option('--notify', 'Will notify users with usernames that are invalid')
+    .option('--delete', 'Will delete users with usernames that are invalid')
+    .description('Manage usernames that are invalid')
+    .action(async (options) => {
```
Ryuno-Ki: My personal view: I move the implementation of the action into another file / function.
This way I can have different interfaces for it (CLI, Web UI etc).
Plus, it is easier to test.
Plus, it is a separation of concerns (CLI vs. using the lib).

bin/lib/common.js
dmitrizagidulin: After Node 8, I'm pretty sure all `fs` methods have a Promise interface natively, so the promisify is not even needed.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+      const config = await loadConfig(program, options)
+      if (!config.multiuser) {
+        return console.error('You are running a single user server, no need to check for invalid usernames')
+      }
+
+      const invalidUsernames = await getInvalidUsernames(config)
+      const host = SolidHost.from({ port: config.port, serverUri: config.serverUri })
```
Ryuno-Ki: Hm, I wonder, whether `const host = SolidHost.from({ config })` would work.
Like in [Unpacking fields from objects passed as function parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_fields_from_objects_passed_as_function_parameter)

bin/lib/common.js
dmitrizagidulin: Definitely a new issue, the required changes would be server-wide. 
That said, it /should/ eventually be done, since one of the most frequently requested features about logging is timestamps, which our current logger does not support.

bin/lib/common.js
RubenVerborgh: https://nodejs.org/api/fs.html#fs_fs_promises_api currently has stability: experimental.

bin/lib/common.js
RubenVerborgh: Logging is great for long-running processes (and the server has logging), but this is just a short one-time job.

megoth: Ah, actually, I should probably reuse `debug` that's laid out via `lib/debug.js`.

bin/lib/common.js
megoth: Yeah, you're right, it's unnecessary. I'll remove it.

bin/lib/common.js
megoth: We might consider a logger anyway, but yeah, that should be its own issue.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+
+      if (options.delete) {
+        return deleteUsers(invalidUsernames, accountManager, config, host)
+      }
+
+      listUsernames(listUsernames)
+    })
+}
+
+async function createNewIndexFile (username, accountManager, invalidUsernameTemplate, dateOfRemoval, supportEmail, fileOptions) {
```
Ryuno-Ki: Personally I don't like functions with so many arguments. I consider them a code smell.
Like the function trying to do too much.
It helps to have smaller functions, which are then called in order.

bin/lib/common.js
dmitrizagidulin: > currently has stability: experimental.

Ohhh right, I was thinking about `fs-extra`, never mind :)

megoth: Ok, I'll write away promisify.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+async function deleteUsers (usernames, accountManager, config, host) {
+  const oidcManager = fromServerConfig({
+    ...config,
+    host
+  })
+  const deletingUsers = usernames
+    .map(async username => {
+      try {
+        const user = accountManager.userAccountFrom({ username })
+        await oidcManager.users.deleteUser(user)
```
Ryuno-Ki: IIRC I've read in the docs, that other auth managers (than oidc) would be supported as well.
So I am wondering here, why it has to be oidc.
Maybe add an comment? Could be that it is a „stupid question” for contributors familiar with the code.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+
+      listUsernames(listUsernames)
+    })
+}
+
+async function createNewIndexFile (username, accountManager, invalidUsernameTemplate, dateOfRemoval, supportEmail, fileOptions) {
+  const userDirectory = accountManager.accountDirFor(username)
+  const currentIndex = path.join(userDirectory, 'index.html')
+  const currentIndexExists = await fileExists(currentIndex)
+  const backupIndex = path.join(userDirectory, 'index.backup.html')
```
Ryuno-Ki: This is still in documentRoot, i.e. accessible from the Internet (for everybody?)?

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+  })
+}
+
+async function getInvalidUsernames (config) {
+  const files = await util.promisify(fs.readdir)(config.root)
+  const hostname = new URL(config.serverUri).hostname
+  const isUserDirectory = new RegExp(`.${hostname}$`)
+  return files
+    .filter(file => isUserDirectory.test(file))
+    .map(userDirectory => userDirectory.substr(0, userDirectory.length - hostname.length - 1))
```
Ryuno-Ki: Why exactly `userDirectory.length - hostname.length - 1`?

bin/lib/invalidUsernames.js
megoth: Oh, should probably copy the acl-file as well O_O

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+    .filter(username => !isValidUsername(username) || !blacklistService.validate(username))
+}
+
+function listUsernames (usernames) {
+  if (usernames.length === 0) {
+    console.info('No invalid usernames was found')
+  }
+  console.info(`${usernames.length} invalid usernames were found:${usernames.map(username => `\n- ${username}`)}`)
```
Ryuno-Ki: Shouldn't that go into an `else` branch?

bin/lib/invalidUsernames.js
megoth: Userdirectory can be read as `[username].[hostname]`, so it's to remove the dot between username and hostname.

bin/lib/invalidUsernames.js
megoth: Oh, forgot a return in the if-block (that's at least how I like to write that logic).

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+}
+
+async function notifyUsers (usernames, accountManager, config) {
+  const twoWeeksFromNow = Date.now() + 14 * 24 * 60 * 60 * 1000
```
Ryuno-Ki: If you ever struggle with datetimes, look at [date-fns](https://www.npmjs.com/package/date-fns) or [moment](https://www.npmjs.com/package/moment).

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+    const configPath = initConfigDir(config)
+    const templates = initTemplateDirs(configPath)
+    const users = await Promise.all(await usernames.map(async username => {
+      const emailAddress = await accountManager.loadAccountRecoveryEmail({ username })
+      const accountUri = accountManager.accountUriFor(username)
+      return { username, emailAddress, accountUri }
+    }))
+    const emailService = new EmailService(templates.email, config.email)
+    const sendingEmails = await users
+      .filter(user => !!user.emailAddress)
```
Ryuno-Ki: Maybe more readable as `.filter(user => Boolean(user.emailAddress))`?

bin/lib/invalidUsernames.js
megoth: I don't like long signatures either, but in this case it's not really reusable functions, but I wanted more readable code by splitting it up into "logical" chunks of code (in what I hope is good named functions).

But yeah, if we wanted to reuse this functionality, I should probably restructure it a bit to decrease the length of the signatures.

bin/lib/invalidUsernames.js
```diff
@@ -0,0 +1,160 @@
...
+    const emailsSent = await Promise.all(sendingEmails)
+    console.info(`${emailsSent.length} emails sent to users with invalid usernames`)
+    return
+  }
+  console.info('You have not configured an email service.')
+  console.info('Please set it up to send users email about their accounts')
+}
+
+async function updateIndexFiles (usernames, accountManager, dateOfRemoval, supportEmail) {
+  const invalidUsernameFilePath = path.join(process.cwd(), 'default-views/account/invalid-username.hbs')
```
Ryuno-Ki: Why do you use `path.join` when the second argument is not OS-agnostic?

bin/lib/invalidUsernames.js
dmitrizagidulin: I disagree, `!!` is more standard, and is used throughout our libs.

default-templates/emails/invalid-username.js
```diff
@@ -0,0 +1,30 @@
...
+function render (data) {
+  return {
+    subject: `Invalid username for account ${data.accountUri}`,
+
+    /**
+     * Text version
+     */
+    text: `Hi,
+
+We're sorry to inform you that the username for account ${data.accountUri} is not allowed after changes to username policy.
```
Ryuno-Ki: Maybe refer to `username policy` somehow? I'd be wondering, what exactly is in there if I would receive such an e-mail.

default-views/account/invalid-username.hbs
```diff
@@ -0,0 +1,23 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Invalid username</title>
+  <link rel="stylesheet" href="/common/css/bootstrap.min.css">
+  <script></script>
```
Ryuno-Ki: Huh? An empty tag?

bin/lib/invalidUsernames.js
megoth: I would do it 100% if I thought this code was reusable, which I'm leaning toward it isn't =/

As I mentioned earlier, I found it difficult to write tests for this. It relies on a lot of file-based functionality, which makes it unsuited for unit tests. I could write integration tests, but didn't see any for the other commands, so didn't know how to attack it.

default-views/account/invalid-username.hbs
```diff
@@ -0,0 +1,23 @@
...
+</head>
+<body>
+<div class="container">
+  <h4>Invalid username</h4>
```
Ryuno-Ki: Why not an `h1`? Semantic Web and so on.

bin/lib/invalidUsernames.js
megoth: I reckon you mean `const host = SolidHost.from(config)`? (As `const host = SolidHost.from({ config })` is a shortcut for `const host = SolidHost.from({ config: config })`.)

Could work, but I wanted to stay specific reg this.

dmitrizagidulin: Agreed, more specific is better in this case (since it's clearer which args are actually used)

bin/lib/common.js
Ryuno-Ki: https://github.com/solid/node-solid-server/issues/920

bin/lib/invalidUsernames.js
Ryuno-Ki: From an Angular point of view I'd argue with Dependency Injection.
That is, providing the `fs` as an argument to the function.
But it feels meh somehow.

bin/lib/invalidUsernames.js
megoth: No, it's not a stupid question; We support WebID-TLS and WebID-OIDC now, and WebID-TLS don't store user-data like WebID-OIDC does.

I could try to abstract away the OIDC-part of this, to accommodate future auth managers better, but I think that's going to might take a long time (for me, at least - I'm not that into auth managers yet). So I was thinking this can be done when we implement support for other auth managers.

bin/lib/invalidUsernames.js
Ryuno-Ki: `!!` reminds me on the output of `uglify` :-)
But being used throughout is a valid argument.

bin/lib/invalidUsernames.js
Ryuno-Ki: Fair enough.

bin/lib/invalidUsernames.js
megoth: Was planning to see if I could find another a way to reference `default-views/account`, but forgot about it. I'll make another try, or at least fix this to be OS-agnostic.

bin/lib/invalidUsernames.js
Ryuno-Ki: Well, if you are struggling with writing tests, this should be argument enough to break the function up, shouldn't it?

default-templates/emails/invalid-username.js
megoth: I don't think this is necessary, but open to be convinced otherwise.

default-views/account/invalid-username.hbs
megoth: Will remove.

bin/lib/invalidUsernames.js
Ryuno-Ki: Maybe you could leave a comment here, so the next person touching the code knows your thoughts (since I'd not look up the associated PR immediately, but rather `git blame` …).

I don't know the conventions in this project, though.

Ryuno-Ki:
I left some comments while walking through the files.

default-views/account/invalid-username.hbs
megoth: Because of legacy; the other views use h4 (I reckon they did it like this initially because they thought h1 was to big; We (mostly) use bootstrap (3.3.4) for the presentation).

dmitrizagidulin: @megoth hehe, that's exactly why, h1 was too big on bootstrap :)

Ryuno-Ki: https://getbootstrap.com/docs/3.3/customize/#typography *cough

bin/lib/invalidUsernames.js
megoth: I'm not that familiar with dependency injection in Node. I've tried [MockRequire](https://www.npmjs.com/package/mockrequire) earlier, but my tests always turned a nightmare to maintain.

bin/lib/invalidUsernames.js
megoth: If I knew how to break up this code into reusable, testable functions, I would do it ;)

bin/lib/invalidUsernames.js
megoth: Yeah, I prefer the former, but didn't want to include it just for this.

bin/lib/invalidUsernames.js
Ryuno-Ki: Okay, I can write the high-level steps I see. Implementation goes into the functions then …

```js
async function createNewIndexFile (/* arguments */) {
  if (isBackupNeeded(/* arguments */)) {
    await Promise.all([
      backupIndexHtml(/* arguments */)
      backupAcl(/* arguments */)
    ])
    await Promise.all([
      generateNewIndexHtml(/* arguments */)
      generateNewAcl(/* arguments */)
    ])
    logCompletion(/* arguments */)
  }
}
```
or maybe group the backup + creation into a single function (like an transaction).
You may want to timestamp the `index.backup.html` also.

Ryuno-Ki: Plus, looking at the function name, `createNewIndexFile` is misleading, since it does more than it says …
so its caller would need to call more functions.

bin/lib/invalidUsernames.js
Ryuno-Ki: Is it fixed that there must be a dot?
If so, can't you `[split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)` the `hostname` (and take the second-to-last value)?

kjetilk:
Great with a lot of comments, I just wanted to note that this code will most likely run only once, so it is not worth spending too much time on it.

PR #928, opened 2018-11-12
A simple node-script equivalent for solid-test

megoth:
I'm uncertain how smart setting a variable for process.env is, but this works for me. Hopefully someone with more Node experience than me can verify if it's smart or not.

(I use [nodemon](https://www.npmjs.com/package/nodemon) when developing, to enable a quick restart when I change code, so I find this script useful)

### DIFF
--- a/bin/solid-test.js
+++ b/bin/solid-test.js
@@ -1,0 +1,6 @@
+#!/usr/bin/env node
+
+process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = 0
+
+const startCli = require('./lib/cli')
+startCli()


rubensworks:
Could it be possible that nodemon passes environment variables down to the restarted processes? In that case setting the variable once might be sufficient?

I would personally not include this in a repo, but I don't have a strong opinion. In any case, I would add it do the `.npmignore` file if this is merged.

Ryuno-Ki:
I put those things in a `.env` file (which is gitignored!) and load it using the `dotenv` package.

kjetilk:
I don't think I have more Node experience than you :-)

kjetilk:
Seems like people have a problem on Windows, so I suppose this is good to have as platform independent.

RubenVerborgh:
#802

megoth:
I've created https://github.com/solid/node-solid-server/issues/954 that should get a proper fix, not this incomplete version.

Issue #882, opened 2018-10-23
Validation of LD-files


megoth:
When posting triples as TTL, RDF+XML, JSON-LD, etc, I think we should do validation on the server. I don't think this should be to hard to implement, but the devils probably in the detail, so want to have a good discussion on this.

Also, I'm uncertain if this should be part of 5.0.0. or v.next. Any thoughts?

kjetilk:
I think it is clearly a part of the v.next architecture, but it seems to touch all the handlers in the current architecture.

akuckartz:
What happens when there is no validation?

kjetilk:
The server would respond with a `400`. It is no replacement for client-side validation, but it is a last stand against breaking stuff.

megoth:
I also think this would limit some of the issues that people have experienced (https://github.com/solid/solid/issues/165, https://github.com/solid/solid/issues/206, https://github.com/solid/node-solid-server/issues/826)

megoth:
I think https://github.com/solid/node-solid-server/issues/826 also relates to this issue.

megoth:
I'm discontinuing my work on this until we have a broader consensus on validation in general (currently discussed in [Validation and filter feature and Discourse](https://forum.solidproject.org/t/validation-and-filter-feature/548)).

I've pushed my work on https://github.com/solid/node-solid-server/tree/feature/simple-validation, in case it's interesting to continue at some point.

megoth:
After a discussion with @timbl, @kjetilk, and @rubensworks on this earlier it seems that validation on the ACL-files might be a good compromise. If noone objects to this, I'll continue working on that tomorrow (it's should be almost done in that regard).

megoth:
This issue has been fixed, we implemented syntax validation for ACL files in https://github.com/solid/node-solid-server/pull/937

PR #937, opened 2018-11-15
Simple RDF validation for ACL files

megoth:
This is PR based on the discussions at https://github.com/solid/node-solid-server/issues/882.

Simply explained, when putting ACL files it will parse the files to make the syntax is valid.

I also extracted a couple of utility-functions for integration-tests that I think make the tests a little bit easier to read.


lib/ldp.js
```diff
@@ -256,25 +256,53 @@ class LDP {
...
+  _put (host, resourcePath, callback, resourceHandler) {
     const ldp = this
     const root = !ldp.multiuser ? ldp.root : ldp.root + host + '/'
     const filePath = utils.uriToFilename(resourcePath, root, host)
 
     // PUT requests not supported on containers. Use POST instead
     if (filePath.endsWith('/')) {
-      return callback(error(409,
-        'PUT not supported on containers, use POST instead'))
+      return callback(error(409, 'PUT not supported on containers, use POST instead'))
```
Ryuno-Ki: Wouldn't be a [405 - Method not allowed](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors) be more appropriate then?

lib/ldp.js
```diff
@@ -538,6 +566,10 @@ class LDP {
     }
     return ensureNotExists(this, path.join(containerURI, filename))
   }
+
+  mimetypeIsRdf (mimeType) {
+    return RDF_MIME_TYPES.indexOf(mimeType) !== -1
```
Ryuno-Ki: Assuming `RDF_MIME_TYPES` is an Array you can use [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) I guess.

test/integration/cors-proxy-test.js
```diff
@@ -1,18 +1,14 @@
...
 var nock = require('nock')
-var { checkDnsSettings } = require('../utils')
-
-var ldnode = require('../../index')
+var { checkDnsSettings, setupSupertestServer } = require('../utils')
 
 describe('CORS Proxy', () => {
-  var ldp = ldnode({
+  var server = setupSupertestServer({
     root: path.join(__dirname, '../resources'),
```
Ryuno-Ki: Well, if you want to go cross-platform, `path.join(__dirname, '..', 'resources')` would make Windows users happy.

test/integration/cors-proxy-test.js
```diff
@@ -1,18 +1,14 @@
 var assert = require('chai').assert
-var supertest = require('supertest')
 var path = require('path')
 var nock = require('nock')
-var { checkDnsSettings } = require('../utils')
-
-var ldnode = require('../../index')
+var { checkDnsSettings, setupSupertestServer } = require('../utils')
```
Ryuno-Ki: Why a `var` instead of a `const`?

lib/ldp.js
RubenVerborgh: Definitely 405.

lib/ldp.js
RubenVerborgh: Better, make RDF_MIME_TYPES a Set.

test/integration/cors-proxy-test.js
RubenVerborgh: Consistency.

lib/ldp.js
```diff
@@ -256,25 +256,53 @@ class LDP {
...
+      resourceHandler(filePath)
+    })
+  }
+
+  putText (host, resourcePath, text, callback) {
+    this._put(host, resourcePath, callback, (filePath) => {
+      try {
+        fs.writeFileSync(filePath, text)
+      } catch (err) {
+        return callback(error(500, err.message))
```
RubenVerborgh: You might want to delete whatever was written do disk.

lib/ldp.js
```diff
@@ -256,25 +256,53 @@ class LDP {
...
+
       // Directory created, now write the file
+      resourceHandler(filePath)
+    })
+  }
+
+  putText (host, resourcePath, text, callback) {
+    this._put(host, resourcePath, callback, (filePath) => {
+      try {
+        fs.writeFileSync(filePath, text)
```
RubenVerborgh: Async?

lib/ldp.js
megoth: > Assuming `RDF_MIME_TYPES` is an Array you can use [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) I guess.

I actually did this initially when I wrote the code, but it failed for me, and I automatically reverted to Array.prototype.indexOf without much more research. My bad.

But yeah, will try making it a Set.

test/integration/cors-proxy-test.js
megoth: > Well, if you want to go cross-platform, `path.join(__dirname, '..', 'resources')` would make Windows users happy.

Is it necessary though? It seems like [path.join will normalize it nevertheless](https://shapeshed.com/writing-cross-platform-node/#use-path-join-to-join-folder-names).

lib/ldp.js
megoth: I'm wondering about the difference  of `fs.writeFileSync` and `await fs.writeFile` in general. It seems that there's [no difference in blocking IO](https://www.daveeddy.com/2013/03/26/synchronous-file-io-in-nodejs/), and it seems that it doesn't affect [the main thread as filesystem operations are handled in worker threads](http://disq.us/p/16tx816).

It doesn't seem to be that much of a difference, so in this case I think I prefer `fs.writeFileSync`. But at some point we might want to have a convention on this, so that we stay consistent throughout the code.

megoth:
@Ryuno-Ki @RubenVerborgh I've pushed a commit with some changes based on your feedback. I've left out a couple of them, which I've commented.

lib/ldp.js
megoth: What if the file existed from before, and we need to recover that file?

If it exists from before, maybe I should do a backup of the content, and roll back if the write failed? Otherwise, I should delete whatever was written to disk?

test/integration/cors-proxy-test.js
Ryuno-Ki: Hm, you made me look at the [source code of `path.join`](https://github.com/nodejs/node/blob/v11.2.0/lib/path.js#L488).

The call to [`win32.normalize`](https://github.com/nodejs/node/blob/v11.2.0/lib/path.js#L284) does some … things. So it could be possible that it would normalise it on its own.

test/integration/validate-tts.js
```diff
@@ -0,0 +1,51 @@
+const fs = require('fs')
+const path = require('path')
+const { setupSupertestServer } = require('../utils')
+
+const server = setupSupertestServer({
+  live: true,
+  dataBrowserPath: 'default',
+  root: path.join(__dirname, '../resources'),
```
Ryuno-Ki: I'd have split it to `path.join(__dirname, '..', 'resources')`, but if it works everywhere …

Ryuno-Ki:
I consider the `405` comment necessary to deal with, but I don't have any roles here ;-)

megoth:
> I consider the `405` comment necessary to deal with, but I don't have any roles here ;-)

But it has been fixed now, hasn't it? Or am I missing something.

lib/ldp.js
```diff
@@ -256,25 +256,53 @@ class LDP {
...
+  _put (host, resourcePath, callback, resourceHandler) {
     const ldp = this
     const root = !ldp.multiuser ? ldp.root : ldp.root + host + '/'
     const filePath = utils.uriToFilename(resourcePath, root, host)
 
     // PUT requests not supported on containers. Use POST instead
     if (filePath.endsWith('/')) {
-      return callback(error(409,
-        'PUT not supported on containers, use POST instead'))
+      return callback(error(405, 'PUT not supported on containers, use POST instead'))
```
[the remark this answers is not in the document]
megoth: Maybe @RubenVerborgh or @kjetilk can answer this?

megoth:
I think I've handled all feedback that's been given on this. Anyone up for approving this PR? @kjetilk @RubenVerborgh (or maybe @rubensworks)?

rubensworks:
Same rebasing issue here as in #953.

Also, should this be targeted to the release branch?

megoth:
I've rebased this branch on `release/v5.0.0` and think it's ready to be merged. Maybe @rubensworks, @RubenVerborgh or @kjetilk wants to give it a last review to verify that it looks good?

rubensworks:
Ok, I'll have a look.

PR #959, opened 2018-11-21
Wire up ResourceMapper

rubensworks:
As requested in #662, this wires up the new ResourceMapper, as the next step for the big ResourceMapper refactor (#946).

A bunch of existing mapping-related issues popped up when working on this PR (see commits), and have also been fixed.

I suggest to not review this before #952 is merged, as this is forked from that branch, and should be rebased to master first. (some small changes are still needed there, will look at those first thing tomorrow)

Issue #946, opened 2018-11-19
ResourceMapper transition actions


RubenVerborgh:
### Context
We are in the process of changing the URL-to-filename correspondence within node-solid-server, such that it conforms to https://www.w3.org/DesignIssues/HTTPFilenameMapping.html. Importantly, this fixes the current Content-Type breakage we have (#925 and others).
As an example:
- v4.0.0 will map `file:///solid/profile/card` to `https://pod.example/profile/card` with content-type `text/turtle`
- in v5.0.0, the file on disk will be `file:///solid/profile/card$.ttl` instead

This transition is quite complex, since the URL-to-filename correspondence is implemented in multiple places instead of just one. Furthermore, this is a semver.major change.

This issue tracks all steps that need to be taken.

### Steps
1. [x] Encapsulate the current filesystem mapping in one component (#660)
1. [x] Replace all ad-hoc mapping code with the single mapping component (#661)
1. [x] Evaluate whether everything still works with the single component, fix where needed (#947)
1. [x] Encapsulate the new filesystem mapping in one component (#643)
1. [x] Replace all mentions of the old mapping component with the new mapping component (#662)
1. [x] Update the new data pod template filenames to use the new mapping (#948)
1. [x] Write a tool to migrate old Solid filesystems to new Solid filesystems (#949)
1. [ ] Upgrade the software in all of our services and migrate the filesystems (#950)
1. [x] Remove LegacyResourceMapper and tests
1. [x] Document upgrade notes (#951)

melvincarvalho:
Great work!

A couple of things.

- Would it be possible to make this a switch, or behind a flag, or config option.  That will enable smoother upgrade, easier testing, and more flexibility for pod owner.

- It seems like quite a big change, especially in terms of technical debt.  ie requires file name changes.  Could I request that we put this functionally live in a version after 5.0.0, so that it possible to get the stability release and quota work, before requiring the file system changes.

Perhaps this is a good topic for a new issue, or the community group.  But I have often wondered why the header information (Content-Type) could not be put in the .meta file.  That would seem to be a more portable solution (across, different versions, different solid language servers, different free software cloud solutions and even different web servers).

kjetilk:
>    Would it be possible to make this a switch, or behind a flag, or config option. That will enable smoother upgrade, easier testing, and more flexibility for pod owners.

I don't think we should do that. We need to move forward, and such a switch would require even more complexity. Such flexibility comes at a significant cost. 

[...]

PR #960, opened 2018-11-22
Wire up ResourceMapper

rubensworks:
(This replaces #959 because I started from a new branch with manual cherry-picking from the old one, instead of doing a highly conflicting rebase)

As requested in #662, this wires up the new ResourceMapper, as the next step for the big ResourceMapper refactor (#946).

A bunch of existing mapping-related issues popped up when working on this PR (see commits), and have also been fixed.


config/defaults.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   'strictOrigin': true,
   'originsAllowed': ['https://apps.solid.invalid'],
   'dataBrowserPath': 'default',
-  'defaultContentType': 'text/turtle'
+  'defaultContentType': 'application/octet-stream'
```
RubenVerborgh: Is that setting used? Probably should be up to the resource mapper used.

rubensworks: Yes, it's being forwarded to the resource mapper. (Because we removed all defaults there in the last PR)

lib/handlers/get.js
```diff
@@ -136,7 +135,9 @@ async function handler (req, res, next) {
 
 async function globHandler (req, res, next) {
   const ldp = req.app.locals.ldp
-  const { path: filename } = await ldp.resourceMapper.mapUrlToFile({ url: req })
+  // TODO: This is a hack, as globbing and resource mapping in combination is complex
+  // TODO: Proper support for this is not implemented, as globbing support will probably be removed in the future.
```
RubenVerborgh: Perhaps explain what the hack does and how it works?

Nit: globbing support _might_ be removed at some point in the future.

lib/handlers/patch.js
```diff
@@ -16,14 +16,24 @@ const PATCH_PARSERS = {
   'text/n3': require('./patch/n3-patch-parser.js')
 }
 
+const DEFAULT_PATCH_CONTENT_TYPE = 'text/turtle'
```
RubenVerborgh: Naming: it's not the patch content type. Maybe default for new?

lib/handlers/put.js
```diff
@@ -8,7 +8,7 @@ async function handler (req, res, next) {
   res.header('MS-Author-Via', 'SPARQL')
 
   try {
-    await ldp.put(req, req)
+    await ldp.put(req, req, req.headers['content-type'])
```
RubenVerborgh: There might be an encoding attached to the Content-Type header. Problem?

lib/header.js
```diff
@@ -43,7 +43,19 @@ function addLinks (res, fileMetadata) {
 
 async function linksHandler (req, res, next) {
   const ldp = req.app.locals.ldp
-  const { path: filename } = await ldp.resourceMapper.mapUrlToFile(req)
+  let filename
+  try {
+    // Hack: createIfNotExists is set to true for PUT or PATCH requests
+    // because the file does not exist yet at this point.
```
RubenVerborgh: _might_ not exist

lib/ldp.js
```diff
@@ -18,7 +16,7 @@ const parse = require('./utils').parse
 const fetch = require('node-fetch')
 const { promisify } = require('util')
 
-const DEFAULT_CONTENT_TYPE = 'text/turtle'
+const DEFAULT_REMOTE_CONTENT_TYPE = 'text/turtle'
```
RubenVerborgh: Remote?

lib/resource-mapper.js
```diff
@@ -8,12 +8,21 @@ const readdir = promisify(fs.readdir)
 // following the principles of the “sweet spot” discussed in
 // https://www.w3.org/DesignIssues/HTTPFilenameMapping.html
 class ResourceMapper {
-  constructor ({ rootUrl, rootPath, includeHost, defaultContentType }) {
+  constructor ({
+    rootUrl,
+    rootPath,
+    includeHost,
+    defaultContentType,
+    indexName = 'index',
```
RubenVerborgh: naming: `indexFilename`?

lib/resource-mapper.js
```diff
@@ -29,16 +38,28 @@ class ResourceMapper {
...
+      fullPath += this._indexName
+    }
+
     // Create the path for a new file
     if (createIfNotExists) {
       path = fullPath
       // If the extension is not correct for the content type, append the correct extension
       if (this._getContentTypeByExtension(path) !== contentType) {
-        path += contentType in extensions ? `$.${extensions[contentType][0]}` : '$.unknown'
+        // Append a '$', unless we map for the index
```
RubenVerborgh: But why? (Document.)

lib/resource-mapper.js
```diff
@@ -29,16 +38,28 @@ class ResourceMapper {
...
+        if (!isIndex) {
+          path += '$'
+        }
```
RubenVerborgh: I don't get this change I'm afraid. Are there situations where we end up with multiple dollar signs now?

lib/resource-mapper.js
```diff
@@ -48,9 +69,17 @@ class ResourceMapper {
       const files = await this._readdir(folder)
 
       // Find a file with the same name (minus the dollar extension)
-      const match = files.find(f => this._removeDollarExtension(f) === filename)
+      let match = files.find(f => this._removeDollarExtension(f) === filename ||
+        (isIndex && f.startsWith(this._indexName + '.')))
```
RubenVerborgh: Just wondering here if the first case and the `isIndex` case are mutually exclusive. If they are, we might want to write this differently.

lib/resource-mapper.js
```diff
@@ -48,9 +69,17 @@ class ResourceMapper {
...
       if (!match) {
-        throw new Error('File not found')
+        // Error if no match was found,
+        // unless the URL ends with a '/',
+        // in that case we fallback to the folder itself.
+        if (isIndex) {
```
RubenVerborgh: Given the above comment, we might want fully separate `isIndex` code paths.

lib/resource-mapper.js
```diff
@@ -8,12 +8,21 @@ const readdir = promisify(fs.readdir)
 // following the principles of the “sweet spot” discussed in
 // https://www.w3.org/DesignIssues/HTTPFilenameMapping.html
 class ResourceMapper {
-  constructor ({ rootUrl, rootPath, includeHost, defaultContentType }) {
+  constructor ({
+    rootUrl,
+    rootPath,
+    includeHost,
+    defaultContentType,
```
RubenVerborgh: This one can be assigned to `'application/octet-stream'` by default; it's the most generic content type.

test/integration/account-manager-test.js
```diff
@@ -30,11 +30,11 @@ describe('AccountManager', () => {
 
     describe('in multi user mode', () => {
       let multiuser = true
-      let resourceMapper = new LegacyResourceMapper({
+      let resourceMapper = new ResourceMapper({
         rootUrl: 'https://localhost:8443/',
         rootPath: process.cwd(),
         includeHost: multiuser,
-        defaultContentType: 'text/turtle'
+        defaultContentType: 'application/octet-stream'
```
RubenVerborgh: Given comment above somewhere, this can go then.

config/defaults.js
RubenVerborgh: Ah yeah, my bad. I was only talking about the very specific defaults, such as the localhost and current folder one. octet-stream is the most basic MIME type, so that seems like a sensible default.

lib/resource-mapper.js
rubensworks: > Are there situations where we end up with multiple dollar signs now?

No, at most 1 dollar sign will be present.

> I don't get this change I'm afraid.

This is needed for this situation:
PUT on `http://localhost/space/` with content-type `text/html` should create `/space/index.html`, instead of `/space/index$.html`.
(Unless we don't want this?...)

rubensworks:
Comments are resolved :-)

PR #964, opened 2018-11-23
Add pre-commit hook to trigger linter

rubensworks:
Every time a commit is about to be done, `npm run standard` will be invoked.
This makes standard-compliance a bit more convenient :-)

Even though this would be possible, the unit tests aren't executed during pre-commit because they take a bit too long to execute.


kjetilk:
Yeah, I like the idea, but I more like the idea of a pre-push hook to run `npm test`, with a question at the end asking whether you'd like to push anyway. Lets talk about it post-release.

kjetilk:
Now, we can reconsider this. Can you have a look, @jaxoncreed ?

PR #965, opened 2018-11-23
Cache node_modules in Travis

rubensworks:
This makes Travis builds slightly faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,10 @@ script:
 after_success:
   - snyk monitor
 
+cache:
+  directories:
+    - node_modules
+
 notifications:
   email:
   - solid@janeirodigital.com


rubensworks:
Ah, TIL `npm ci` deletes `node_modules`.

In any case, it looks like `npm install` without changes dependencies takes about 7,5 seconds, while `npm ci` takes 30 seconds.

So `npm install` is definitely faster if `node_modules` can be cached. So do we want to transition to that instead? (`npm ci` is of course better in the sense that it starts from a clean environment)

RubenVerborgh:
I strongly suggest `npm ci` for consistency reasons. It's slower, but doesn't make mistakes.

rubensworks:
Sure, I can live with that :-)

In any case, installation is not the main bottleneck in Travis builds.

RubenVerborgh:
@rubensworks Apparently we can cache npm though: https://docs.travis-ci.com/user/caching/#npm-cache

rubensworks:
Apparently GitHub doesn't allow you to re-open PRs after force-pushing to the branch. Created #971 instead.

Issue #787, opened 2018-09-19
Test for careful handling of all errors loading ACL files Group files, etc


timbl:
Checking the ACL files involves zooming up the tree looking for the first one which exists.  It is tempting for the developer to just zoom up they get a 200 good load the ACL file, passing over everything that fails, 404 or not.  This is a very bad idea!  It must be explicitly tested for.

*The ACL file must only be skipped over if and only  if generated a only a 404 (not found) error*

Failure scenarios include:

- The net being down for a second, giving an error, when in fact the file exists. (with net access)

- The process has some internal fleeting problem for a second, giving an error, when in fact the file exists. 

- The ACL file has a parse error, and then  is treated as though it didn't exist.  

- The ACL file is not authorized at the unix file permissions level.

and so on.  As the ACL file in question may be more constraining than the default one higher in the tree, and so these error would end up giving access to someone who would not have been allowed it.

An ACL file parse error must generate a 500 response.

(Some but not all of these considerations apply to Group membership resources.)

timbl:
When I say "generates a 404" I don't mean you have to do an HTTP request.  You do hae to chek that the file does not exist (as opposed to anu other error like no such file system or parse error etc

RubenVerborgh:
We have had a privilege escalation bug in the past, where an ACL file could not be read (due to a server bug), and it was happily skipped.

So I think the issue here is: if an ACL file exists on disk, never ever read the parent.

kjetilk:
Right, good point.

Ryuno-Ki:
> https://github.com/solid/node-solid-server/blob/master/test/integration/acl-oidc-test.js#L176

As a side note: If you hit <kbd>y</kbd> while on that page, GitHub will expand the URL to a canonical form, which does not change over time. This way, you can look up what was referred to in a few month from now (when the page likely changed). Hit <kbd>?</kbd> to learn about more GitHub shortcuts :-)

kjetilk:
> Neither; it's a 401/403.
> (Reasoning: it is a valid Turtle file, and specifies no access rights for the current user.)

Come to think of it, when codifying this in tests: Wouldn't an empty .acl for a file lead to loss of control (i.e., you cannot again change the controls) for that file?

I haven't got time to look into this now, but we should have some strategies to mitigate the effects of stuff like that happening by mistake, error or malice.

dmitrizagidulin:
@kjetilk Yes, writing an empty .acl does lead to a loss of control for that resource or container.

[...]

Issue #785, opened 2018-09-17
Infinite loop caused by intentionally invalid test data


kjetilk:
To test the server's resilience against errors, I made tests in https://github.com/solid/node-solid-server/blob/master/test/integration/acl-oidc-test.js#L581 that should do the Right Thing even if an `Authorization` points at an invalid group listing, like it does in https://github.com/solid/node-solid-server/blob/master/test/resources/accounts-acl/tim.localhost/group/test-folder/group-listing-error.ttl

It turns out that the presence of this `Authorization` in https://github.com/solid/node-solid-server/blob/master/test/resources/accounts-acl/tim.localhost/group/test-folder/.acl#L19 makes the tests go into an infinite loop.

This should be investigated further, but for now, it seems to have relatively small impact.

kjetilk:
It might look as if this is connected to the ACL traversal, and if so, this may also be fixed by #777. 

Lets test again when we have that.

kjetilk:
It would not enter that loop anymore (because it has been entirely removed from the code), but the test is still commented out, as it causes all group tests to fail with a `500` and we should determine if that behaviour is what we expect. My hunch is that it is appropriate.

kjetilk:
Is fixed.

Issue #784, opened 2018-09-14
Server Tripped up by bugs in Turtle Docs


kidehen:
While testing the new Group ACL functionality I noticed regular crashes on my pod which ultimately boiled down to missing namespaces in my Turtle doc. 

You can use the following Group description to recreate the crashes (note I've commented out specific namespace prefixes for the desired effect. 

```
@prefix                :  <#> .
@prefix             acl:  <http://www.w3.org/ns/auth/acl#> .
@prefix            cmsb:  <http://ods-qa.openlinksw.com:8896/DAV/home/nobody/cmsb_handcrafted_webid_170223.ttl#> .
@prefix         csarven:  <http://csarven.ca/#> .
@prefix            foaf:  <http://xmlns.com/foaf/0.1/> .
@prefix         kidehen:  <https://kingsley.idehen.net/public_home/kidehen/profile.ttl#> .
@prefix   kidehen-solid:  <https://kidehen.solid.openlinksw.com:8444/profile/card#> .
@prefix  kidehen2-solid:  <https://kidehen2.solid.openlinksw.com:8444/profile/card#> .
@prefix  kidehen5-solid:  <https://kidehen5.solid.openlinksw.com:8444/profile/card#> .
@prefix  kidehen7-solid:  <https://kidehen7.solid.openlinksw.com:8444/profile/card#> .
@prefix  kidehen-community:  <https://kidehen.solid.community/profile/card#> .
@prefix          kjetil:  <https://solid.kjernsmo.net/profile/card#> .  
@prefix          melvin:  <https://melvincarvalho.com/#> .
@prefix   melvin2-solid:  <https://melvincarvalho.solid.openlinksw.com:8444/profile/card#> .
@prefix           ruben:  <https://ruben.verborgh.org/profile/#> .
@prefix     ruben-solid:  <https://ruben-just-works.solid.community/profile/card#> .
@prefix        smalinin:  <http://id.myopenlink.net/DAV/home/smalinin/YouID/Sergey_Malinin_Fb/profile.ttl#> .
@prefix  software-agent:  <https://id.myopenlink.net/DAV/home/KingsleyUyiIdehen/Public/software-agent.ttl#> .
@prefix         tallted:  <https://tallted.solid.openlinksw.com:8444/profile/card#> .
@prefix           timbl:  <https://www.w3.org/People/Berners-Lee/card#> .
@prefix       turnguard:  <https://ldp.turnguard.com/turnguard#> .
[...]

Issue #777, opened 2018-09-05
Replace solid-permissions with acl-check.js


kjetilk:
We need to replace https://github.com/solid/solid-permissions with https://github.com/solid/acl-check/ . This will happen in https://github.com/solid/node-solid-server/blob/master/lib/acl-checker.js

kjetilk:
Should be done with the merge of #970.

PR #970, opened 2018-11-25
Integrate acl-check

kjetilk:
This integrates the https://github.com/solid/acl-check package into the Access Control System of NSS.

We think this addresses the following issues:
  * #777 
  * #787
  * #785 
  * #784 

In addition, there are a few tests that have been changed after an audit, indicating that previous versions had flawed ACL handling.

This is intended to be merged to release a beta of NSS ASAP, so, please consider that in your reviews. The sustaining track will continue working on it for the official release, and so minor issues should be fixed then. Spare the nits for that. :-) Notably, the infrastructure for solving #468 has been provided, but there is something odd in the uppermost layers that prevents this from being communicated to the client. There is a large number of TODOs noted in the tests for this. #468 should be solved for 5.0.0, but possibly after the beta release.

There are also some tests that ideally should return a 403 rather than 401, like when Mallory is trying to attack with a malicious cookie. However, that is also not needed for the beta, as Mallory is not authorized.

BTW, you will also see a number of `console.log` coming from `acl-check.js` itself. We should fix this for the release, but I figured it is actually nice to have for the beta.

### DIFF
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -74,19 +74,20 @@ async function patchHandler (req, res, next) {
 function handler (req, res, next) {
   readEntity(req, res, () => patchHandler(req, res, next))
 }
+
 const readEntity = bodyParser.text({ type: () => true })
 
 // Reads the RDF graph in the given resource
 function readGraph (resource) {
   // Read the resource's file
   return new Promise((resolve, reject) =>
-    fs.readFile(resource.path, {encoding: 'utf8'}, function (err, fileContents) {
+    fs.readFile(resource.path, { encoding: 'utf8' }, function (err, fileContents) {
       if (err) {
         // If the file does not exist, assume empty contents
         // (it will be created after a successful patch)
         if (err.code === 'ENOENT') {
           fileContents = ''
-        // Fail on all other errors
+          // Fail on all other errors
         } else {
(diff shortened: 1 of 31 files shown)


lib/handlers/patch.js
```diff
@@ -53,30 +53,31 @@ function patchHandler (req, res, next) {
       .then(patchObject => checkPermission(target, req, patchObject))
   ])
   // Patch the graph and write it back to the file
-  .then(([graph, patchObject]) => applyPatch(patchObject, graph, target))
```
rubensworks: Not sure, but something may have gone wrong when rebasing.
Because I remember refactoring this file to be fully async, as can be seen here: https://github.com/solid/node-solid-server/blob/release/v5.0.0/lib/handlers/patch.js#L55-L66
So it's possible that there are other changes that are lost as well.

Maybe this can be double-checked?

rubensworks: Ah, never mind. That's just because this branch hasn't been rebased yet.

rubensworks:
Rebased on the release branch.

Let's merge once Travis finished.

megoth:
I think we can merge this even though security/snyk is failing - I'm guessing it's the bootstrap report it's complaining about?

PR #984, opened 2018-11-28
Fix/missing root acl

### LINKED ISSUE #977: Missing root .acl on new and existing servers
With the recent acl-check changes, .acl on root (such as `data/localhost/.acl`) is required.
However, this is not being generated because it is missing in `default-templates`.

An even bigger problem is that existing servers also won't have this .acl file, so all servers that upgrade will suddenly throw 401's everywhere.
An easy fix would be to auto-generate this root .acl file (from default-templates) when it can not be found at startup time.

rubensworks:
This adds a root .acl file to the templates folder,
and ensures that the file is generated at startup time for old servers.

Closes #977


rubensworks:
Some tests are failing because of this, will look into it.

kjetilk:
Oh, it got a bit harder to cooperate on your branch... I think you should be privileged to work directly on our? Here's a patch for four of the failures:

```
From 0ff262457ccdaab76378e9376cfa817eaec6b863 Mon Sep 17 00:00:00 2001
From: Kjetil Kjernsmo <kjetil@kjernsmo.net>
Date: Wed, 28 Nov 2018 16:07:34 +0100
Subject: [PATCH] Tests test for private resource

---
 test/integration/authentication-oidc-test.js | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/test/integration/authentication-oidc-test.js b/test/integration/authentication-oidc-test.js
index b92a34b..2088de5 100644
--- a/test/integration/authentication-oidc-test.js
+++ b/test/integration/authentication-oidc-test.js
@@ -147,7 +147,7 @@ describe('Authentication API (OIDC)', () => {
         describe('without that cookie', () => {
           let response
           before(done => {
-            alice.get('/')
+            alice.get('/private-for-alice.txt')
               .end((err, res) => {
                 response = res
                 done(err)
@@ -197,7 +197,7 @@ describe('Authentication API (OIDC)', () => {
           let response
           before(done => {
             var malcookie = cookie.replace(/connect\.sid=(\S+)/, 'connect.sid=l33th4x0rzp0wn4g3;')
-            alice.get('/')
+            alice.get('/private-for-alice.txt')
               .set('Cookie', malcookie)
               .end((err, res) => {
                 response = res
@@ -267,7 +267,7 @@ describe('Authentication API (OIDC)', () => {
         describe('without that cookie but with globally configured origin', () => {
           let response
           before(done => {
-            alice.get('/')
+            alice.get('/private-for-alice.txt')
               .set('Origin', 'https://apps.solid.invalid')
               .end((err, res) => {
                 response = res
@@ -285,7 +285,7 @@ describe('Authentication API (OIDC)', () => {
           let response
           before(done => {
             var malcookie = cookie.replace(/connect\.sid=(\S+)/, 'connect.sid=l33th4x0rzp0wn4g3;')
-            alice.get('/')
+            alice.get('/private-for-alice.txt')
               .set('Cookie', malcookie)
               .set('Origin', 'https://apps.solid.invalid')
               .end((err, res) => {
-- 
2.11.0
```

kjetilk:
Not sure about the two failures in `test/integration/account-creation-oidc-test.js`, nice if you can look further into that.

The failure in `test/integration/acl-tls-test.js` is testing to see if there is a server error if for some reason the root ACL wasn't there, and now it is... I think you could just skip it, we have the same test in acl-oidc-test

Also, this copies in .acl files into the tree, I suppose there should be something in the tests that delete those files.

rubensworks:
Ok, should be good to go now :-)

test/integration/account-creation-oidc-test.js
```diff
@@ -250,10 +250,10 @@ describe('Single User signup page', () => {
     fs.removeSync(rootDir)
   })
 
-  it('should return a 401 unauthorized without accept text/html', done => {
+  it('should return a 406 not acceptable without accept text/html', done => {
     server.get('/')
       .set('accept', 'text/plain')
-      .expect(401)
+      .expect(406)
```
kjetilk: Good catch!

test/integration/acl-tls-test.js
```diff
@@ -88,11 +88,15 @@ describe('ACL with WebID+TLS', function () {
...
-    it('should return 500 for any resource', function (done) {
+    before(() => {
+      rm(path.join(rootPath, '.acl'))
+    })
+
+    it('should return 403 for any resource', function (done) {
       var options = createOptions('/acl-tls/no-acl/', 'user1')
       request(options, function (error, response, body) {
         assert.equal(error, null)
-        assert.equal(response.statusCode, 500)
```
kjetilk: Actually, I think we shouldn't change this test, because it is a server misconfiguration to not have a root ACL. So, here it is mostly just that it became a tad harder to test. Since it is covered by a test in acl-oidc-test, if we can't easily fix the test, I think we should just skip it for now.

test/resources/accounts-acl/localhost/.acl
```diff
@@ -0,0 +1,10 @@
+# Root ACL resource for the root
```
kjetilk: I suppose these were generated by the test, so I would think that we shouldn't commit them, but rather have the test clean up after itself?

rubensworks: @kjetilk I actually left this in on purpose, since there is also an index.html and index.html.acl file in this folder.
Should I remove those as well? Or just leave it as-is?

kjetilk: Hmmm, good point. I guess it depends on how they are updated, if they are updated by the test script and that will overwrite any changes to it, then it shouldn't be committed.

rubensworks: Here's a list of the new .acl files + reason:
* `test/resources/accounts-acl/config/templates/server/.acl`: generated, like all other files in this config folder. (if we delete this, we should probably delete this whole config folder)
* `test/resources/accounts-acl/localhost/.acl`: needed for the unit tests, like the other files in this dir
* `test/resources/accounts/localhost/api/.acl`: needed for the unit tests
* `test/resources/headers/.acl`: needed for the unit tests, like the other files in this dir

kjetilk: OK, sounds good. And besides, not much can go wrong.

Issue #995, opened 2018-12-02
Is "solid" http or https?


JornWildt:
At https://github.com/solid/solid/blob/master/proposals/data-discovery.md we have:

    @prefix solid: <https://www.w3.org/ns/solid/terms#>

That is relevant for looking up `solid:publicTypeIndex`.

But in my profile document I have:

    @prefix solid: <http://www.w3.org/ns/solid/terms#>.

    <https://elfisk.solid.community/profile/card#me>
        solid:publicTypeIndex </settings/publicTypeIndex.ttl> ;

Now - which is right? The HTTPS version of `solid` or the HTTP :-)

Should I fix my profile document to use HTTPS - or should I fix my code to look for HTTP?

megoth:
I'm not able to find documentation that backs it up, but I would say the correct version should be HTTPS.

Some vocabularies and ontologies are using HTTP for legacy reasons, but if you dereference `http://www.w3.org/ns/solid/terms#` you'll see that it redirects to `https://www.w3.org/ns/solid/terms#`.

JornWildt:
I would also expect HTTPS ... and since I haven't touched my profile document, I'll be so free as to report it as a bug: solid-node-server should put HTTPS in the default profile document it generates for new users.

megoth:
Very good, I actually see a lot uses of `http://www.w3.org/ns/solid/terms#` in the default templates... Will fix very soon.

RubenVerborgh:
@timbl Which one is correct?

elf-pavlik:
It seems to have `http://` set with `@base`
https://github.com/solid/vocab/blob/master/solid-terms.ttl

JornWildt:
So does all this mean the spec at at https://github.com/solid/solid/blob/master/proposals/data-discovery.md is wrong and the solid-node-server template is right?

melvincarvalho:
> So does all this mean the spec at at solid/solid:proposals/data-discovery.md@master is wrong and the solid-node-server template is right?

It's a spec proposal.  Just a typo.  I've fixed it now.

> https://elfisk.solid.community/settings/privateTypeIndex.ttl 

I cant read that file, but looking at my own it seems to do the right thing.

csarven:
[...]

PR #996, opened 2018-12-02
Changed all cases of HTTP in solid terms to HTTPS

megoth:
Should fix https://github.com/solid/node-solid-server/issues/995


megoth:
Do not merge this before there is a conclusion in https://github.com/solid/node-solid-server/issues/995. (Not going to merge before tests are fixed anyway, that for some reason are failing, even though they run without problems for me locally...)

melvincarvalho:
Other way round.

PR #998, opened 2018-12-03
Add disable-password-checks option, Closes #983

### LINKED ISSUE #983: Add option to disable strict password checks
This may be annoying when creating and using test accounts during development.

rubensworks:
This makes development related to registration a lot more convenient,
and also allows users that do not need these strict checks to disable them.


rubensworks:
@megoth I agree with you on that (I initially had `enablePasswordChecks`).

However, I inverted it because I wanted to not make this a breaking change, and keep backwards compatibility with existing config files, where no `disablePasswordChecks`/`enablePasswordChecks` entry exists yet.
So by using `disablePasswordChecks`, existing config files do not have to be changed when they update to 5.0.0.

PR #1003, opened 2018-12-04
Fix registration from solid-auth-client not redirecting after login

rubensworks:
This is fixed by redirecting directly to the loginUrl after registration,
which includes the original query parameters,
and includes the original (encoded) redirect URL.

Closes solid/solid-auth-client#98


leinue:
Can this be merged as soon as possible?

PR #1010, opened 2018-12-05
Fix listings showing extensionless files incorrectly, Closes #1009

### LINKED ISSUE #1009: Profile is broken (release/v5.0.0)
The databrowser seems to handle the new resource mapping incorrectly.

For example, it lists the profile card as `card$.ttl`, while this `$.ttl` suffix should not be present.

As such, when displaying the contents, it gives an error:
screen shot 2018-12-05 at 10 02 20

rubensworks:
Fixes #1009

### DIFF
--- a/test/integration/ldp-test.js
+++ b/test/integration/ldp-test.js
@@ -285,18 +285,18 @@ describe('LDP', function () {
         '   dcterms:title "This is a container" ;' +
         '   o:limit 500000.00 .', 'sampleContainer/basicContainerFile.ttl')
 
-      return ldp.listContainer(path.join(__dirname, '../resources/sampleContainer/'), 'https://server.tld/resources/sampleContainer/', 'https://server.tld', '', 'text/turtle')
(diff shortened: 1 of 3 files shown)


test/integration/ldp-test.js
```diff
@@ -324,15 +324,18 @@ describe('LDP', function () {
     })
 
     it('should ldp:contains the same files in dir', () => {
-      ldp.listContainer(path.join(__dirname, '../resources/sampleContainer/'), 'https://server.tld/resources/sampleContainer/', 'https://server.tld', '', 'text/turtle')
+      ldp.listContainer(path.join(__dirname, '../resources/sampleContainer/'), 'https://server.tld/resources/sampleContainer/', '', 'server.tld')
         .then(data => {
           fs.readdir(path.join(__dirname, '../resources/sampleContainer/'), function (err, expectedFiles) {
+            // Strip dollar extension
+            expectedFiles = expectedFiles.map(ldp.resourceMapper._removeDollarExtension)
```
RubenVerborgh: We might not want to rely on `_` methods (but I guess okay in tests).

PR #1013, opened 2018-12-06
Fix server printing stacktrace after logging in, Closes #1000

### LINKED ISSUE #1000: Server stacktrace after login (release/v5.0.0)
After logging in, the server prints the following stacktrace:

```
(node:29945) UnhandledPromiseRejectionWarning: Error: Can't set headers after they are sent.
    at validateHeader (_http_outgoing.js:491:11)
    at ServerResponse.setHeader (_http_outgoing.js:498:3)
    at ServerResponse.header (/Users/rtaelman/Documents/UGent/nosync/node-solid-server/node_modules/express/lib/response.js:767:10)
    at addLink (/Users/rtaelman/Documents/UGent/nosync/node-solid-server/lib/header.js:22:9)
    at linksHandler (/Users/rtaelman/Documents/UGent/nosync/node-solid-server/lib/header.js:77:3)
    at <anonymous>
```

rubensworks:
This was because header.js tried to add headers after the response was already sent.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -74,8 +74,7 @@ async function handler (req, res, next) {
   if (!includeBody) {
     debug('HEAD only')
     res.setHeader('Content-Type', ret.contentType)
-    res.status(200).send('OK')
-    return next()
+    return res.status(200).send('OK')
   }
 
   // Handle dataBrowser


rubensworks:
Hmm, yes, you're right. I'll fix that right away!

rubensworks:
I've reverted my last change, as it broke a bunch of things.

So actually my initial fix was correct. All necessary headers are present, because`header.js` is called _before_ the `get.js`.
`next()` not being called here anymore is consistent with the databrowser and file logic.

RubenVerborgh:
Sweet, GitHub now shows force pushes.

Issue #729, opened 2018-08-02
Deleting a file should require write permission on the container


RubenVerborgh:
Possibly needs to be addressed in solid-permissions.

dmitrizagidulin:
Why on the container? (and not the file itself, which is how it currently is.)

dmitrizagidulin:
That doesn't quite make sense. If I upload the file somewhere, I should be able to delete it. (Without necessarily being able to delete or edit the container where I uploaded it..)

dmitrizagidulin:
(Also, we should probably have this discussion on https://github.com/solid/web-access-control-spec, no?)

RubenVerborgh:
This was just brought up in a meeting today, I'm mainly marking it for follow-up. The idea is that write access should enable people to empty a file, but not delete it. As on Linux systems, deleting should require write permissions on the folder, as you're changing the listing of that folder.

dmitrizagidulin:
Ahh, I didn't realize that was the Linux behavior..

RubenVerborgh:
Handling this in https://github.com/solid/solid-permissions/issues/28.

rubensworks:
Done in #1014.

PR #1014, opened 2018-12-06
Make DELETE require write permissions on container

rubensworks:
DELETE will now basically always check the permissions of the parent directory.

As requested in #729.

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -8,7 +8,7 @@ const fs = require('fs')
 const { promisify } = require('util')
 const HTTPError = require('../http-error')
 
-function allow (mode, relativePath = '') {
+function allow (mode, testDirectory) {
   return async function allowHandler (req, res, next) {
     const ldp = req.app.locals.ldp || {}
     if (!ldp.webid) {
@@ -25,8 +25,8 @@ function allow (mode, relativePath = '') {
       : req.path
 
     // If a relativePath has been provided, check permissions based on that
-    if (relativePath) {
-      reqPath = path.join(reqPath, relativePath)
+    if (testDirectory) {
+      reqPath = path.dirname(reqPath)
     }
 
     // Check whether the resource exists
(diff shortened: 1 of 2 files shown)


lib/ldp-middleware.js
```diff
@@ -26,7 +26,7 @@ function LdpMiddleware (corsSettings) {
   router.post('/*', allow('Append'), post)
   router.patch('/*', allow('Append'), patch)
   router.put('/*', allow('Write'), put)
-  router.delete('/*', allow('Write'), del)
+  router.delete('/*', allow('Write', '..'), del)
```
kjetilk: Not sure, but wouldn't `./` be the path to the container?

rubensworks: The reason why that was needed is because I used `path.join`, which considers everything to be a path segment, including the filename itself, so `./` would do nothing.

In any case, I rewrote this PR a bit, as it was indeed a bit confusing.

kjetilk: OK; perhaps you can have a look too, @RubenVerborgh ?

lib/handlers/allow.js
```diff
@@ -1,13 +1,14 @@
...
 const $rdf = require('rdflib')
+const path = require('path')
 const ACL = require('../acl-checker')
 const debug = require('../debug.js').ACL
 const fs = require('fs')
 const { promisify } = require('util')
 const HTTPError = require('../http-error')
 
-function allow (mode) {
+function allow (mode, testDirectory) {
```
RubenVerborgh: Unclear parameter name IMHO

lib/handlers/allow.js
```diff
@@ -23,6 +24,11 @@ function allow (mode) {
       ? res.locals.path
       : req.path
 
+    // If a relativePath has been provided, check permissions based on that
```
RubenVerborgh: Comment out of sync?

lib/handlers/allow.js
```diff
@@ -23,6 +24,11 @@ function allow (mode) {
...
+    if (testDirectory) {
+      reqPath = path.dirname(reqPath)
```
RubenVerborgh: Maybe deceptive to use `reqPath` for it then. Perhaps should be something like `resourcePath`.

PR #1027, opened 2018-12-13
Use http schema for solid terms

### LINKED ISSUE #1002: type index uses the wrong namespace
As reported here https://github.com/solid/node-solid-server/issues/995#issuecomment-443675337

```
@prefix solid: <https://www.w3.org/ns/solid/terms#>.
<> a solid:TypeIndex
```

Is being produced in some type indexes.

My own is fine : https://melvin.solid.community/settings/publicTypeIndex.ttl

```
@prefix solid: <http://www.w3.org/ns/solid/terms#>.
```

But I can confirm that more recent accounts have the incorrect https namespace.  This is a **major** bug, and we need to urgently find the source.

kjetilk:
Should fix #1002. 

Done against 5.0.0, can be cherry-picked to 4.4 if needed.


melvincarvalho:
LGTM

PR #1034, opened 2018-12-19
Add the serverUri as trusted origin for multiuser

### LINKED ISSUE #1026: POD provider Origin also trusted by default?
The error message in https://github.com/EFForg/privacybadger/issues/2240#issuecomment-446429064 made my think about the following: May it be that a apps on user's POD on a multiuser setup, e.g. https://testpb.dev.inrupt.net/ will not only load from self, but also from the Provider's https://dev.inrupt.net/ ? In which case the latter should also be trusted by default?

kjetilk:
This should fix #1026 .

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -43,6 +43,10 @@ function allow (mode, checkPermissionsForDirectory) {
       resourcePath += '/'
     }
 
+    let trustedOrigins = [ldp.resourceMapper.resolveUrl(req.hostname)].concat(ldp.trustedOrigins)
+    if (ldp.multiuser) {
+      trustedOrigins.push(ldp.serverUri)
+    }
     // Obtain and store the ACL of the requested resource
     req.acl = new ACL(rootUrl + resourcePath, {
       agentOrigin: req.get('origin'),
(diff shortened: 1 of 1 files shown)


lib/handlers/allow.js
```diff
@@ -43,6 +43,10 @@ function allow (mode, checkPermissionsForDirectory) {
       resourcePath += '/'
     }
 
+    let trustedOrigins = [ldp.resourceMapper.resolveUrl(req.hostname)].concat(ldp.trustedOrigins)
```
megoth: An even more readable version would be to have it in separate function; but this is good ^_^

lib/handlers/allow.js
```diff
@@ -43,6 +43,10 @@ function allow (mode, checkPermissionsForDirectory) {
...
+    if (ldp.multiuser) {
```
dmitrizagidulin: Shouldn't we also include `serverUri` as a trusted origin in single-user mode?

kjetilk: Yeah, so the `resolveUrl(req.hostname)` would then be that... So, you could turn this patch around, actually, with the same result. I think.

PR #1036, opened 2018-12-24
Add npm script for solid-test

ninachaubal:
Adds an npm script for `solid-test` so it has a similar workflow to `solid`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
   "main": "index.js",
   "scripts": {
     "solid": "node ./bin/solid",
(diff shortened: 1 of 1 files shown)


kjetilk:
Just returning to review this. I'm hesitant to add this, because solid-test is currently a shell-script, so if users on other platforms come in and expect it to work, they will be surprised. Therefore, I would have preferred to have a the solid-test script portable first.

ninachaubal:
That's fair. Closing this PR.

Issue #468, opened 2016-12-06
When ACL fails 401 due to Origin check, important to say so in logs and HTTP status


timbl:
Currently, there is an option (but important) feature which checks, when a Web App (a given Origin) accesses data on behalf of  user, that not only is the user authorized but also the origin.
At the moment, when the origin test fails (which devs are not used to) the console log and HTTP message are identical to when the user themselves is not authorized -- and quite misleading.

Suggest the log messages ask as well as

```
"Can <agent>  <method> <resource>?"
```
separately ask
```
"Can agent <agent>  <method> <resource>?"
"Can origin <origin>  <method> <resource>?"
```
and give a different error

**401   Unauthorized Application**

as distinct from say

401  Unauthorized person

kjetilk:
We could do this by rather than return 
```js 
Promise.resolve(false)
```
when the agent is not authorized, return 
```js 
Promise.reject(reason)
```

It is a bit complicated to derive the `reason` though.

kjetilk:
The fix should be #1050, please review.

PR #1042, opened 2019-01-11
Serves statusMessage based on what response acl-check gives

megoth:
This should fix https://github.com/solid/node-solid-server/issues/468 .

There might be a couple of quota-related tests that fail (they do on my local setup), which is weird, as these changes should not affect that code.

@kjetilk and I are working on the same files, so if this PR is merged before his, I'll do the work when his PR is ready to be merged.


megoth:
The new commits are rebased on https://github.com/solid/node-solid-server/pull/1043, so do not merge this before that one is merged.

megoth:
The diffs seems to a off, so I'll try to recreate this PR

Issue #1029, opened 2018-12-15
Cannot register account ("Account already exists")


angelo-v:
Solid Server Versions:
* **5.0.0-beta.2**
* **5.0.0-beta.3**
* **5.0.0-beta.4**

*Expected Behaviour*: After setting up a solid server it should be possible to register an account

*Actual Behaviour*: When trying to register any account by submitting the registration form, the following error occurs: "Account already exists"

*Steps to reproduce*:

1. Visit https://angelo.veltens.org or set up your own solid-server on version 5.0.0-beta.3
2. Click `Register`
3. Fill out the registration form
4. Click `Submit`
5. See the mentioned error message on top of the page

RubenVerborgh:
Are you in multiuser mode?

angelo-v:
Since I did not enable it explicitly I guess not

angelo-v:
This is the form data that is transmitted to `api/accounts/new`:

```
username: solid-demo
password: *******
repeat_password: *******
name: solid-demo
email: me@mail.example
externalWebId: 
returnToUrl: 
response_type: 
display: 
scope: 
client_id: 
redirect_uri: 
state: 
nonce: 
request: 
```
Is it correct, that the latter params are all empty?

angelo-v:
Unfortunately the solid-server logs are quite silent... It's hard to find out what is going on. Is there some verbose mode or something?

kjetilk:
Sorry, I want to introduce a multi-level logger for exactly these occasions, but we haven't gotten to that yet. I'll try to see if I can reproduce the behaviour when getting back to work on Monday.

kjetilk:
> Unfortunately the solid-server logs are quite silent... It's hard to find out what is going on. Is there some verbose mode or something?

Actually, up to recently, I had verbose output with no further ado, but for some reason, I now need to set an environment variable, e.g.
```bash
export DEBUG=solid:*
```

does that improve the logging situation for you, @angelo-v ?

angelo-v:
Thanks, I tried that already, but didn't help. If you say it should, I will try again, to be sure I had no typo or something

angelo-v:
Ok, got it working, but doesn't help that much:
```
 solid:accounts Canceling account creation, https://angelo.veltens.org/profile/card#me already exists 
```

megoth:
I've been able to reproduce this bug (not being able to create accounts when server is set to singleuser; I'll look into this when I have time.

kjetilk:
[...]

PR #1051, opened 2019-01-15
Fix/singleuser

joachimvh:
This fixes issue #1029 .

The problem was that the root .acl file was used to check if the user was already registered. But in the case of single user mode, there already was an .acl file present in the root since there is no seperate user folder there. The solution was to check the existence of a file that only exists after profile creation, the profile card file in this case.

For some reason this causes a corresponding unit test to fail though. There seems to be a problem with the acl file being used in that test, but I didn't manage to figure out yet what exactly the problem is now. Profile registration works for me in the server when running normally with this fix so it might be something to do with how the dummy files are generated for the tests, but I'm not sure.

### DIFF
--- a/lib/models/account-manager.js
+++ b/lib/models/account-manager.js
@@ -83,18 +83,18 @@ class AccountManager {
    */
   accountExists (accountName) {
     let accountUri
-    let rootAclPath
+    let cardPath
 
     try {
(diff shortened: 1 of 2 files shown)


joachimvh:
Yea I also looked into having the same structure in single user mode, but it noticed that there are several locations in the code that depend on having this structe in single user mode.

PR #1052, opened 2019-01-15
Fix that should create a correct base URI wrt validation when PUTing TTL-files

megoth:


megoth:
This is a fix for https://github.com/solid/solid/issues/235

kjetilk:
Still LGTM.

lib/handlers/put.js
```diff
@@ -36,7 +36,8 @@ async function putStream (req, res, next, stream = req) {
 async function putAcl (req, res, next) {
   const ldp = req.app.locals.ldp
   const contentType = req.get('content-type')
-  const requestUri = `${req.protocol}//${req.get('host')}${req.originalUrl}`
+  const path = ldp.getFullPath(req)
+  const requestUri = await ldp.getRequestUri(path, req.hostname)
```
rubensworks: It should be possible to simplify this to something like `requestUri = await ldp.resourceMapper.resolveUrl(req.hostname, req.path)`

Then you don't need to introduce the new helper functions in `ldp.js`, as they server a similar purpose to `resolveUrl` as far as I can see.

megoth: I'll do these changes and make a new commit

megoth:
Reverted some of my previous changes, realized I didn't understand the uses of some methods in resourceMapper properly.

But this should fix the original problem with faulty URL in `handlers/put.js`.

Issue #1053, opened 2019-01-16
Acknowledge all contributors to this repository


RubenVerborgh:
The current README only lists people who have contributed through lines of code. We should acknowledge people who contributed in various other means, such as through issues, testing, interactions on chat, etc.

RubenVerborgh:
Two possible solutions have been contributed so far:
- #1054, with a sectioned list to explicitly acknowledge non-code contributions
- #1055, with a flat list

RubenVerborgh:
Following up for the Solid project as a whole: https://github.com/solid/solid/issues/243

melvincarvalho:
No strong view on this.  But perhaps some food for thought.  Last time this topic came up, I mentioned the all-contributors standard : 

https://github.com/all-contributors/all-contributors

Which has some guidelines, tooling and bots.  It is used by some other free software projects and contains various [categories](https://all-contributors.js.org/docs/overview#emoji-key) of contribution, each with their own icon.  

Not specifically endorsing this approach, but it may be of interest to see how others go about solving this issue.

RubenVerborgh:
Good point, I linked to it as well in the PR for this issue: https://github.com/solid/node-solid-server/pull/1054#issuecomment-454961331. I'm cautious about the maintenance overhead it might generate, but would be nice to explore. Perhaps #1054 can be a temporary solution, and then be replaced by a more advanced version.

PR #1054, opened 2019-01-16
Add contributors document with sections

RubenVerborgh:
Alternative to #1055. Difference: this version organizes contributors in sections to distinguish current code contributors from past code contributors, and to list non-code contributors (issues, support, testing, …).

----

This PR aims to address #1053 by:
- adding a `CONTRIBUTORS.md` document
- adjusting `LICENSE.md` and `package.json` to explicitly include names of code contributors
- adjusting `README.md` to link to these

May I ask for your feedback regarding:
- the wording used
- the people included on those lists
- anything else you want to change or add

I emphasize that my effort is not a value judgement whatsoever; I arrived a long time after the Solid project started, so it is very likely that I have not included sufficient people in those lists. Please see this PR as a template, not a gospel.

Thanks for your help.

Closes https://github.com/solid/node-solid-server/issues/1053.

PR #1055, opened 2019-01-16
Add contributors document without sections

RubenVerborgh:
Alternative to #1054. Difference: this version lists all contributors alphabetically, without distinguishing between code, issues, and currently active or not.

----

This PR aims to address #1053 by:
- adding a `CONTRIBUTORS.md` document
- adjusting `LICENSE.md` and `package.json` to explicitly include names of code contributors
- adjusting `README.md` to link to these

May I ask for your feedback regarding:
- the wording used
- the people included on those lists
- anything else you want to change or add

I emphasize that my effort is not a value judgement whatsoever; I arrived a long time after the Solid project started, so it is very likely that I have not included sufficient people in those lists. Please see this PR as a template, not a gospel.

Thanks for your help.

Closes https://github.com/solid/node-solid-server/issues/1053.


dan-f:
Hey @RubenVerborgh -

Looks like you forgot to include me on the list 😉

RubenVerborgh:
> Looks like you forgot to include me on the list 😉

That was a deliberate omission 🙂 You might have forgotten, but when releasing v4.0.0 in the summer if 2017, I was updating the collaborators lists and asked you whether you wanted to be included, and you preferred not to.

Happy of course to add you, but going by the heuristic in https://github.com/solid/node-solid-server/pull/1054#issuecomment-454957666 (`git shortlog -sn --no-merges release/v5.0.0 -- "**/*.js"`), the number  of commits I find for you is 2, so we would probably be including 4 others then as well.

Let me know what you prefer, and I will adjust accordingly.

dan-f:
So I'm assuming that the context of this list is scoped down to people directly involved in `node-solid-server`. Since most of my work was focused on front-end, architecture, and querying, I'm certainly not claiming profound influence over `node-solid-server`. Others on the team, including yourself, @deiu, @nicola, and @dmitrizagidulin, all of whom I worked with, focused much more effort on building out the server itself.

I think there is a broader discussion to be had that should not necessarily block the recognition of folks for their hard work (don't let this hold up the PR). That said, I'll leave my thoughts here, as I'm not sure where else to place them.

My main concern, from the perspective of a community member, as well as a former core team member, is regarding the idea that certain contributions "count" while others don't. And in particular, the notion that a contribution is a quantitative value, for which we can add numbers to determine what counts.

> git shortlog -sn --no-merges release/v5.0.0 -- "**/*.js"

Some thoughts on the particular heuristic:
- Is there a strong correlation between "number of commits" and value of contributions? What about number of bytes flipped in a diff?
- Is there a static number of commits below which an author has not contributed?
- Does the community feel that only commits touching JS files (`"**/*.js"`) constitute a contribution?

Interestingly, one of the commits that matched the `**/*.js` criteria was my simply updating a bundle of minified JS (separate discussion as to whether vendoring is a good practice here or not). This also probably would have probably satisfied a "size of changes" predicate. Another patch updating documentation for 4.0.0 got filtered out. Which change is more important?

Also -

> That was a deliberate omission 🙂 You might have forgotten, but when releasing v4.0.0 in the summer if 2017, I was updating the collaborators lists and asked you whether you wanted to be included, and you preferred not to.

Yikes, I don't remember that, and while I don't want to discount your experience here, I'd be pretty surprised if that's what I actually said to you. This is probably the only point I'd prefer to discuss offline  if you feel like you need to follow-up.

And finally, lest this come off as nitpicky and/or ungrateful, I've got a lot of respect for the core contributors to `node-solid-server`, community members on gitter, and the faculty, staff, and students at CSAIL, many of whom offered piercing questions, valuable architectural feedback, and generally excellent company.

RubenVerborgh:
> So I'm assuming that the context of this list is scoped down to people directly involved in `node-solid-server`. Since most of my work was focused on front-end, architecture, and querying, I'm certainly not claiming profound influence over `node-solid-server`.

Also taking @csarven's comments into account, we probably want a list of contributors for the Solid project as a whole, as this list is indeed specific to node-solid-server. Will add a similar PR for solid/solid.

> My main concern, from the perspective of a community member, as well as a former core team member, is regarding the idea that certain contributions "count" while others don't.

Any heuristic is of course flawed. The only way to not miss anyone, is to add every single person who committed or commented to the list.

However, this might also not be fair: there are (literally) commits where a one-letter typo was changed, and this is the only contribution of that person. Putting such contributions, while useful, on the same level as a one-year dedication to server code, would likely also send the wrong signal.

Hence my attempt to distinguish between incidental committers and regular committers, without any value judgement.

>     * Is there a strong correlation between "number of commits" and value of contributions? What about number of bytes flipped in a diff?

No strong correlation indeed.

For instance, I'm the number two committer in terms of commits, but that is largely because I make more granular commits than most. I only worked on the server for 3 months, and on a very specific part.

In number of lines of code added, I'm the very last, with a contribution of -5,673 lines.

However, it so happens that this metric shows two clear groups: one with than 10 commits each, and one with tens / hundreds each.

>     * Is there a static number of commits below which an author has not contributed?

All authors have contributed, I would say. But there's a difference in significance. I have committed to dozens of projects on GitHub, but I consider myself only a co-author on a part of them.

>     * Does the community feel that only commits touching JS files (`"**/*.js"`) constitute a contribution?

They are all contributions.

But *.js files are code contributions.

This was mainly to weed out large diffs from package-lock.json and version bumps.

> Interestingly, one of the commits that matched the `**/*.js` criteria was my simply updating a bundle of minified JS (separate discussion as to whether vendoring is a good practice here or not). This also probably would have probably satisfied a "size of changes" predicate. Another patch updating documentation for 4.0.0 got filtered out. Which change is more important?

This is not a judgment on importance; none of these heuristics would be.

The only thing we're trying to do is distinguishing regular and occasional contributors from
incidental contributors.

> This is probably the only point I'd prefer to discuss offline if you feel like you need to follow-up.

Sending you an email.

> And finally, lest this come off as nitpicky and/or ungrateful

Certainly not, we should recognize the effort of people.

deiu:
Looking at this objectively, the idea of counting contributions of a former core team member based purely on metrics is just plain wrong.

RubenVerborgh:
Making a list of contributors for solid/solid at https://github.com/solid/solid/pull/244.

RubenVerborgh:
Making a list of contributors for solid/solid at https://github.com/solid/solid/pull/244.

dan-f:
I'm fine to be left out of the `node-solid-server` contributors doc, but would appreciate being included in the `solid/solid` one. Thanks!

RubenVerborgh:
@dan-f Thanks for the feedback! You're in solid/solid: https://github.com/solid/solid/pull/244

PR #1058, opened 2019-01-21
[Discuss] Read-only access

michielbdejong:
With this PR, the pod owner can decide to log in to their pod read-only when using a third-party app.

### DIFF
--- a/default-templates/new-account/.acl
+++ b/default-templates/new-account/.acl
@@ -19,5 +19,20 @@
     acl:mode
         acl:Read, acl:Write, acl:Control.
 
+<#ownerreadonly>
+    a acl:Authorization;
+
+    acl:agent <{{webIdReadOnly}}> ;
+
+    # Set the access to the root storage folder itself
+    acl:accessTo </>;
+
+    # All resources will inherit this authorization, by default
+    acl:defaultForNew </>;
+
+    # When authenticated in read-only mode, the owner only has the read mode
+    acl:mode
+        acl:Read.
+
 # Data is private by default; no other agents get access unless specifically
 # authorized in other .acls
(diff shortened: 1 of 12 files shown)


michielbdejong:
This PR adds a 'Read-only access' checkbox to the login form. If checked, you'll be logged in as https://pod.com/profile/card#me-read-only instead of the usual https://pod.com/profile/card#me.
Then when they try to create a file, it will be like when you try that on someone else's pod; you'll see an error like:

`Error: FAILED to save new tool at: https://localhost:8443/public/kjhllujh/index.ttl : Web error: 403 (Forbidden) on PUT of <https://localhost:8443/public/kjhllujh/index.ttl>`

michielbdejong:
Great! Note that this is a bit of a strawman proposal; maybe it helps us arrive at a better proposal that can replace it.

I first thought it would be useful to give Read+Append to e.g. a Chess app, but that's awkward because then the app can still only write the game state once, and never edit it. So then I implemented the read-only option, which only solves part of the open issues discussed in https://forum.solidproject.org/t/inter-app-access-control/184/26 and https://forum.solidproject.org/t/read-only-or-sub-folder-oidc-scopes/767.

RubenVerborgh:
Good thinking work @michielbdejong, it's indeed one way to do it, and we should come up with some alternatives, so we can compare.

I think what ultimately should happen (but that is my rough opinion here) is that an app should be associated with an agent, that agent is acting on behalf of me, and that agent is delegated certain permissions.

@kidehen might also have an opinion on this.

So let's not merge this one just yet (I will formally reject so it does not get merged by acccident), but let's discuss indeed.

kjetilk:
I think this could be interesting as a temporary measure, i.e., we could consider having it in 5.0.0, without committing to bring it on to later releases, where we should have identity on apps and ACL associated with them.

michielbdejong:
Superseded by strictOrigins+trustedApps system.

Issue #918, opened 2018-11-08
UI to read the remaining space


kjetilk:
Quota enforcement is nearly finished in #841 , but obviously, we'd like the user to be able to track the amount that they are using. 

The easiest fix for this is to have a simple API exposure, but AFAIK, we can't use an ACL to restrict who can see the remaining quota, and I don't think we should expose it.

So, the better solution is to put it in the RDF along with the quota. That's more work, but it also means that the space left doesn't need to be computed real-time, so it is a solution that would scale better. 

However, as there are many issues open, it needs a different prioritization, so I'm opening this issue for it.

megoth:
I've proposed a property to indicates how much disk space is in use (https://github.com/solid/vocab/issues/37), but think I also need a property that indicates where serverSide.ttl resides, so that I can fetch that data. Or am I missing an existing property for this already?

RubenVerborgh:
Regarding serverSide.ttl: https://github.com/solid/vocab/issues/37#issuecomment-438685697

TallTed:
This may not be the best issue to add this comment; feel free to redirect me/it. 

In catching up on a lot of discussion and updates, it seems that when the "test quota on write" fails it results in a notification to the attempting writer -- so if @kjetilk tries to write to @RubenVerborgh's INBOX and that write would exhaust quota, *@kjetilk* gets notified -- but @RubenVerborgh does not. 

This is troublesome, as @kjetilk's best (and possibly only) way to inform @RubenVerborgh of this problem is via his inbox ... which is full. I think that this is a place where NSS might use @RubenVerborgh's email address to send such a notification without consuming further NSS-quota.

megoth:
> This may not be the best issue to add this comment; feel free to redirect me/it.
> 
> In catching up on a lot of discussion and updates, it seems that when the "test quota on write" fails it results in a notification to the attempting writer -- so if @kjetilk tries to write to @RubenVerborgh's INBOX and that write would exhaust quota, _@kjetilk_ gets notified -- but @RubenVerborgh does not.
> 
> This is troublesome, as @kjetilk's best (and possibly only) way to inform @RubenVerborgh of this problem is via his inbox ... which is full. I think that this is a place where NSS might use @RubenVerborgh's email address to send such a notification without consuming further NSS-quota.

I think this is better suited as its own issue? But I think it's definitely something we should consider some solution for.

megoth:
[...]

PR #1060, opened 2019-01-22
Virtual resource for account resource and visualizing that data on frontpage

megoth:
This is a fix for https://github.com/solid/node-solid-server/issues/918

I have not written tests for this, but plan to do that ASAP. But would be nice to have some eyes on the code in case I've missed something.

### DIFF
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -74,8 +74,26 @@
     // do stuff
     storagePanel.classList.remove('hidden')
     const quotaValue = quotaNode.value
-    storageQuota.innerText = quotaValue
+    storageQuota.innerText = formatBytes(quotaValue)
     const usageValue = usageNode.value
-    storageUsage.innerText = usageValue
+    storageUsage.innerText = formatBytes(usageValue)
+  }
+
+  function formatBytes (number, units = [
+    'B',
+    'kB',
+    'MB',
+    'GB',
+    'TB',
+    'PB',
+    'EB',
+    'ZB',
+    'YB'
+  ]) {
+    const unit = units.shift()
+    if (number > 1000 && units.length > 0) {
+      return formatBytes(number / 1000, units)
+    }
+    return `${Math.round(number * 100) / 100} ${unit}`
   }
 })(solid, window.$rdf)
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -72,7 +72,7 @@
       return
     }
 
-    // do stuff
+    // manipulate DOM to present data
     storagePanel.classList.remove('hidden')
     const quotaValue = quotaNode.value
     storageQuota.innerText = formatBytes(quotaValue)
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -1,7 +1,7 @@
 /* Provide functionality for authentication buttons */
 /* global URL, location, alert, solid */
 
-(({ auth }) => {
+(({ auth }, $rdf) => {
   // Wire up DOM elements
   const [loginButton, logoutButton, registerButton, accountSettings] =
     ['login', 'logout', 'register', 'account-settings'].map(id =>
@@ -9,7 +9,9 @@
   loginButton.addEventListener('click', login)
   logoutButton.addEventListener('click', logout)
   registerButton.addEventListener('click', register)
-  document.addEventListener('DOMContentLoaded', getAccountInfo)
+  if ($rdf) {
+    document.addEventListener('DOMContentLoaded', getAccountInfo)
+  }
 
   // Track authentication status and update UI
   auth.trackSession(session => {
@@ -50,8 +52,30 @@
   }
 
   async function getAccountInfo () {
-    const account = new URL('/account', location)
-    const response = await auth.fetch(account)
-    console.log('test3', response)
+    const SOLID = $rdf.Namespace('http://www.w3.org/ns/solid/terms#')
+
+    // load data
+    const accountUrl = new URL('/account', location).href
+    const accountStore = $rdf.graph()
+    const accountFetcher = new $rdf.Fetcher(accountStore)
+    await accountFetcher.load(accountUrl)
+
+    // get DOM and data-dependencies
+    const quotaNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageQuota'))
+    const usageNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageUsage'))
+    const storagePanel = document.getElementById('StoragePanel')
+    const storageQuota = document.getElementById('StorageQuota')
+    const storageUsage = document.getElementById('StorageUsage')
+    if (!quotaNode || !usageNode || !storagePanel || !storageQuota || !storageUsage) {
+      // If anything of this is not available, we do not need to do anything more
+      return
+    }
+
+    // do stuff
+    storagePanel.classList.remove('hidden')
+    const quotaValue = quotaNode.value
+    storageQuota.innerText = quotaValue
+    const usageValue = usageNode.value
+    storageUsage.innerText = usageValue
   }
-})(solid)
+})(solid, window.$rdf)
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -1,4 +1,5 @@
 /* Provide functionality for authentication buttons */
+/* global URL, location, alert, solid */
 
 (({ auth }) => {
   // Wire up DOM elements
@@ -8,6 +9,7 @@
   loginButton.addEventListener('click', login)
   logoutButton.addEventListener('click', logout)
   registerButton.addEventListener('click', register)
+  document.addEventListener('DOMContentLoaded', getAccountInfo)
 
   // Track authentication status and update UI
   auth.trackSession(session => {
@@ -45,5 +47,11 @@
     const registration = new URL('/register', location)
     registration.searchParams.set('returnToUrl', location)
     location.href = registration
+  }
+
+  async function getAccountInfo () {
+    const account = new URL('/account', location)
+    const response = await auth.fetch(account)
+    console.log('test3', response)
   }
 })(solid)
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -54,21 +54,25 @@
   async function getAccountInfo () {
     const SOLID = $rdf.Namespace('http://www.w3.org/ns/solid/terms#')
 
+    // get DOM and data-dependencies
+    const storagePanel = document.getElementById('StoragePanel')
+    const storageQuota = document.getElementById('StorageQuota')
+    const storageUsage = document.getElementById('StorageUsage')
+    const storageUsedBar = document.getElementById('StorageUsedBar')
+    if (!storagePanel || !storageQuota || !storageUsage || !storageUsedBar) {
+      // If anything of these elements are not available, we do not need to do anything more
+      return
+    }
+
     // load data
     const accountUrl = new URL('/account', location).href
     const accountStore = $rdf.graph()
     const accountFetcher = new $rdf.Fetcher(accountStore)
     await accountFetcher.load(accountUrl)
-
-    // get DOM and data-dependencies
     const quotaNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageQuota'))
     const usageNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageUsage'))
-    const storagePanel = document.getElementById('StoragePanel')
-    const storageQuota = document.getElementById('StorageQuota')
-    const storageUsage = document.getElementById('StorageUsage')
-    const storageUsedBar = document.getElementById('StorageUsedBar')
-    if (!quotaNode || !usageNode || !storagePanel || !storageQuota || !storageUsage || !storageUsedBar) {
-      // If anything of this is not available, we do not need to do anything more
+    if (!quotaNode || !usageNode) {
+      // If data is not available, we do not need to do anything more
       return
     }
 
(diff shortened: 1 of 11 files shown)


common/js/auth-buttons.js
```diff
@@ -46,4 +50,55 @@
...
+    const storagePanel = document.getElementById('StoragePanel')
+    const storageQuota = document.getElementById('StorageQuota')
+    const storageUsage = document.getElementById('StorageUsage')
+    const storageUsedBar = document.getElementById('StorageUsedBar')
+    if (!quotaNode || !usageNode || !storagePanel || !storageQuota || !storageUsage || !storageUsedBar) {
+      // If anything of this is not available, we do not need to do anything more
+      return
+    }
+
+    // do stuff
```
kjetilk: Not quite a fan of such comments, they don't really add any value... :-)

megoth: Fixed in commit ee1a116fca8c569051465432bfae1ffdb911811a ^_^

lib/acl-checker.js
```diff
@@ -144,6 +146,49 @@ class ACLChecker {
   isAcl (resource) {
     return resource.endsWith(this.suffix)
   }
+
```
kjetilk: Is this refactoring necessary to solve implement this issue? I can't tell. If not, would it be better to do the refactor in a separate PR to ease the review of this feature?

megoth: If I don't do it like this, I will copy the code instead of refactoring it =/ I thought this a better approach.

That said, I agree it's a bit messy, and am open for suggestions on other approaches.

megoth:
A problem with this implementation is that it returns quite a lot of 404-statuses, to which there are two problems:

1) Sometimes it is actually referring to a missing value, e.g. serverSide.ttl doesn't have triple for `solid:storageQuota`, and
2) It may look a bit weird to get a 404 when serverSide.ttl doesn't exist; it _might_ be by design, since POD providers don't necessarily enforce a quota on their PODs

Any thoughts on this? My thinking was that it's ok to return 404, since the result for the consuming client code is that it's just ignored. But it might be annoying for developers checking console for various reasons to see a 404 being returned (especially annoying if they're trying to debug and is side-railed).

kjetilk:
> It may look a bit weird to get a 404 when serverSide.ttl doesn't exist; it might be by design, since POD providers don't necessarily enforce a quota on their PODs

Oh, yeah, that is by design and documented as the way to disable the quota system. So, basically, parsing the file takes around ~200 ms, and then computing the size of the file system takes ~100 ms, so the idea was that you can avoid this cost by not having the `serverSide.ttl` file, then the cost is just that of a `stat` on the file. 

But then, I might not quite understand what will return a 404, is it just the `get-account-info-request` resource?

megoth:
> But then, I might not quite understand what will return a 404, is it just the `get-account-info-request` resource?

As the code is written now, there are various things that could go wrong:

1) [Line 46](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R46): User does not have a userId (i.e. an unauthenticated visitor) --> Return 401
2) [Line 57](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R57): User does not have access to POD (i.e. an authenticated visitor) --> Return whatever acl-check returns
3) [Line 66](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R66): There was an error when dereferencing the graph from user's WebID --> Return the error
4) [Line 74](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R74): There wasn't a triple denoting where storage is located --> Return 404
5) [Line 84](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R84): There was an error when dereferencing the graph for serverSide.ttl --> Return the error
6) [Line 92](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R92): There wasn't a triple denoting the quota -> Return 404
7) [Line 107](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R107): There was an error when serializing the graph that represents the content of `/account` --> Return the error (probably 500)

The missing triples might be a stretch to return 404s, but my thinking was that there is some data missing, hence we don't have an account resource to return. But thinking more of it, we could also return 200 and a graph that doesn't have those triples (i.e. an empty graph for now).

We could apply the same strategy when we cannot dereference the various graphs (WebID, serverSide.ttl).

common/js/auth-buttons.js
```diff
@@ -46,4 +50,55 @@
...
+    await accountFetcher.load(accountUrl)
+
+    // get DOM and data-dependencies
+    const quotaNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageQuota'))
+    const usageNode = accountStore.any($rdf.sym(accountUrl), SOLID('storageUsage'))
+    const storagePanel = document.getElementById('StoragePanel')
+    const storageQuota = document.getElementById('StorageQuota')
+    const storageUsage = document.getElementById('StorageUsage')
+    const storageUsedBar = document.getElementById('StorageUsedBar')
+    if (!quotaNode || !usageNode || !storagePanel || !storageQuota || !storageUsage || !storageUsedBar) {
```
rubensworks: I suggest moving this check _before_ the fetch call. This would avoid any unneeded calls for pages that don't show the quota.

lib/requests/get-account-info-request.js
```diff
@@ -0,0 +1,115 @@
...
+const AuthRequest = require('./auth-request')
+const { actualSize } = require('../utils')
+const $rdf = require('rdflib')
+const HTTPError = require('../http-error')
+const LRU = require('lru-cache')
+
+const PIM = $rdf.Namespace('http://www.w3.org/ns/pim/space#')
+const SOLID = $rdf.Namespace('http://www.w3.org/ns/solid/terms#')
+
+const cache = new LRU({
```
rubensworks: I see the benefit of caching, but this could make this quite annoying for users.

It may be possible that the user's quota is at 100%, but that the view does not resemble this yet. We may be getting bug reports because of this.

So either we should remove this cache, or we should add a cache invalidation step every time a file system change happens in the user's directories.

lib/requests/get-account-info-request.js
```diff
@@ -0,0 +1,115 @@
...
+    if (body.status && body.status >= 400) {
+      return res.status(body.status).send(body.message)
+    }
+    return res.status(200).set('Content-Type', contentType).send(body)
+  }
+}
+
+async function getBody (req, contentType) {
+  const userId = req.session.userId
+  const cacheKey = `${userId}-${contentType}`
```
rubensworks: I would not include `contentType` in the cache key.
Instead, we could just cache the graph, and re-serialize the graph for every request.
This would solve the problem where the whole quota calculation needs to be re-done if someone asks for a different content type.
Also, this would make cache invalidation a bit easier.

lib/utils.js
```diff
@@ -233,11 +234,11 @@ async function overQuota (root, serverUri) {
  * so it needs to be rewritten.
  */
 
-function actualSize (root) {
+async function actualSize (root) {
```
rubensworks: I don't think this `async` is needed here, as `promisify` should already return a promise.

lib/utils.js
```diff
@@ -233,11 +234,11 @@ async function overQuota (root, serverUri) {
...
   return util.promisify(getSize)(root)
 }
 
-function _asyncReadfile (filename) {
+async function _asyncReadfile (filename) {
```
rubensworks: I don't think this `async` is needed here, as `promisify` should already return a promise.

megoth:
> I'm just wondering what happens if the quota system is disabled, will the user also get a 404 on `/account`?
> I assume that we want all quota-related things to be hidden in that case?

If the quota system is disabled, i.e. serverSide.ttl does not have a triple with `solid:storageQuota`, it will return a 404.

megoth:
I'm unsure of the caching-thing myself, as @rubensworks mentions. It could be an annoyance for people uploading a lot of files in a short timespan, and not see the updated info.

Right now I've cached the data for one hour, but we could make it less, like a minute. But then there might not be the big need for this cache anyway.

Another way of doing this could be to have the cache as a service that size-altering commands (ie POST, PUT, PATCH, DELETE) could call whenever something is changed on the server; it could simply remove the cached value for the userId, which would regenerate the info on next call. I'm worried that this is a bit overcomplicating it though =/

rubensworks:
> premature optimization is the root of all evil

If we're not sure about the performance impact, I would just remove the cache, and look for solutions if we experience a performance issue.

megoth:
> > premature optimization is the root of all evil
> 
> If we're not sure about the performance impact, I would just remove the cache, and look for solutions if we experience a performance issue.

I agree wholeheartedly to this; I'll remove the caching mechanism for now.

lib/utils.js
megoth: Fixed in b79bd563d3c582b6babb7018149a55e45e53ebb6

lib/utils.js
megoth: Fixed in b79bd563d3c582b6babb7018149a55e45e53ebb6

common/js/auth-buttons.js
megoth: Fixed in b79bd563d3c582b6babb7018149a55e45e53ebb6

megoth:
> > ```
> > 4. [Line 74](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R74): There wasn't a triple denoting where storage is located --> Return 404
> > ```
> 
> Right. Indeed an error. Actually, I wonder if that should result in a `500`, you could argue that a WebID without an associated storage is a server misconfig... OTOH, it is a [SHOULD](https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#storage-discovery) for that link. Still, I think it makes sense to make that a server side error since it is no fault of the user. We still need to tell the user why, that no storage has been configured.

I agree it is very strange, but as the wording is SHOULD, we have to expect cases where people have removed `pim:storage` from their WebID (maybe it's they don't want anyone to know where their storage is).

One of the problems is also whether or not we should return a 200 or 404 (or maybe 500 in some cases). The problem is that we have no way of knowing whether the server enforces a quota-system, expect to infer whether serverSide.ttl and `solid:storageQuota` is present or not. If that was intentional, return 200 with an empty graph, if it was not, return 404.

Maybe we should have it in config.json whether or not POD providers want to enforce a quota-system?

lib/requests/get-account-info-request.js
megoth: Removed cache in 86a33e536aa8398cc4cd8b8aad650e26e0b67a33

lib/acl-checker.js
```diff
@@ -144,6 +146,49 @@ class ACLChecker {
...
+  static createFromLDPAndRequest (resource, ldp, req) {
```
megoth: This could be moved as a method to `ldp.js` instead, something like `createACLFromRequest(resource, req)` if that's a better abstraction? @kjetilk @rubensworks

kjetilk:
> > > ```
> > > 4. [Line 74](https://github.com/solid/node-solid-server/pull/1060/files#diff-0c7366a5df447f793ba324d58a9a7731R74): There wasn't a triple denoting where storage is located --> Return 404
> > > ```
> > 
> > 
> > Right. Indeed an error. Actually, I wonder if that should result in a `500`, you could argue that a WebID without an associated storage is a server misconfig... OTOH, it is a [SHOULD](https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#storage-discovery) for that link. Still, I think it makes sense to make that a server side error since it is no fault of the user. We still need to tell the user why, that no storage has been configured.
> 
> I agree it is very strange, but as the wording is SHOULD, we have to expect cases where people have removed `pim:storage` from their WebID (maybe it's they don't want anyone to know where their storage is).

Right, but in that case, we either have to require `pim:storage` to be present if quota system is in use, or not rely on it. We have to make sure that people can't evade the quota system just by removing their `pim:storage` relation. Perhaps we should set it in `serverSide.ttl`?

> One of the problems is also whether or not we should return a 200 or 404 (or maybe 500 in some cases). The problem is that we have no way of knowing whether the server enforces a quota-system, expect to infer whether serverSide.ttl and `solid:storageQuota` is present or not. If that was intentional, return 200 with an empty graph, if it was not, return 404.
> 
> Maybe we should have it in config.json whether or not POD providers want to enforce a quota-system?

Yeah, we could... I just didn't want to spend so much time on this stop-gap measure, but I suppose we might have to.

Regarding caching, Tim was (rightly) worried about the write performance, and caching in RAM should help, but indeed, just a max-age wouldn't do the trick, the cache would need to be invalidated on write. More work, and it isn't strictly necessary for small Pods, so I can live with not having it in.

lib/acl-checker.js
rubensworks: I think it's fine to keep it here, but not strong opinion.

megoth:
> > One of the problems is also whether or not we should return a 200 or 404 (or maybe 500 in some cases). The problem is that we have no way of knowing whether the server enforces a quota-system, expect to infer whether serverSide.ttl and `solid:storageQuota` is present or not. If that was intentional, return 200 with an empty graph, if it was not, return 404.
> > Maybe we should have it in config.json whether or not POD providers want to enforce a quota-system?
> 
> Yeah, we could... I just didn't want to spend so much time on this stop-gap measure, but I suppose we might have to.

Adding an option to the init-script is an easy and quick thing to do. Not necessarily smart though, so am open to being convinced to not do it. I'll add it as another branch though, so that we can discuss it separately from this issue.

(Just to add what I'm thinking of doing: I'll add an option called "enforceQuota" or something, that is only prompted if the user is setting up a multiuser-server, and set it to false by default. I guess we should expand upon it as well, i.e. copy serverSide.ttl as serverSide.ttl.inactive or serverSide.ttl. I should have something in place by the end of day latest.)

megoth:
I'll close this until we have enforceQuota in place. I'll reopen the PR then.

Issue #1063, opened 2019-01-24
Regression bug in v5: Need to check for a file's ACL when it is served as resource for a container


megoth:
When a file is served as response to dereferencing a container, we also need to check that file's ACL settings when determining whether a user has access to that resource.

How it works on v4: as a visitor I have read access to both https://megoth.solid.community/ and https://megoth.solid.community/index.html
But on v5: as a visitor I have read access to https://megoth-test1.dev.inrupt.net/index.html, but not to https://megoth-test1.dev.inrupt.net/

kjetilk:
So, thinking about it, I'm thinking that we should be using `/.acl` even if the server serves up `/index.html`.

The default ACL file says `/` is public, so could you double check that your `/.acl` says that?

megoth:
Sorry, forgot to answer this.

The template for root ACL is

```
# Root ACL resource for the user account
@prefix acl: <http://www.w3.org/ns/auth/acl#>.

<#owner>
    a acl:Authorization;

    acl:agent <{{webId}}> ;

    # Optional owner email, to be used for account recovery:
    {{#if email}}acl:agent <mailto:{{{email}}}>;{{/if}}

    # Set the access to the root storage folder itself
    acl:accessTo </>;

    # All resources will inherit this authorization, by default
    acl:default </>;

    # The owner has all of the access modes allowed
    acl:mode
        acl:Read, acl:Write, acl:Control.

# Data is private by default; no other agents get access unless specifically
# authorized in other .acls
```

So the default ACL file for root says that all files are private.

megoth:
So making the root folder publicly available is a solution for this, that is changing `default_templates/new-account/.acl` to:

```
# Root ACL resource for the user account
@prefix acl: <http://www.w3.org/ns/auth/acl#>.
@prefix foaf: <http://xmlns.com/foaf/0.1/>.

<#owner>
    a acl:Authorization;

    acl:agent <{{webId}}> ;

    # Optional owner email, to be used for account recovery:
    {{#if email}}acl:agent <mailto:{{{email}}}>;{{/if}}

    # Set the access to the root storage folder itself
    acl:accessTo </>;

    # All resources will inherit this authorization, by default
    acl:default </>;

    # The owner has all of the access modes allowed
    acl:mode
        acl:Read, acl:Write, acl:Control.

# Data is private by default; no other agents get access unless specifically
# authorized in other .acls

# The public has read permissions to root folder
<#public>
    a acl:Authorization;
[...]

PR #1065, opened 2019-01-25
Default pod template sets rights on / instead of index.html

RubenVerborgh:
Partial fix for #1063.


default-templates/new-account/.acl
```diff
@@ -1,23 +1,25 @@
 # Root ACL resource for the user account
 @prefix acl: <http://www.w3.org/ns/auth/acl#>.
 
-<#owner>
+# The homepage is readable by the public
+<#public>
     a acl:Authorization;
+    acl:agentClass foaf:Agent;
+    acl:accessTo </>;
```
megoth: Should we state that `index.html` is publicly readable as well, since we're deleting `index.html.acl`? 

If we don't, people won't be able to access `/index.html`, but they would be able to access `/`. That might be a use case that's not that important anyway though, so might not be necessary.

megoth: Hmm, was planning to say we could change line 8 to `acl:accessTo </>, </index.html>`, but that doesn't work either...

RubenVerborgh: I would refrain from including anything `/index/` in the ACL files. As far as the user is concerned, these don't really (need to) exist. They are a convention; we could start naming them _default.html_ in the future.

RubenVerborgh: It's part of the broader point in https://github.com/solid/web-access-control-spec/issues/36.

kjetilk:
So, to add to the confusion, I just did 
```
kjetil@robin:~$ dog -k https://kjetiltest.inrupt.net/
HTTP/1.1 401 Unauthorized
```
(where `dog` is an alias analogous to `cat` but with network :-) )

So, it appears to me that the root is not public readable by default in the current code...?

RubenVerborgh:
> So, it appears to me that the root is not public readable by default in the current code...?

Indeed, this PR should fix that for new pods.

Also, you should really publish your dotfiles on GitHub, so we can all use those funny aliaes 😉

kjetilk:
> > So, it appears to me that the root is not public readable by default in the current code...?
> 
> Indeed, this PR should fix that for new pods.

What I mean is that it is not public readable for 4.4... Which would imply no regression. But I'm probably just confused.

> Also, you should really publish your dotfiles on GitHub, so we can all use those funny aliases wink

Hehe, there's not a lot. That is, I have a `/site/` dir that are mounted or rsynced to all my boxes. Some stuff in there, but the aliases I'm actually using are these:

```bash
alias ..='cd ../'
alias ...='cd ../../'
alias ....='cd ../../../'
alias .....='cd ../../../../'
alias deluntrack='for f in $(git st --porcelain | grep ?? | cut -b4-) ; do rm -r $f; done'
alias dir='ls -al | less -X -E'
alias dog='curl -D - '
alias lr='ls -ltr'
```

Perhaps one of these days, I might put them somewhere. :-) Meanwhile, apologies for the OT

RubenVerborgh:
No regression: this is a consequence of acl-check, which isn’t in 4.4,
right?

leinue:
I encountered this problem after upgrading NSS5, looking forward to solving.

kjetilk:
> No regression: this is a consequence of acl-check, which isn’t in 4.4, right?

It wouldn't be a direct consequence of acl-check, it might be a consequence of rewriting the tree traversal code, or the introduction of ResourceMapper, but, the initial observation that @megoth made was that:

> How it works on v4: as a visitor I have read access to both https://megoth.solid.community/ and https://megoth.solid.community/index.html

This is what I fail to reproduce, both on inrupt.net and solid.community, both of which runs 4.x series server. 

I haven't got read access to https://kjetiltest.solid.community/ nor to https://kjetiltest.inrupt.net/ so with these newly created accounts, the behaviour of 4.x is identical to 5.x-beta.5, as far as I can tell.

Now, if that makes sense is another matter, I tend to agree that `/` (but only that, i.e. no `acl:default` predicate) should be readable to public by default, but there may be design decisions behind that it isn't.

So, there are two questions:

1. If @megoth 's observed behaviour is representative, or a result of having changed the ACL.
2. If we should make `/` public readable by default.

But I may just be confused.

PR #1067, opened 2019-01-26
Added special ACL handling for root

megoth:
This is a hack that should fix https://github.com/solid/node-solid-server/issues/1063.

In order to not break old PODs that have a root ACL that doesn't give READ access to root, we want to also check the ACL for the representation that the server intends to return for root.

I'm uncertain of the tests I changed in `test/integration/authentication-oidc-test.js`, so review those with extra scrutiny.

I've rewritten the creation of ACLChecker into createFromLDPAndRequest, the same as I did in https://github.com/solid/node-solid-server/pull/1060. This might not be a very good solution, but it works, and I propose we rewrite it as a separate PR some time in the future when we have thought of a better pattern.

### DIFF
--- a/lib/acl-checker.js
+++ b/lib/acl-checker.js
@@ -113,9 +113,8 @@ class ACLChecker {
     if (!returnAcl) {
       throw new HTTPError(500, `No ACL found for ${resource}, searched in \n- ${acls.join('\n- ')}`)
     }
-    const groupUrls = returnAcl.graph
-    .statementsMatching(null, ACL('agentGroup'), null)
-    .map(node => node.object.value.split('#')[0])
+    const groupNodes = returnAcl.graph.statementsMatching(null, ACL('agentGroup'), null)
+    const groupUrls = groupNodes.map(node => node.object.value.split('#')[0])
     await Promise.all(groupUrls.map(groupUrl => {
       this.requests[groupUrl] = this.requests[groupUrl] || this.fetch(groupUrl, returnAcl.graph)
       return this.requests[groupUrl]
--- a/lib/acl-checker.js
+++ b/lib/acl-checker.js
@@ -114,8 +114,8 @@ class ACLChecker {
       throw new HTTPError(500, `No ACL found for ${resource}, searched in \n- ${acls.join('\n- ')}`)
     }
     const groupUrls = returnAcl.graph
-      .statementsMatching(null, ACL('agentGroup'), null)
-      .map(node => node.object.value.split('#')[0])
+    .statementsMatching(null, ACL('agentGroup'), null)
+    .map(node => node.object.value.split('#')[0])
     await Promise.all(groupUrls.map(groupUrl => {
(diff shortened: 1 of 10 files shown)


RubenVerborgh:
> I think that the ACL of the representation shouldn't be used for the request URL

We agree, but have you seen https://github.com/solid/node-solid-server/issues/1063#issuecomment-457649464?

kjetilk:
@RubenVerborgh Yeah, but I commented on that too, please see my last comment there, where I say that I just want to accomodate existing PODs by supplying a script, not to make any changes to accomodate for it in the code. So I guess we disagree based on that comment. :-) Lets discuss over in #1063

kjetilk:
I'm OK with it now, but I have to admit that I find it harder to review when it contains that extensive refactorings. Perhaps we should do refactoring in separate PRs in the future?

Nice if we can get another review too.

megoth:
> I'm OK with it now, but I have to admit that I find it harder to review when it contains that extensive refactorings. Perhaps we should do refactoring in separate PRs in the future?
> 
> Nice if we can get another review too.

I could copy a lot of code instead of refactoring it, yes, but please lets do a separate refactoring at once after, ok? If we don't, we'll get (even more) unmaintainable code...

kjetilk:
> I could copy a lot of code instead of refactoring it, yes, but please lets do a separate refactoring at once after, ok? If we don't, we'll get (even more) unmaintainable code...

No, I like the fact that you're refactoring, we should do that. I just wonder if we could do it differently. Perhaps there is something around that could tell us some practices around that.

What if you first branch, refactor the stuff you need to refactor, then branch off of that branch again to implement the new functionality. PRs could be made for both, so people can review them apart or together. Or perhaps that would add too much overhead?

We should discuss at some point.

megoth:
> > I could copy a lot of code instead of refactoring it, yes, but please lets do a separate refactoring at once after, ok? If we don't, we'll get (even more) unmaintainable code...
> 
> No, I like the fact that you're refactoring, we should do that. I just wonder if we could do it differently. Perhaps there is something around that could tell us some practices around that.
> 
> What if you first branch, refactor the stuff you need to refactor, then branch off of that branch again to implement the new functionality. PRs could be made for both, so people can review them apart or together. Or perhaps that would add too much overhead?
> 
> We should discuss at some point.

Ah, so have a separate branch with the refactorings... yeah, I could do that. Not something I have done  in other projects, but than again, not so much work regarding branches and stuff before either =P

I'll do a bit of research, and see what good practices are out there.

megoth:
> I'll do a bit of research, and see what good practices are out there.

I don't have that much time to do research, but I propose using the pattern `refactor/*`-pattern for this kinds of changes. (I know it's not really a refactor in this case, and I'm adding to the widely used trend of abusing the word refactor; but I hope everyone understands the intention, and am open to to better named pattern.) I've added a PR using that pattern on https://github.com/solid/node-solid-server/pull/1075.

Issue #1069, opened 2019-01-27
CORS proxy doesn't follow redirects


csarven:
When the proxy receives a redirect from the target server, it doesn't follow, and instead returns it back to the requesting agent.

PR #1070, opened 2019-01-27
Make CORS proxy follow redirects

csarven:
Implements https://github.com/solid/node-solid-server/issues/1069 .

http-proxy-middleware has an option to `followRedirects`. http-proxy has max redirects set to 10.

I've also set the `proxyTimeout` to 5 seconds (arbitrary but reasonable I guess).

### DIFF
--- a/lib/handlers/cors-proxy.js
+++ b/lib/handlers/cors-proxy.js
@@ -19,6 +19,8 @@ const PROXY_SETTINGS = {
   target: 'dynamic',
   logLevel: 'silent',
   changeOrigin: true,
+  followRedirects: true,
+  proxyTimeout: 5000,
   router: req => req.destination.target,
   pathRewrite: (path, req) => req.destination.path
 }


csarven:
This is kinda outside of my knowledgebase :) The fail seems doesn't seem to be related to what's in this PR'd/changed, no? Requires package-lock update? :S

kjetilk:
Uhm, dunno either... But it was in the proxy section. I'll put it on the project board and see what we can do about it.

kjetilk:
I have cloned your branch now, @csarven , updated package-lock and all, but those tests still fail. I'm not sure what to make of it, it is in code that I haven't worked with. @RubenVerborgh , can you have a look at it and see what might be the cause of breakage?

kjetilk:
I just have to admit I don't know this part of the code well enough to be able to merge this...

csarven:
@kjetilk Is there anything in particular that I can help to clarify? or specific concerns relating to this?

I've left the changes out of auth-proxy.js as @RubenVerborgh suggested (marked it as a separate issue), so there is no issue with the tests for this PR.

Aside: There are other issues related to the proxy, but I'll try to address them after this one.

kjetilk:
Uhm, sorry, I didn't see that the tests had passed.

Issue #1031, opened 2018-12-17
API-endpoints should only be served via serverUri


megoth:
Today the server doesn't handle [domain]/register and [subdomain]/register any different. We should make sure that only the former works.

RubenVerborgh:
Yes, with the exception of auth, so people can fill out their own domain.

PR #1073, opened 2019-01-28
Restricting access to some HTTP API endpoints to top domain

megoth:
This is to prevent attempts at spoofing such as `https://[subdomain]/register`.


megoth:
This is an attempt to fix https://github.com/solid/node-solid-server/issues/1031

megoth:
Ok, have laxed a bit on some resources that I think should be accessible via subdomain. These are `/login`, `/logout`, and `/api/oidc/rp/:issuer_id` (and various alternatives). I'm uncertain about the latter one, so please confirm if my assumption is correct.

megoth:
Capability discovery should be working as expected now, if I have understood it correctly.

Maybe @RubenVerborgh, @dmitrizagidulin and @kjetilk can give this a new look?

PR #1077, opened 2019-01-30
Add an origin that can be trusted by default

kjetilk:
To ease deployment of apps, we add an origin that can be trusted by default, but now controlled by the Solid project.

It can be overridden in the server config.


kjetilk:
See #1090 for follow-up.

Issue #1080, opened 2019-01-31
Validate RDF files


kjetilk:
We should implement @elf-pavlik 's suggestion in #1079 to use the https://github.com/IDLabResearch/TurtleValidator to validate all `.ttl` and `.acl` files on Travis. 

Perhaps something you could look into, @joachimvh ?

joachimvh:
I can have a look

PR #1083, opened 2019-02-04
Validate template turtle files

joachimvh:
This validates all turtle files in the default-templates folder when testing on travis, as requested in #1080 .


kjetilk:
Thanks a lot!

Issue #1085, opened 2019-02-05
PATCH requests with non-matching deletions should succeed


RubenVerborgh:
We currently have evidence (https://github.com/Inrupt-inc/generator-solid-react/pull/41#discussion_r254013484) that a patch with a DELETE block that has no matches will cause a 409. Is this correct? Should there at least be one match with a DELETE?

kjetilk:
Hmmm,  no, `409` doesn't make sense to me either. That sounds like a `404`.

rubensworks:
From the SPARQL UPDATE spec:

> 3.1.3.3 DELETE WHERE
> Analogous to DELETE/INSERT, deleting triples that are not present, or from a graph that is not present will have no effect and will result in success.

So DELETE's without matches should return a 200 it seems. So this indeed looks like a bug in NSS.

timbl:
# Note: 409 Conflict
--
  | <p>A SPARQL update message often contains both a DELETE and then an INSERT.
  | This may be used to update a field from one value to another.
  | When more than one application or user is using the same data,
  | there may arise times when the DELETE fails
  | because another user has already deleted the same data.
  | In this case it <strong>very important</strong>
  | that the delete does not fail silently.
  | The HTTP server MUST return error status 409.
  | ("409 Conflict" indicates that the request could not be processed because of conflict in the request, such as an edit conflict).
  | The client can then for example inform the user by backing out the
  | change the user was trying to make, or it can retry a reservation later.
  | The atomicity of the DELETE,INSERT function can be used to provide
  | various mutual exclusion systems, such as reserving a resource
  | or generating unique sequential numbers, and so on.
  

- https://www.w3.org/DesignIssues/ReadWriteLinkedData.html

RubenVerborgh:
@timbl I agree, but unfortunately that is in conflict with the [SPARQL 1.1 UPDATE W3C Recommendation]:

> Note that the deletion of non-existing triples has no effect, i.e., triples in the QuadData that did not exist in the Graph Store are ignored.

That's why I created https://github.com/linkeddata/rdflib.js/pull/299, could you have a look at that one?

timbl:
Grumble.   a system has to have semaphores
https://en.wikipedia.org/wiki/Semaphore_(programming)
https://en.wikipedia.org/wiki/Semaphore_(programming)

kjetilk:
[...]

PR #1086, opened 2019-02-06
Fix PATCH requests with non-matching deletions

RubenVerborgh:
Fixes https://github.com/solid/node-solid-server/issues/1085

Depends on https://github.com/linkeddata/rdflib.js/pull/298


kjetilk:
Ah, got it, it relies on changes in rdflib. Indeed, then approving, since the changes here LGTM.

RubenVerborgh:
Discussion continued in https://github.com/linkeddata/rdflib.js/pull/299

PR #1092, opened 2019-02-07
Report server version in X-Powered-By header

kjetilk:
I propose we report the server version in the `X-Powered-By` header. When users report errors, we will sometimes need to stop to ask them to verify the version.

Now, some say it has security implications, since if people aren't upgrading when facing security problems, a spider can map vulnerable servers. However, we are now at a prototype, and we are likely to need to help users, and so it should not be as much of a concern.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -25,7 +25,7 @@ const path = require('path')
 const { routeResolvedFile } = require('./utils')
 const ResourceMapper = require('./resource-mapper')
 const aclCheck = require('@solid/acl-check')
-const version = require('../package.json').version
+const { version } = require('../package.json')
 
 const corsSettings = cors({
   methods: [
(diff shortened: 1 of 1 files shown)


lib/create-app.js
```diff
@@ -25,6 +25,7 @@ const path = require('path')
 const { routeResolvedFile } = require('./utils')
 const ResourceMapper = require('./resource-mapper')
 const aclCheck = require('@solid/acl-check')
+const version = require('../package.json').version
```
RubenVerborgh: You might want to `const { version } = require('../package.json')`.

kjetilk: OK, done!

kjetilk:
Or could/should we use `${argv.version}` like in `start.js`?

RubenVerborgh:
> Or could/should we use `${argv.version}` like in `start.js`?

No that `argv` stuff is not healthy.

Issue #1104, opened 2019-02-18
/index.html.acl present, /index.html missing: requesting / returns /index.html.acl


csarven:
Issue occurs in v5.0.0-beta.7-6-g217e8e3 .

If index.html is absent and index.html.acl is present, request to / returns the contents of /index.html.acl . *Expected*: return container

If /index.html.acl is also absent, returns container.

How to reproduce:
Fresh install. (Added public read policy on /.acl to simplify)

```
$ bin/solid start
$ mv data/index.html data/foo.html
$ curl -ikL https://localhost:8443/
HTTP/1.1 200 OK
X-Powered-By: solid-server/5.0.0-beta.7
Vary: Accept, Authorization, Origin
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate
Allow: OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE
Link: <.acl>; rel="acl", <.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
WAC-Allow: user="read",public="read"
MS-Author-Via: SPARQL
Updates-Via: wss://localhost:8443
Content-Type: text/turtle
Date: Mon, 18 Feb 2019 10:20:52 GMT
Connection: keep-alive
Transfer-Encoding: chunked

@prefix acl: <http://www.w3.org/ns/auth/acl#>.
@prefix foaf: <http://xmlns.com/foaf/0.1/>.

<#public>
    a acl:Authorization;

    acl:agentClass foaf:Agent;  # everyone

    acl:accessTo <./index.html>;

    acl:mode acl:Read.
```

kjetilk:
Oh, OK. So, I suppose something went wrong with the #1067 fixing? Could you check, @megoth ?

megoth:
Yes, this sounds like an unexpected outcome of the special handling root...

I'll take a look!

megoth:
So after a bit of research it seems that the problem is in the resourceMapper that picks `index.html.acl` because it tries to find a file that matches with `index`. So nothing to do with #1067. I'll push a PR with a fix soon.

PR #1109, opened 2019-02-19
Removes acl files from list of files to match when mapping url to file

megoth:
Should fix https://github.com/solid/node-solid-server/issues/1104

### DIFF
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -15,7 +15,8 @@ class ResourceMapper {
     includeHost = false,
     defaultContentType = 'application/octet-stream',
     indexFilename = 'index',
-    overrideTypes = { acl: 'text/turtle', meta: 'text/turtle' }
+    overrideTypes = { acl: 'text/turtle', meta: 'text/turtle' },
+    fileSuffixes = { acl: '.acl', meta: '.meta' }
   }) {
     this._rootUrl = this._removeTrailingSlash(rootUrl)
     this._rootPath = this._removeTrailingSlash(rootPath)
@@ -24,6 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
+    this._suffixes = fileSuffixes
 
     // If the host needs to be replaced on every call, pre-split the root URL
     if (includeHost) {
@@ -93,7 +95,7 @@ class ResourceMapper {
 
     // Find a file with the same name (minus the dollar extension)
     return (files.find(f => this._removeDollarExtension(f) === filename ||
-      (isIndex && f.startsWith(this._indexFilename + '.') && !f.endsWith('.acl'))))
+      (isIndex && f.startsWith(this._indexFilename + '.') && !f.endsWith(this._suffixes.acl))))
   }
 
   async getRepresentationUrlForResource (resourceUrl) {
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -25,7 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
-    this._suffixes = fileSuffixes
+    this._isControlFile = new RegExp(`.(?:${fileSuffixes.join('|')})$`)
 
     // If the host needs to be replaced on every call, pre-split the root URL
     if (includeHost) {
@@ -92,10 +92,13 @@ class ResourceMapper {
 
   async _getMatchingFile (folder, filename, isIndex) {
     const files = await this._readdir(folder)
-
-    // Find a file with the same name (minus the dollar extension)
-    return (files.find(f => this._removeDollarExtension(f) === filename ||
-      (isIndex && f.startsWith(this._indexFilename + '.') && !this._suffixes.some(suffix => f.endsWith(suffix)))))
+    const hasSameName = (f) => {
+      return this._removeDollarExtension(f) === filename
+    }
+    const isIndexFile = (f) => {
+      return isIndex && f.startsWith(this._indexFilename + '.') && !this._isControlFile.test(f)
+    }
+    return files.find(f => hasSameName(f) || isIndexFile(f))
   }
 
   async getRepresentationUrlForResource (resourceUrl) {
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -25,7 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
-    this._isControlFile = new RegExp(`.(?:${fileSuffixes.join('|')})$`)
+    this._isControlFile = new RegExp(`(?:${fileSuffixes.map(fs => fs.replace('.', '\\.')).join('|')})$`)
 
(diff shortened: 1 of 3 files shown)


megoth:
I'll push an update soon that doesn't hardcode it ^_^

megoth:
@rubensworks @kjetilk I've pushed an update, hope it looks better ^_^ (Was uncertain how to include the knowledge of suffixes into resourceMapper, so hope the way I did it looks ok)

lib/resource-mapper.js
```diff
@@ -90,6 +90,14 @@ class ResourceMapper {
     return { path, contentType: contentType || this._defaultContentType }
   }
 
+  async _getMatchingFile (folder, filename, isIndex) {
+    const files = await this._readdir(folder)
+
+    // Find a file with the same name (minus the dollar extension)
+    return (files.find(f => this._removeDollarExtension(f) === filename ||
+      (isIndex && f.startsWith(this._indexFilename + '.') && !f.endsWith(this._suffixes.acl))))
```
rubensworks: I would go one step further, and loop over all `this._suffixes` to exclude _all_ file names that end with any of the defined suffixes. Right now, only `index.html.acl` will be blacklisted, but `index.html.meta` will still cause the same problem. A loop over `this._suffixes` would solve this.

For this purpose, it may be easier to refactor `this._suffixes` to be an array instead of a hash, as this is (AFAICS) only being used here.

megoth:
@rubensworks then I've implemented your suggested changes (changing fileSuffixes from an object to an array). Could you have a look at the latest change, and see if you think it's good?

lib/resource-mapper.js
```diff
@@ -15,7 +15,8 @@ class ResourceMapper {
     includeHost = false,
     defaultContentType = 'application/octet-stream',
     indexFilename = 'index',
-    overrideTypes = { acl: 'text/turtle', meta: 'text/turtle' }
+    overrideTypes = { acl: 'text/turtle', meta: 'text/turtle' },
+    fileSuffixes = ['.acl', '.meta']
```
RubenVerborgh: Unclear variable name. I would name it something like `controlExtensions`.

lib/resource-mapper.js
```diff
@@ -24,6 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
+    this._suffixes = fileSuffixes
```
RubenVerborgh: Idem. They're not just suffixes, they are suffixes for a specific kinds of files that needs special behavior.

lib/resource-mapper.js
```diff
@@ -68,11 +70,9 @@ class ResourceMapper {
       // Read all files in the corresponding folder
```
RubenVerborgh: Outdated comment

lib/resource-mapper.js
```diff
@@ -68,11 +70,9 @@ class ResourceMapper {
...
       const filename = fullPath.substr(fullPath.lastIndexOf('/') + 1)
       const folder = fullPath.substr(0, fullPath.length - filename.length)
-      const files = await this._readdir(folder)
 
       // Find a file with the same name (minus the dollar extension)
-      let match = !searchIndex ? '' : (files.find(f => this._removeDollarExtension(f) === filename ||
-        (isIndex && f.startsWith(this._indexFilename + '.'))))
+      let match = searchIndex ? await this._getMatchingFile(folder, filename, isIndex) : ''
```
RubenVerborgh: I'm not sure if I would move this specific bit to a new function. Is there potential for reuse?

lib/resource-mapper.js
```diff
@@ -90,6 +90,14 @@ class ResourceMapper {
     return { path, contentType: contentType || this._defaultContentType }
   }
 
+  async _getMatchingFile (folder, filename, isIndex) {
+    const files = await this._readdir(folder)
+
+    // Find a file with the same name (minus the dollar extension)
+    return (files.find(f => this._removeDollarExtension(f) === filename ||
```
RubenVerborgh: We should probably expand this condition across multiple lines now, given the increased complexity.

lib/resource-mapper.js
```diff
@@ -90,6 +90,14 @@ class ResourceMapper {
...
+      (isIndex && f.startsWith(this._indexFilename + '.') && !this._suffixes.some(suffix => f.endsWith(suffix)))))
```
RubenVerborgh: Seems wasteful to always iterate over the extensions.

How about, in the constructor, we instead use a regex to determine whether something is a control file? Then the default value is `/.(?:acl|meta)$/`, and this routine just does `this._controlFile.test(f)`.

megoth:
Ah, yes, I've been lazy on tests here. I'll write them when I have time, hopefully during the next couple of days.

lib/resource-mapper.js
megoth: That sounds reasonable - I'll write tests that make sure stuff works, and then try to optimize if per your suggestions 😸

lib/resource-mapper.js
megoth: My thinking was to make it more readable by giving it a name. Don't know of any potential reuse, no.

megoth:
@RubenVerborgh think I have addressed your concerns; I'm uncertain of how to run coverage reports on this actually, could you check? (And perhaps drop a line on how you do it here? Could read up more on it, but am in bit of a hurry right now =P )

lib/resource-mapper.js
```diff
@@ -24,6 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
+    this._isControlFile = new RegExp(`.(?:${fileSuffixes.join('|')})$`)
```
RubenVerborgh: Not the right regex: a) fileSuffixes already has a leading dot, so there are two dots now b) dots need to be escaped in regexes (but, since you are creating the regex from a string, actually needs double backslash '`\\.`).

RubenVerborgh: Alternatively, you can just pass the regex itself instead of passing fileSuffixes.

megoth: That makes it difficult to configure the suffixes though =/

megoth: Hope the latest commit looks good.

Ryuno-Ki:
If you're concerned about coverage... why not use a tool like [CodeCov](https://codecov.io/)?
You could make it even part of the conditions to flag a PR as mergeable.

lib/resource-mapper.js
```diff
@@ -24,6 +25,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._indexFilename = indexFilename
     this._types = { ...types, ...overrideTypes }
+    this._isControlFile = new RegExp(`(?:${fileSuffixes.map(fs => fs.replace('.', '\\.')).join('|')})$`)
```
RubenVerborgh: `fs.replace(/\./g, '\\.')` strictly, but good enough 😉

RubenVerborgh:
> If you're concerned about coverage... why not use a tool like [CodeCov](https://codecov.io/)?

Coverage in general is not great yet; but it was for this particular class 😎

lib/resource-mapper.js
megoth: Regex is so difficult =P

PR #1111, opened 2019-02-20
Display server version at startup when not running from git repository

gklyne:
See #1110 

This fix raises an error when the version information returned by running git is empty, thereby falling back to using package.json.

(I'm not very conversant with PR mechanics, so apologies if the form is wrong here.)

Issue #1110, opened 2019-02-20
Server version not displayed at startup when not running in git repository


gklyne:
(I imagine this could make life easier when processing bug reports)

csarven:
I see eg:

```
$ bin/solid start
Solid server (v5.0.0-beta.7-6-g217e8e3) running on https://localhost:8443/
Press <ctrl>+c to stop
```

Did you mean something else?

gklyne:
Hmm: I'm not seeing that.  I'm using solid-test.

I'm also not seeing a version of the front page (https://localhost:8443/)

Here's what I see:

```
nerthus:solid graham$ DEBUG=solid:* ./node_modules/solid-server/bin/solid-test start
  solid:settings Server URI: https://localhost:8443 +0ms
  solid:settings Auth method: oidc +1ms
  solid:settings Strict origins: true +0ms
  solid:settings Allowed origins: https://apps.solid.invalid +1ms
  solid:settings Db path: ./.db +0ms
  solid:settings Config path: ./config +0ms
  solid:settings Suffix Acl: .acl +0ms
  solid:settings Suffix Meta: .meta +0ms
  solid:settings Filesystem Root: /Users/graham/solid/data +0ms
  solid:settings Allow WebID authentication: true +0ms
  solid:settings Live-updates: true +0ms
  solid:settings Multi-user: false +0ms
  solid:settings Suppress default data browser app: undefined +0ms
  solid:settings Default data browser app file path: default +0ms
  solid:settings Base URL (--mount): / +19ms
  solid:settings SSL Private Key path: ../solid-certs/localhost.key +0ms
  solid:settings SSL Certificate path: ../solid-certs/localhost.crt +0ms
Solid server () running on https://localhost:8443/
Press <ctrl>+c to stop
  solid:authentication Provider keys loaded from config +23ms
  solid:authentication Provider keychain initialized +171ms
  solid:authentication Not sleeping before client registration... +1ms
  solid:authentication Client fetched for issuer https://localhost:8443 +7ms
  solid:authentication Local RP client initialized +0ms
```
I tried some variations too, still no version.  I see the parentheses there, but are empty

kjetilk:
That's pretty weird. You have the statement:
```
Solid server () running on https://localhost:8443/
```
but the `argv.version` (or whatever) is empty. I have never seen that, but I hear it is somewhat problematic to use. 

BTW, there is also #1092 .

gklyne:
My curl outpout shows just:

[...]

PR #1112, opened 2019-02-22
Display server version at startup when not running from git repository

gklyne:
See #1110, #1111 

This fix raises an error when the version information returned by running git is empty, thereby falling back to using package.json.  (This is a new attempt to submit previous PR #1111, but this time based upon and updating the release/v5.0.0 branch.)

### DIFF
--- a/bin/lib/cli.js
+++ b/bin/lib/cli.js
@@ -25,10 +25,15 @@ function getVersion () {
     // Obtain version from git
     const options = { cwd: __dirname, encoding: 'utf8' }
     const { stdout } = spawnSync('git', ['describe', '--tags'], options)
-    return stdout.trim()
+    const version = stdout.trim()
+    if (version === '') {
+      throw new Error('No git version here')
+    }
+    return version
   } catch (e) {
     // Obtain version from package.json
     const { version } = require(path.join(__dirname, '../../package.json'))
     return version
   }
 }
+


linonetwo:
Why it is showing `Solid server (v0.33.8)`?

I've installed latest v5.1.5

gklyne:
@linonetwo I'd check to see what `git describe --tags` is returning when run from the `bin/cli` directory.  (My fix _should_ only affect installs that are not from a local git repo.)

linonetwo:
I've install solid via `npm i -g`.

After install, it shows 5.1.5 installed, but then I run `solid -V` shows `Solid server (v0.33.8)`

kjetilk:
strange... I can't reproduce, and I don't know what could produce such a result. What does 
`which solid` say? Could it be something else entirely?

gklyne:
Is it possible that the npm install is inside some other git repository?

Running `git status`, or `git describe --tags`, in the directory with the installed software (i.e. the directory from `which solid`) might quickly eliminate this possibility.

linonetwo:
It is not a git repo:


```shell
~$ which solid
/home/labs/.nvm/versions/node/v12.4.0/bin/solid
~$ solid -V
v0.33.8
~$ cd /home/labs/.nvm/versions/node/v12.4.0/bin/solid && git status
-bash: cd: /home/labs/.nvm/versions/node/v12.4.0/bin/solid: Not a directory
~$ cd /home/labs/.nvm/versions/node/v12.4.0/bin && git status
Not currently on any branch.
nothing to commit, working directory clean
~/.nvm/versions/node/v12.4.0/bin$
```

gklyne:
Thanks - I agree, it's not that.

Next places I'd look are:

.../bin/lib/cli.js, line 33:

```
  } catch (e) {
    // Obtain version from package.json
    const { version } = require(path.join(__dirname, '../../package.json'))
    return version
  }
```

.../package.json, line 4:

```
  "version": "5.1.5",
```

Beyond that, as far as I can tell, the version display is handled by the `commander` command line parser.

I'd be tempted to try inserting a `console.log(version)` in `cli.js` before the `return version` to (a) confirm it's taking that code path, and (b) check the value of `version` that comes from package.json.  If it's correct the then it looks as if going astray in the handoff to commander at line 11: `program.version(getVersion())`

linonetwo:
Oh, note that, even if `git status` returns `Not currently on any branch.`,

`git describe --tags` can still return something:

```
xxx@xxx-ubuntu:~/.nvm/versions/node/v12.4.0/bin$ git describe --tags
v0.33.8
xxx@xxx-ubuntu:~/.nvm/versions/node/v12.4.0/bin$ git status
Not currently on any branch.
nothing to commit, working directory clean
```

Issue #1105, opened 2019-02-18
Prevent data  loss in concurrent updates - Using synchronous code


timbl:
The solid system requires that inconsistent changes from concurrent users editing the same thing should never  lose data, returning 409 to the change which failed.  (Semaphore operation).  Currently the server can fail to do this because the second request can come in between the first one reading the file and writing the modified file.  

This issue solves the problem by changing the patch handler so that the 
- target file read
- target file parse
- patch data 
- target file serialize
- target file write 

Are all synchronous, blocking the server from handling other requests in between

megoth:
I suggest closing this issue in favor of https://github.com/solid/node-solid-server/issues/1106.

timbl:
Agreed. Is there a way of expressing that to the system?

kjetilk:
Mentioning the bug number #1106 creates a cross-reference, I think that is the best we can do, so, hereby done. :-)

PR #1114, opened 2019-02-23
Patch uses synchronous code block for concurrent editing

timbl:
Responds to https://github.com/solid/node-solid-server/issues/1105


timbl:
Ooop sPR to wrong branch  .. ignore

PR #1116, opened 2019-02-23
Lock file during PATCH

### LINKED ISSUE #1106: Prevent data loss in concurrent updates - using file locking
The solid system requires that inconsistent changes from concurrent users editing the same thing should never lose data, returning 409 to the change which failed. (Semaphore operation). Currently the server can fail to do this because the second request can come in between the first one reading the file and writing the modified file.

This issue solves the problem by changing the patch handler so that the

* target file read
* target file parse
* patch data
* target file serialize
* target file write

Are all covered by a file lock , blocking the server from handling other requests in between, 
and ALSO the put handler's  file write.

RubenVerborgh:
Closes #1106.

This does not address other HTTP methods.

### DIFF
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -10,6 +10,7 @@ const $rdf = require('rdflib')
 const crypto = require('crypto')
 const overQuota = require('../utils').overQuota
 const getContentType = require('../utils').getContentType
+const { lock } = require('proper-lockfile')
 
 // Patch parsers by request body content type
 const PATCH_PARSERS = {
@@ -23,6 +24,7 @@ const DEFAULT_FOR_NEW_CONTENT_TYPE = 'text/turtle'
 async function patchHandler (req, res, next) {
   debug(`PATCH -- ${req.originalUrl}`)
   res.header('MS-Author-Via', 'SPARQL')
+  let releaseLock
   try {
     // Obtain details of the target resource
     const ldp = req.app.locals.ldp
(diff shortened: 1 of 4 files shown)


RubenVerborgh:
Agreed; as stated in the issue text, other methods are not done.

RubenVerborgh:
What benchmark, what does "reliably" mean, and compared to what?

If we are testing 100s of PATCH requests to the same file, then yes, you will see degraded performance compared to the original code. That's because the original code will happily handle two PATCH requests simultaneously and mess up the file.

The good test is 100s of PATCH requests to different files.

kjetilk:
I'll let @megoth fill in the details, his tests are here:
https://github.com/megoth/lock-test
but the problem appeared to be that it did in fact allow overwriting the file in some cases even when the file was locked.

RubenVerborgh:
Ah the proper-lockfile module, not this PR. That seems very weird, as the library is heavily tested and widely used. Happy to use any other locking library.

kjetilk:
Yeah. Pretty weird. 

I might add that we reviewed like 25 different modules on npm. Most relied on a db to lock, mostly redis, which we found to be a too big dep to introduce. Many were mutex modules that locked the whole process rather than a file, some, like proper-lockfile and lockfile did not work as advertised...

We also went through some options with @justinwb on Friday.

Which left us with the possibility that we need to do something POSIX-specific... Anyway, you might want to talk with @megoth about it on Monday.

RubenVerborgh:
> some, like proper-lockfile and lockfile did not work as advertised...

I'm inclined to say this is a test error rather than a library error, given the popularity of those libraries.

kjetilk:
Perhaps, but I wouldn't bet on it before we have confirmed it for ourselves.

megoth:
Ok, so with https://github.com/solid/node-solid-server/pull/1118 rebased on this branch, and with a minor tweak to https://github.com/Inrupt-inc/generator-solid-react/tree/release/profile that enables me to spam the server with PATCH calls my findings are (drumroll) *that this PR seems to be work*!!! 🎉 

Spamming the server also does not seem to slow down the handling of other resources noteworthy, which is also as expected with this solution.

Granted, this test is not 100% proof that this prevents data from concurrent writing , but I'm adding my approval fully behind this PR at least 😸

RubenVerborgh:
Great! Still need to address @timbl's concern of also locking other methods.

RubenVerborgh:
Locking added on GET/PUT/DELETE; all good to go now.

kjetilk:
@timbl are you good now?

Issue #1117, opened 2019-02-26
Returns 403 instead of 401 for certain requests


megoth:
An error that Jairo in Janeiro Digital figured out what was going wrong.

## Please describe what you did in reproducible steps

Logging in with a v5 server using the [release/profile-branch of generator-solid-react](https://github.com/Inrupt-inc/generator-solid-react/tree/release/profile) and trying to edit profile fields.

## How did it work with 4.x series servers?

It returned 401, which made [solid-auth-client](https://github.com/solid/solid-auth-client) make another request with credentials.

## What happened when you tried the same with the 5.0.0 beta?

The server returns 403 on all PATCH-requests. This stops `solid-auth-client` from sending another request.

kjetilk:
There was a test failure when I merged #1118 into the v5 branch, which made me take another look at the tests. There are [two failing tests](https://travis-ci.org/solid/node-solid-server/jobs/503870654#L1800) now that I think should work, one is different from previous branch. I have branched it out in [fix/new-branch-faulty-403](https://github.com/solid/node-solid-server/tree/fix/new-branch-faulty-403) as I had to revert the merge from #1118.

kjetilk:
Now the fix is in #1140

PR #1118, opened 2019-02-28
Making proper use of strictOrigins

megoth:
We had some faulty handling where we returned 403 even if the user wasn't authenticated. (This fixes https://github.com/solid/node-solid-server/issues/1117.)

# Some notes 

1. Discovered some redundant tests in test/integration/authentication-oidc-test.js, and moved some tests into groups for better overview. This makes for a lot of changes, but I think it's worth it.
2. Duplicated authentication-oidc-test into authentication-oidc-with-strict-origins-turned-off-test to make sure we handle those permutations correctly as well. This required a lot of setup, which are all of the new files in this commit.
3. Tried to consolidate the use of getTrustedOrigins in create-app.js, which made a test obsolete

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -206,14 +206,17 @@ function initWebId (argv, app, ldp) {
   // without permission by including the credentials set by the Solid server.
   app.use((req, res, next) => {
     const origin = req.get('origin')
-    const trustedOrigins = argv.trustedOrigins
+    const trustedOrigins = ldp.getTrustedOrigins(req)
     const userId = req.session.userId
     // Exception: allow logout requests from all third-party apps
     // such that OIDC client can log out via cookie auth
     // TODO: remove this exception when OIDC clients
     // use Bearer token to authenticate instead of cookie
     // (https://github.com/solid/node-solid-server/pull/835#issuecomment-426429003)
-    if (!argv.host.allowsSessionFor(userId, origin, trustedOrigins) && !isLogoutRequest(req)) {
+    // We only want to do this check if strictOrigin is set to false, since we trust WebACL to handle origins otherwise
+    // If we don't allow authentication with cookies we can't handle ACLs properly wrt trusted origins
+    // https://github.com/solid/node-solid-server/issues/1117
+    if (!argv.strictOrigin && !argv.host.allowsSessionFor(userId, origin, trustedOrigins) && !isLogoutRequest(req)) {
       debug.authentication(`Rejecting session for ${userId} from ${origin}`)
       // Destroy session data
       delete req.session.userId
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -213,8 +213,18 @@ function initWebId (argv, app, ldp) {
     // TODO: remove this exception when OIDC clients
     // use Bearer token to authenticate instead of cookie
     // (https://github.com/solid/node-solid-server/pull/835#issuecomment-426429003)
-    // We only want to do this check if strictOrigin is set to false, since we trust WebACL to handle origins otherwise
-    // If we don't allow authentication with cookies we can't handle ACLs properly wrt trusted origins
+    //
+    // Authentication cookies are an optimization:
+    // instead of going through the process of
+    // fully validating authentication on every request,
+    // we go through this process once,
+    // and store its successful result in a cookie
+    // that will be reused upon the next request.
+    // However, that cookie can then be sent by any server,
+    // even servers that have not gone through the proper authentication mechanism.
+    // However, if trusted origins are enabled,
+    // then any origin is allowed to take the shortcut route,
+    // since malicious origins will be banned at the ACL checking phase.
     // https://github.com/solid/node-solid-server/issues/1117
     if (!argv.strictOrigin && !argv.host.allowsSessionFor(userId, origin, trustedOrigins) && !isLogoutRequest(req)) {
       debug.authentication(`Rejecting session for ${userId} from ${origin}`)
--- a/lib/models/solid-host.js
+++ b/lib/models/solid-host.js
@@ -77,7 +77,11 @@ class SolidHost {
     const serverHost = getHostName(this.serverUri)
     if (originHost === serverHost) return true
     if (originHost.endsWith('.' + serverHost)) return true
-    return !!trustedOrigins.includes(origin)
+    // Allow the user's own domain
+    const userHost = getHostName(userId)
+    if (originHost === userHost) return true
+    if (trustedOrigins.includes(origin)) return true
+    return false
   }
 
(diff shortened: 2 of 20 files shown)


megoth:
I realize the number of changes introduced in `test/integration/authentication-oidc-test.js` might be a bit too much...

@kjetilk @RubenVerborgh @dmitrizagidulin let me know if you want me to ungroup the tests in that file, it should reduce the number of changes a lot. But it also makes it harder to get an overview of the tests =/ (which had already created some redundant tests)

megoth:
Could also move `test/integration/authentication-oidc-with-strict-origins-turned-off-test.js` into a separate commit, if wanted; that would also reduce the number of changes a lot.

lib/create-app.js
```diff
@@ -206,14 +206,17 @@ function initWebId (argv, app, ldp) {
...
+    const trustedOrigins = ldp.getTrustedOrigins(req)
     const userId = req.session.userId
     // Exception: allow logout requests from all third-party apps
     // such that OIDC client can log out via cookie auth
     // TODO: remove this exception when OIDC clients
     // use Bearer token to authenticate instead of cookie
     // (https://github.com/solid/node-solid-server/pull/835#issuecomment-426429003)
-    if (!argv.host.allowsSessionFor(userId, origin, trustedOrigins) && !isLogoutRequest(req)) {
+    // We only want to do this check if strictOrigin is set to false, since we trust WebACL to handle origins otherwise
+    // If we don't allow authentication with cookies we can't handle ACLs properly wrt trusted origins
```
RubenVerborgh: That's not the right explanation. We can handle ACLs properly; with non-cookie auth mechanisms.

The full explanation is:

Authentication cookies are an optimization:
instead of going through the process of
fully validating authentication on every request,
we go through this process once,
and store its successful result in a cookie
that will be reused upon the next request.
However, that cookie can then be sent by any server,
even servers that have not gone through the proper authentication mechanism.
However, if trusted origins are enabled,
then any origin is allowed to take the shortcut route,
since malicious origins will be banned at the ACL checking phase.

lib/models/solid-host.js
```diff
@@ -77,11 +77,7 @@ class SolidHost {
     const serverHost = getHostName(this.serverUri)
     if (originHost === serverHost) return true
     if (originHost.endsWith('.' + serverHost)) return true
-    // Allow the user's own domain
-    const userHost = getHostName(userId)
-    if (originHost === userHost) return true
-    if (trustedOrigins.includes(origin)) return true
-    return false
+    return !!trustedOrigins.includes(origin)
```
RubenVerborgh: which by default included the user's own domain?

RubenVerborgh: Apparently not, given the test: is this intended?

lib/create-app.js
megoth: I'm having problem wrapping my head around the reasoning for this part of the code, so thank you for the explanation. I think it's a good explanation, so I added it to the PR, hope that's ok.

lib/models/solid-host.js
megoth: I tried to [reuse `ldp.getTrustedOrigins`](https://github.com/solid/node-solid-server/blob/fix/faulty-403/lib/ldp.js#L467), which includes the user's own domain if the server is a multiuser server.

I thought we could remove some redundant code, but maybe I should revert this change back, since others might not use `ldp.getTrustedOrigins`?

megoth:
I just discovered that some tests will fail if run by themselves, but succeed if run as part of running all tests in that file...

An example of this is ["ACL with WebID+OIDC over HTTP > Origin > user1 should be able to access test directory"](https://github.com/solid/node-solid-server/blob/master/test/integration/acl-oidc-test.js#L285). This should actually fail, at least if I understand the logic correctly.

I will try to commit a fix for this soon.

megoth:
Added two more commits that removes the need of ordering in tests running in `test/integration/acl-oidc-test.js` and `test/integration/acl-tls-test.js`.

I think these commits could've been extracted into a separate PR for clarity. Let me know if this is desired.

megoth:
@RubenVerborgh I moved some code into the constructor of the ACLChecker, tests verifies it works nicely ^_^

I've fell a bit out of loop on this PR, were there more feedback you had that I should fix?

megoth:
> OK, it looks good. What makes me uneasy is the very extensive changes to authentication-oidc-test.js, since a lot of lines have changed, it had very thorough review before this, so I was thinking that not much should change in there.
> 
> I suppose your tooling did many of those changes, @megoth , so perhaps it is just that, but it is hard for me to see what has actually changed there.

It was a deliberate choice, because the tests were badly grouped which made them very hard to read.

But I can revert those changes and we could live with it instead of having this many changes.

kjetilk:
> It was a deliberate choice, because the tests were badly grouped which made them very hard to read.
> 
> But I can revert those changes and we could live with it instead of having this many changes.

I'm all for cleaning up, but this ended up being a rather large change, and I find it hard to review. So, perhaps the grouping changes could be done in a separate PR? Would that be a lot of work at this point?

megoth:
> > It was a deliberate choice, because the tests were badly grouped which made them very hard to read.
> > But I can revert those changes and we could live with it instead of having this many changes.
> 
> I'm all for cleaning up, but this ended up being a rather large change, and I find it hard to review. So, perhaps the grouping changes could be done in a separate PR? Would that be a lot of work at this point?

I'm going to try to revert the big changes to authentication-oidc-test.js, so that you only have to review my added tests ^_^

megoth:
@kjetilk I've reverted a bunch of unnecessary changes, most of the changes are now simply new files. Let me know if you want me to do more 😸

test/integration/authentication-oidc-test.js
```diff
@@ -332,8 +378,8 @@ describe('Authentication API (OIDC)', () => {
               })
           })
 
-          it('should return a 403', () => {
-            expect(response).to.have.property('status', 403)
+          it('should return a 401', () => {
```
kjetilk: Not terribly important, but as per the comment above the test, I still think 403 is the correct status. How hard would it be to fix?

test/integration/authentication-oidc-test.js
```diff
@@ -369,8 +432,8 @@ describe('Authentication API (OIDC)', () => {
...
```
kjetilk: Again here I would prefer a 403, since it is good to fend off an attacker at once, but it is even less important than the previous comment, since it is most important to just deny them.

kjetilk:
Two small comments inline, the rest looks good, thanks a lot. :-) 

Please feel free to submit a separate PR with the cleanup later!

Issue #1124, opened 2019-03-04
Make redirect working


y0va:
On my platform, ubuntu 18.04, node 11.10.1 the command line switch `--redirect-http-from [port]`does not work without any errors in the debug logs.

megoth:
Which version of the server are you working with?

megoth:
And could you provide some of the errors you're seeing perhaps?

megoth:
Oh, great, you have a fix already! Awesome 😸

y0va:
There had been no errors. I'm using master.

 :gift:

megoth:
This should be easy to merge into [release/v5.0.0](https://github.com/solid/node-solid-server/tree/release/v5.0.0) as well I think. Will talk to @kjetilk about it when he's back from conference.

PR #1125, opened 2019-03-04
repairs redirect due to case typo correction

y0va:
This fixes https://github.com/solid/node-solid-server/issues/1124

### DIFF
--- a/lib/create-server.js
+++ b/lib/create-server.js
@@ -80,20 +80,20 @@ function createServer (argv, app) {
   }
 
   // Look for port or list of ports to redirect to argv.port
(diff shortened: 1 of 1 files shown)


megoth:
The question is if we want to release a version 4.4.2 for this. I think we should, but I'm good either way.

kjetilk:
OK, lets talk about when I get home tomorrow.

kjetilk:
Yeah, it makes sense, thanks a lot @y0va ! We'll do a 4.4.2 release, and merge it into 5.0.0 by merging master into the 5.0.0 branch.

PR #1127, opened 2019-03-04
Expand SolidHost config constructor, plus tests

dmitrizagidulin:
Minor refactor, expand the `SolidHost` constructor.

Extracted from PR #702, partly addresses issue #483.


dmitrizagidulin:
@kjetilk so, the impact of the change (just passing in the `argv` config object directly) is quite minimal. You can think of the `SolidHost` instance as a class that holds (some of) the config attributes of `argv`.

dmitrizagidulin:
@michielbdejong Yes, there is downstream code that checks those params. (And no, they're not live-configurable, they're just from the config file only)

PR #1136, opened 2019-03-11
Fix new branch faulty 403

michielbdejong:


michielbdejong:
* even if strictOrigin is set to false, cookies should not be trusted cross-origin; fixed this [here](https://github.com/solid/node-solid-server/pull/1136/commits/bef2be11bed5d0b33adf22b0d477861f7f2ac4ac#diff-e3e86891d0a2b9d1affa17a9954643daR252)
* puzzled by the behaviour [here](https://github.com/solid/node-solid-server/pull/1136/commits/bef2be11bed5d0b33adf22b0d477861f7f2ac4ac#diff-e3e86891d0a2b9d1affa17a9954643daR238), added a FIXME note
* added a test for bearer token auth with untrusted origin and strictOrigin=true (expects a  403)
* added a test for bearer token auth with untrusted origin and strictOrigin=true (expects a  200)

PR #1137, opened 2019-03-11
Fix faulty 403

kjetilk:
This is a fix for #1117 , building on #1118, but the merge process here has become somewhat muddled. Creating PR anyway.


michielbdejong:
@kjetilk can this be merged now? All this is is your notes and my test fixes from #1136, right?

kjetilk:
@michielbdejong but I think that is what is wrong about it, as I reverted the changes from Arne, and so it seems like we do not have the rest of the issue resolved. I'm in doubt here, lets check.

PR #1140, opened 2019-03-12
Redo #1118

michielbdejong:
@kjetilk I think this is the changeset you were expecting to see in #1137 right?
What I did was:
```sh
git checkout origin/fix/new-branch-faulty-403
git reset origin/release/v5.0.0
git add .
git branch redo-1118
git checkout redo-1118
git commit -am"Redo #1118"
```

All tests are passing now when I ran them locally.


kjetilk:
Yes, thanks a lot! I would have preferred if all the commits were visible, but I think this is something we can re-review. Please have a go at it!

PR #1158, opened 2019-03-29
Fixed JSON not being written to file

jaxoncreed:


.gitignore
```diff
@@ -15,3 +15,4 @@
 /coverage
 /node_modules
 /npm-debug.log
+.DS_Store
```
kjetilk: It looks good, except for this, which I think should be in a local gitignore file :-)

Issue #1171, opened 2019-04-10
Set default response type to text/turtle for the root folder


Otto-AA:
## Please describe what you did in reproducible steps
Fetch the root folder of a pod on a v5.0.0 server without specifying the "Accept" header.
(e.g. GET https://testomat.solid.open.ac.uk/, `Accept: */*`)

## How did it work with 4.x series servers?
It returned a text/turtle file.

## What happened when you tried the same with the 5.x series server?
It returns a text/html file.

## Any material that will help, logs, error messages, etc.
This is a follow up question of [this issue](https://github.com/solid/node-solid-server/issues/1120).

As far as I understand the [spec](https://github.com/solid/solid-spec/blob/master/api-rest.md#reading-resources) it should return a `text/turtle` response by default if a container (folder) is requested. This also was the case for the root folder in 4.x, but changed with 5.0.0.

> IMPORTANT: a default Content-Type: text/turtle will be used for requests for RDF resources or views (containers) that do not have an Accept header.

[Here](https://github.com/jeff-zucker/solid-file-client/issues/33#issuecomment-481823805) is a more precise explanation of the behavior of the different versions, if the above was not expressive enough.

Otto-AA:
I've noticed another difference from 5.x to older versions regarding the root container:
POST to the root responds with 403 (e.g. creating a new folder or file in the root).
GET of an item in the root responds with 403 if it doesn't exist (e.g. GET {root}/non-existent/).

Both work with the same settings in my private folder. In the profile I've set read/write/append/control access. I sadly don't have the ability to look into server settings, so maybe that could be the reason too.

Is this related to this issue, or should I open a new one?

![image](https://user-images.githubusercontent.com/21040751/55935481-05fc8200-5c34-11e9-9a04-424c8e577864.png)

jaxoncreed:
Fixed in #1173

Issue #1120, opened 2019-03-01
Error translating between RDF formats


NoelDeMartin:
## Please describe what you did in reproducible steps
- Added the url of my app (localhost:8080) to `trustedOrigins`
- Registered a user using `/api/accounts/new`
- Logged in from [my app](https://noeldemartin.github.io/solid-focus/) running on localhost.
- Created documents and containers -- worked as expected.
- Refreshed the page: got a "user not logged in" response from reading the root: `https://localhost:8443`.
- After looking at the network request, I noticed the header `Accept` was `*/*`, so I changed it to `text/turtle`. Now I get the following error: "Error translating between RDF formats"

## How did it work with 4.x series servers?
I got a turtle document with information of documents in my POD.

## What happened when you tried the same with the 5.0.0 beta?
I got the error "Error translating between RDF formats".

## Any material that will help, logs, error messages, etc.
Request headers:

```
GET / HTTP/1.1
Host: localhost:8443
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:65.0) Gecko/20100101 Firefox/65.0
Accept: text/turtle
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br
Referer: http://localhost:8080/
[...]

PR #1173, opened 2019-04-16
Fix/#1171

jaxoncreed:
Fix for issue #1171. Caused by the fix for issue #1120.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -45,7 +45,7 @@ async function handler (req, res, next) {
 
   let ret
   try {
-    ret = await ldp.get(options, req.headers.accept.includes('text/html'))
+    ret = await ldp.get(options)
   } catch (err) {
     // use globHandler if magic is detected
     if (err.status === 404 && glob.hasMagic(path)) {
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -45,7 +45,7 @@ async function handler (req, res, next) {
 
   let ret
   try {
-    ret = await ldp.get(options)
+    ret = await ldp.get(options, req.headers.accept.includes('text/html'))
   } catch (err) {
     // use globHandler if magic is detected
     if (err.status === 404 && glob.hasMagic(path)) {
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -42,7 +42,7 @@ class ResourceMapper {
   }
 
   // Maps the request for a given resource and representation format to a server file
-  // When searchIndex is true and the URL ends with a '/', indexFilename will be matched.
+  // When searchIndex is true and the URL ends with a '/', and contentType includes 'text/html' indexFilename will be matched.
   async mapUrlToFile ({ url, contentType, createIfNotExists, searchIndex = true }) {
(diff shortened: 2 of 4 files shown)


michielbdejong:
I'm trying to decipher what this code did and what you changed. It looks like the decision about whether or not to return index.html is made through many steps, looking at https://github.com/solid/node-solid-server/blob/5e0748ec776b2ebdc495c7fca1a0ac1f9d41b37c/lib/resource-mapper.js:

* searchIndex, a boolean that indicates whether we should try to serve index.html if the requests ends in a slash.
* isIndex, a boolean that indicates whether the URL actually does end in a slash first made on line #L48 (which basically says, if a URL ends in / then isIndex is true).
* createIfNotExists check (don't fully understand that yet)
* this._indexContentType check (don't fully understand that yet)
* Line 104 checks if an index.html file actually exists (note that this question is irrelevant if the http request wants a Turtle response)
* Lines https://github.com/solid/node-solid-server/blob/5e0748ec776b2ebdc495c7fca1a0ac1f9d41b37c/lib/resource-mapper.js#L74-L80 then possibly reverse the previous decision to server index.html, based on the contentType header.

We cannot touch this code without refactoring that decision tree. I would propose to replace it with the following:
* the comment on line https://github.com/solid/node-solid-server/blob/5e0748ec776b2ebdc495c7fca1a0ac1f9d41b37c/lib/resource-mapper.js#L45 to say "Under the following conditions, a trailing `/` in the URL will be interpreted as `/index.html`:
  * the URL ends in a /
  * the searchIndex option is true (which is the default)
  * the content-type header of the request indicates that text/html is prefered over text/turtle
This means that a request for a URL that ends in a / may for instance return a 404 even if the container does exist."

PS: looks to me like the `searchIndex &&` on line https://github.com/solid/node-solid-server/blob/5e0748ec776b2ebdc495c7fca1a0ac1f9d41b37c/lib/resource-mapper.js#L62 should be removed, the decision to add a dollar extension should not be conditional on searchIndex being set to true, right?

jaxoncreed:
I think you're right that the way I fixed the bug was incorrect and I should instead do it with the flags rather than reversing it at the end.

I am unsure that using a trailing "/" to determine if you should serve an index.html correctly aligns with the specs. https://github.com/solid/solid-spec/blob/master/api-rest.md#reading-resources It's clearly stated that if a content-type header is not provided the result should always be in text/turtle

michielbdejong:
ah ok, but then `if isIndex is true, and the URL ends in a slash, and the content-type header of the request indicates that text/html is preferred over text/turtle, fullPath += this._indexFilename` is all we need, right?

jaxoncreed:
In general, this code is pretty messy, but after looking around a bit, I determined that this is the best place to determine whether or not to send html based on the accept header

michielbdejong:
Will have a look tomorrow!

michielbdejong:
Thanks for moving the fix to a better location, the next person touching this code will love you for that! :)

jaxoncreed:
Okay, the last solution had some major problems. This one is less clean, but given the way the code is structured, this is the proper way to do it.

Issue #1179, opened 2019-04-26
.acl files not deleted, become unfindable


jeff-zucker:
If I delete a file /foo/bar.txt that is controlled by /foo/bar.txt.acl, the .acl file does not get deleted.  It becomes invisible because without being able to query /foo/bar.txt's link header, there's no way to know where its .acl should be and .acl files are not discover-able with a GET on a container.  This now-unfindable file then prevents us from deleting it's container /foo/ since the container is non-empty and returns 409 on attempt to delete it even though it "looks" empty.

jeff-zucker:
Another problem with this: if I delete /foo/bar.txt but not /foo/bar.txt.acl and I later create /foo/bar.txt - it will be controlled by /foo/bar.txt.acl without me specifically telling it to be.

RubenVerborgh:
See https://github.com/solid/solid-spec/issues/157

RubenVerborgh:
Suggested solution: https://github.com/solid/node-solid-server/issues/1179
Alternative solution: https://github.com/solid/solid-spec/issues/190

RubenVerborgh:
Also need to consider security, in https://github.com/solid/solid-spec/issues/196

Issue #190, opened 2016-01-18
Add preferences file and default workspace to new accounts.


deiu:
When creating new accounts, the server should also generate a preferencesFile and set the default workspace in it.

nicola:
See my PR on this.

dmitrizagidulin:
Looks like this is safe to close?

deiu:
Fixed in #199

PR #1180, opened 2019-04-27
Delete .acl file for resource when resource is deleted

fabiancook:
Delete .acl file for resource when resource is deleted, also ignore any acl or meta files in directory when deleting. 

I had troubles with some TLS not passing, locally I skipped them. I also had issues with the quota tests, its like the signature of ldp changed at some point, so I redid the test so it tests correctly.

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -412,7 +412,7 @@ class LDP {
     }
 
     // Ensure the container is empty (we ignore .meta and .acl)
-    if (list.some(file => file !== this.suffixMeta && file !== this.suffixAcl)) {
+    if (list.some(file => !file.endsWith(this.suffixMeta) && !file.endsWith(this.suffixAcl))) {
       throw error(409, 'Container is not empty')
     }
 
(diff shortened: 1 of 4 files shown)


fabiancook:
Fixes https://github.com/solid/node-solid-server/issues/1179

fabiancook:
I'll rebase in & re-push once the other failing tests are corrected.

jaxoncreed:
@fabiancook There seems to be a problem with quotas causing the test to fail. Would you mind looking at that?

fabiancook:
The quota tests look like they've never worked, I don't have the bandwidth to debug those errors at this time.

kjetilk:
I think they did work. But I'm unsure why this should need to touch those?

jaxoncreed:
Yeah. I checked yesterday. The current master has 100% passing tests.

fabiancook:
Perfect, I'll rebase to current master

fabiancook:
Can someone help me with the quota issue? A pointer would be helpful. 

The original quota tests never would have run, either, which is why I added a new one.

fabiancook:
The only thing I'm wondering now is if I should change this line here so only `acl` files are checked for, where `meta` files are only ignored if it is the containers `meta` file:

https://github.com/solid/node-solid-server/pull/1180/commits/4e268e3bb2f1731da190343bf1f9db01cdb0a872#diff-4b9b8e50857ade6dd5db969387fe5f20R415 

This would depend on this issue: https://github.com/solid/solid-spec/issues/172

RubenVerborgh:
See https://github.com/solid/solid-spec/issues/157 as well

jaxoncreed:
So I think we're good to go to merge this then.

RubenVerborgh:
> So I think we're good to go to merge this then.

I'm not sure, do we have consensus that this is the desired behavior?
There are security consequences when this happens.

Also, given that we're considering #190, there are other means of discovering and deleting the ACL file.

kjetilk:
Technically, it looks good to me, but I think we don't yet have consensus on whether this is what we want.

jaxoncreed:
It sounds like @RubenVerborgh 's two suggestions should be fixed first.

RubenVerborgh:
Also need to consider security, in https://github.com/solid/solid-spec/issues/196

michielbdejong:
Continued in #1415.

Issue #1182, opened 2019-04-29
Can't access login on iOS fullscreen


emiliocortina:
Hi!
I am part of a team of students who are developing a chat application based in the Solid principles.
We have a problem with the application and it is because we have made it to work fullscreen on iOS just adding two meta tags to the html documents in Angular.

The thing is that we are unable to use the application in this mode because when we try to do the login, the Solid login page doesn't implement these tags and hence it redirects the user to the browser, resulting in the application not getting the session.
I was wondering if I could add [these tags](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html) to the head element of the file in node-solid-server/default-views/auth/login.hbs.
The specific tags would be these:
<meta name="apple-mobile-web-app-capable" content="yes">
<meta name="apple-mobile-web-app-status-bar-style" content="default">

I am not 100% sure this will solve all the problems running the application fullscreen on iOS, but I thought it was worth a try.

Thank you!

PR #1183, opened 2019-04-29
Adding fullscreen support for login on iOS

emiliocortina:
#1182

### DIFF
--- a/default-views/auth/login.hbs
+++ b/default-views/auth/login.hbs
@@ -6,6 +6,8 @@
   <title>Login</title>
   <link rel="stylesheet" href="/common/css/bootstrap.min.css">
   <link rel="stylesheet" href="/common/css/solid.css">
+  <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


emiliocortina:
@jaxoncreed Sorry for taking so long to answer. 
It also seems to happen when I run the tests locally. I checked the tests failing in travis and locally and they appear to be the same.
The thing is that I ran the tests against the repository before the changes I made and it still shows the same failing tests.
Screenshot 2019-05-06 at 12 43 54

jaxoncreed:
Oh that's interesting. I'm seeing the same thing. I'll look into that and get back to you.

jaxoncreed:
Okay, @emiliocortina you're code's all good. Looks like one of the sites we have deployed to help us with tests is done.

I've pinged @RubenVerborgh to help us out with that, so once it's back up I'll run the tests.

emiliocortina:
@jaxoncreed Great! Thanks for your help.

PR #1192, opened 2019-05-20
Update databrowser for check, that session wasn't expired.

smalinin:
Now databrowser doesn't work properly, when session token was expired or damaged, the NSS returns 401 code, but mashlib code ignore it in ajax/fetch calls, when user tried to update data or upload new data, so we check, if session token is still valid and not expired, when `databrowser.html` is loaded.

### DIFF
--- a/static/databrowser.html
+++ b/static/databrowser.html
@@ -5,10 +5,25 @@
     <script type="text/javascript" src="/common/js/mashlib.min.js"></script>
     <script>
       const $SOLID_GLOBAL_config = {}
-      document.addEventListener('DOMContentLoaded', function () {
+      document.addEventListener('DOMContentLoaded', async function () {
         const panes = require('mashlib')
         const UI = panes.UI
 
+        try {
+          var session = await SolidAuthClient.currentSession();
+          if (session && session.webId) {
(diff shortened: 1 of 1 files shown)


static/databrowser.html
```diff
@@ -5,10 +5,25 @@
...
+      document.addEventListener('DOMContentLoaded', async function () {
         const panes = require('mashlib')
         const UI = panes.UI
 
+        try {
+          var session = await SolidAuthClient.currentSession();
+          if (session && session.webId) {
+            // Check, that session has access to user data and it wasn't expired
+            var url = new URL(session.webId);
+            url.pathname = '/settings/privateTypeIndex.ttl';
```
jaxoncreed: Good idea, though I would be hesitant to use `/settings/privateTypeIndex.ttl`. The spec doesn't reference that it MUST exist. The only reference is in a proposal written 3 years ago (https://github.com/solid/solid/blob/a25b6d11477fc3ec11d061c4945ff1c63e1ee65b/proposals/data-discovery.md)

But, databrowser.html does only exist on NSS, so I'll bring in @dmitrizagidulin for the final word on this

michielbdejong:
Continued in the mashlib repo ^

PR #1202, opened 2019-05-24
Does no show consent screen on subdomains of NSS

### LINKED ISSUE #1200: don't show consent dialog for data browser
when logging in to the databrowser, it does not make sense to show the user a consent screen. the databrowser should always have read+write+control, since it's same-origin.

jaxoncreed:
Fix #1200

### DIFF
--- a/lib/requests/consent-request.js
+++ b/lib/requests/consent-request.js
@@ -66,11 +66,13 @@ class ConsentRequest extends AuthRequest {
   static async get (req, res) {
     const request = ConsentRequest.fromParams(req, res)
 
-    const appOrigin = request.getAppOrigin()
+    const appUrl = request.getAppUrl()
+    const appOrigin = `${appUrl.protocol}//${appUrl.host}`
+    const serverUrl = url.parse(req.app.locals.ldp.serverUri)
     // Check if is already registered or is data browser
     if (request.isUserLoggedIn()) {
       if (
-        appOrigin === req.app.locals.ldp.serverUri ||
+        (appUrl && appUrl.host.includes(serverUrl.host) && appUrl.protocol === serverUrl.protocol) ||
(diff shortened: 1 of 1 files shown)


lib/requests/consent-request.js
```diff
@@ -66,11 +66,13 @@ class ConsentRequest extends AuthRequest {
...
 
-    const appOrigin = request.getAppOrigin()
+    const appUrl = request.getAppUrl()
+    const appOrigin = `${appUrl.protocol}//${appUrl.host}`
+    const serverUrl = url.parse(req.app.locals.ldp.serverUri)
     // Check if is already registered or is data browser
     if (request.isUserLoggedIn()) {
       if (
-        appOrigin === req.app.locals.ldp.serverUri ||
+        (appUrl && appUrl.host.includes(serverUrl.host) && appUrl.protocol === serverUrl.protocol) ||
```
RubenVerborgh: server: `https://safe.com/`
app: `https://unsafe.com/`

server: `https://safe.com/`
app: `https://safe.com.evil.com/`

jaxoncreed: Ah right

RubenVerborgh: Tests are really important for these security cases.

jaxoncreed: Yep, that was a lapse in judgment on my part. If we get the green light on deploying going forward with this version I'll crank out some more tests.

Though thankfully this wouldn't have been a security hole as this only controls if the consent screen pops up, and not if the app is allowed.

RubenVerborgh: The logic should only reside in one place though.

michielbdejong:
What's the status of this, can we get this merged before the deploy?

PR #1210, opened 2019-05-31
Sharing patch

### LINKED ISSUE #1209: scrambled profile documents
Several users have reported their profile document on dev.inrupt.net got corrupted. One of the things that happens is that references to https://localhost:8443 show up where they shouldn't.

This may be caused by the consent dialog, by the data browser, or by the react-sdk-app.

Sounds like a significant bug that we should at least understand before we next deploy NSS to any production servers.

jaxoncreed:
Patches for what was previously called the consent dialog and is now called the sharing dialog:

 - [X] Fix #1209 
 - [x] Include the new dialog text
 - [x] Rename everything to "sharing" rather than "consent"
 - [x] Support for independently hosted webIds
 - [x] Bump databrowser version


jaxoncreed:
I guess the consent dialog isn't 100% approved yet... but I guess we're good for a deploy

default-views/auth/sharing.hbs
```diff
@@ -0,0 +1,49 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>{{title}}</title>
+  <!-- Bootstrap CSS and Theme for demo purposes -->
+  <link rel="stylesheet" href="/common/css/bootstrap.min.css">
+  <link rel="stylesheet" href="/common/css/solid.css">
+</head>
+<body>
+<div class="container title">
+  <h1>Authorize {{app_origin}} to access your Pod?</h1>
+  <p>Solid allows you to precisely choose what other people and apps can read and write in a Pod. This version of the authorization user interface (node-solid-server V5.1.1) only supports the toggle of global access permissions to all of the data in your Pod.</p>
```
kjetilk: Not sure it is a good idea to write a version number in the text, it has to be updated by hand, right?

lib/requests/sharing-request.js
```diff
@@ -64,19 +64,26 @@ class ConsentRequest extends AuthRequest {
    * @param res {ServerResponse}
    */
   static async get (req, res) {
-    const request = ConsentRequest.fromParams(req, res)
+    const request = SharingRequest.fromParams(req, res)
 
-    const appOrigin = request.getAppOrigin()
-    // Check if is already registered or is data browser
+    const appUrl = request.getAppUrl()
+    const appOrigin = `${appUrl.protocol}//${appUrl.host}`
```
kjetilk: Could you construct this URL using setters on the object? For something this simple, it is probably OK to do it this way, but in general, there has been a lot of bugs that comes because suddenly, the protocol doesn't have a colon and stuff like that. So, I'd like to avoid string manipulation on URLs, and rather use setters on the object.

default-views/auth/sharing.hbs
michielbdejong: Good point. But I don't see an easy way to grab it from the package.json and insert it here. I'll change it to 5.1, because hopefully 5.2 or at least 5.3 will have granular access scopes and we can update or remove this text again already.

lib/requests/sharing-request.js
michielbdejong: I switched it so uses `new url.URL` instead of `url.parse` so now `appUrl.origin` is available and we don't have to construct it from strings.

lib/requests/sharing-request.js
```diff
@@ -64,19 +64,26 @@ class ConsentRequest extends AuthRequest {
...
     if (request.isUserLoggedIn()) {
       if (
-        appOrigin === req.app.locals.ldp.serverUri ||
+        new url.URL(request.session.subject._id).origin !== serverUrl.origin ||
+        (appUrl && request.isSubdomain(serverUrl.host, appUrl.host) && appUrl.protocol === serverUrl.protocol) ||
         await request.isAppRegistered(req.app.locals.ldp, appOrigin, request.session.subject._id)
       ) {
-        request.setUserConsent(appOrigin)
-        request.redirectPostConsent()
+        console.log('We be here')
```
kjetilk: This got a bit chatty when trying it out. We shouldn't have `console.log` in at this point, but it could be a `debug` there, but then something more informative.

michielbdejong:
Tested locally and seems to work (don't forget to `cp default-views/auth/sharing.hbs config/views/auth/`). @kjetilk can you merge and tag v5.1.2 so I can deploy it to dev.inrupt.net?

PR #1213, opened 2019-06-02
Restrict knowledge of the URL/file mapping again to ResourceMapper

### LINKED ISSUE #1211: Putting an ACL file whose URI contains encoded spaces fails because server unencodes the spaces and crashes
## Please describe what you did in reproducible steps

Edited am ACL file using the data bowser

## How did it work with 4.x series servers?

There was no validity check
## What happened when you tried the same with the 5.x series server?
Client log:
```
T http://localhost:3080/timbl/Public/Test/2018/Cascade%20test/ACL%20Pane%20tests/.acl 400 (Bad Request)
```
Server log:

```
  solid:ldp VALIDATE -- Error parsing data: Error: Error trying to parse <http://localhost:3080/Library/WebServer/Documents/timbl/Public/Test/2018/Cascade test/ACL Pane tests/.acl> as text/turtle:
  solid:ldp Error: Error: NamedNode IRI "http://localhost:3080/Library/WebServer/Documents/timbl/Public/Test/2018/Cascade test/ACL Pane tests/.acl" must not contain unencoded spaces.:
  solid:ldp Error: Error: NamedNode IRI "http://localhost:3080/Library/WebServer/Documents/timbl/Public/Test/2018/Cascade test/ACL Pane tests/.acl" must not contain unencoded spaces.
  solid:ldp     at new NamedNode (/devel/github.com/solid/node-solid-server/node_modules/rdflib/lib/named-node.js:48:13)
  solid:ldp     at IndexedFormula.sym (/devel/github.com/solid/node-solid-server/node_modules/rdflib/lib/formula.js:502:14)
  solid:ldp     at new __SinkParser (/devel/github.com/solid/node-solid-server/node_modules/rdflib/lib/n3parser.js:240:29)
  solid:ldp     at SinkParser (/devel/github.com/solid/node-solid-server/node_modules/rdflib/lib/n3parser.js:215:16)
  solid:ldp     at Object.parse (/devel/github.com/solid/node-solid-server/node_modules/rdflib/lib/parse.js:27:15)
  solid:ldp     at LDP.isValidRdf (/devel/github.com/solid/node-solid-server/lib/ldp.js:195:12)
  solid:ldp     at putAcl (/devel/github.com/solid/node-solid-server/lib/handlers/put.js:41:11)
  solid:ldp     at processTicksAndRejections (internal/process/next_tick.js:81:5) +0ms
  solid:server Error page because of: { [HTTPError: RDF file contains invalid syntax]
  name: 'HTTPError',
  message: 'RDF file contains invalid syntax',
  status: 400 } +1ms
```

### LINKED ISSUE #1212: Get ResourceMapper back to its minimal interface
The purpose of `ResourceMapper` was to put the knowledge for URL-to-file and file-to-URL conversion on one place and one place only. That's why only two methods were exposed: `mapUrlToFile` and `mapFileToUrl`, and they were very deeply tested. This avoided a whole range of bugs caused by spaces, escapes, etc. You can see this old state [here](https://github.com/solid/node-solid-server/blob/51ee61e7d9b89f48db957f7eadee5ca120f892c8/lib/resource-mapper.js).

However, [later versions](https://github.com/solid/node-solid-server/blob/v5.1.1/lib/resource-mapper.js) started exposing extra methods, such as `resolveUrl`, `getFullPath`.

On the one hand, it is uncertain whether these are as well-tested. On the other hand, it leads to modules again [rolling their own resolving](https://github.com/solid/node-solid-server/blob/bdc5acba326f215c3d32eb2b234ec73d0f5cc9ce/lib/handlers/put.js#L39-L40) and introducing errors in the process. @timbl has found cases where URLs suddenly have unescaped spaces and are thus invalid (#1211). In the linked example, we see a bug happening:

```javascript
// This is a file path (unescaped spaces)
const path = ldp.resourceMapper.getFullPath(req)
// Woops, now suddenly it is a URL (invalid, because unescaped spaces!)
const requestUri = await ldp.resourceMapper.resolveUrl(req.hostname, path)
```

Propose resolution: do not expose `resolveUrl` and `getFullPath` (underscore them); only `mapUrlToFile` and `mapFileToUrl` should be used.

timbl:
Closes #1211.
Closes #1212.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -136,7 +136,6 @@ async function globHandler (req, res, next) {
   // TODO: This is a hack, that does not check if the target file exists, as this is quite complex with globbing.
   // TODO: Proper support for this is not implemented, as globbing support might be removed in the future.
   const filename = ldp.resourceMapper.getFilePath(req)
-  const requestUri = (await ldp.resourceMapper.mapFileToUrl({ path: filename, hostname: req.hostname })).url
 
   const globOptions = {
     noext: true,
@@ -144,7 +143,7 @@ async function globHandler (req, res, next) {
     nodir: true
   }
 
-  glob(filename, globOptions, function (err, matches) {
+  glob(filename, globOptions, async (err, matches) => {
     if (err || matches.length === 0) {
       debugGlob('No files matching the pattern')
       return next(error(404, 'No files matching glob pattern'))
@@ -154,7 +153,7 @@ async function globHandler (req, res, next) {
     const globGraph = $rdf.graph()
 
     debugGlob('found matches ' + matches)
-    Promise.all(matches.map(match => new Promise(async (resolve, reject) => {
+    await Promise.all(matches.map(match => new Promise(async (resolve, reject) => {
       const urlData = await ldp.resourceMapper.mapFileToUrl({ path: match, hostname: req.hostname })
       fs.readFile(match, {encoding: 'utf8'}, function (err, fileData) {
         if (err) {
@@ -178,15 +177,15 @@ async function globHandler (req, res, next) {
         })
       })
     })))
-    .then(() => {
-      const data = $rdf.serialize(undefined, globGraph, requestUri, 'text/turtle')
-      // TODO this should be added as a middleware in the routes
-      res.setHeader('Content-Type', 'text/turtle')
-      debugGlob('returning turtle')
 
-      res.send(data)
-      return next()
-    })
+    const requestUri = ldp.resourceMapper.getRequestUrl(req)
+    const data = $rdf.serialize(undefined, globGraph, requestUri, 'text/turtle')
+    // TODO this should be added as a middleware in the routes
+    res.setHeader('Content-Type', 'text/turtle')
+    debugGlob('returning turtle')
+
+    res.send(data)
+    next()
   })
 }
 
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -143,7 +143,7 @@ async function globHandler (req, res, next) {
 
   // Extract the folder on the file system from the URL glob
   const folderUrl = requestUrl.substr(0, requestUrl.length - 1)
-  const folderPath = (await ldp.resourceMapper.mapUrlToFile({ url: folderUrl })).path
+  const folderPath = (await ldp.resourceMapper.mapUrlToFile({ url: folderUrl, searchIndex: false })).path
 
   const globOptions = {
     noext: true,
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -132,10 +132,18 @@ async function handler (req, res, next) {
 }
 
 async function globHandler (req, res, next) {
-  const ldp = req.app.locals.ldp
-  // TODO: This is a hack, that does not check if the target file exists, as this is quite complex with globbing.
-  // TODO: Proper support for this is not implemented, as globbing support might be removed in the future.
-  const filename = ldp.resourceMapper.getFilePath(req)
+  const { ldp } = req.app.locals
+
+  // Ensure this is a glob for all files in a single folder
+  // https://github.com/solid/solid-spec/pull/148
+  const requestUrl = await ldp.resourceMapper.getRequestUrl(req)
+  if (!/^[^*]+\/\*$/.test(requestUrl)) {
+    return next(error(404, 'Unsupported glob pattern'))
+  }
+
+  // Extract the folder on the file system from the URL glob
+  const folderUrl = requestUrl.substr(0, requestUrl.length - 1)
+  const folderPath = (await ldp.resourceMapper.mapUrlToFile({ url: folderUrl })).path
 
   const globOptions = {
     noext: true,
@@ -143,7 +151,7 @@ async function globHandler (req, res, next) {
     nodir: true
   }
 
-  glob(filename, globOptions, async (err, matches) => {
+  glob(`${folderPath}*`, globOptions, async (err, matches) => {
     if (err || matches.length === 0) {
       debugGlob('No files matching the pattern')
       return next(error(404, 'No files matching glob pattern'))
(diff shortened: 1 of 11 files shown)


lib/handlers/get.js
```diff
@@ -135,7 +135,7 @@ async function globHandler (req, res, next) {
   const ldp = req.app.locals.ldp
   // TODO: This is a hack, that does not check if the target file exists, as this is quite complex with globbing.
   // TODO: Proper support for this is not implemented, as globbing support might be removed in the future.
-  const filename = ldp.resourceMapper.getFullPath(req)
+  const filename = ldp.resourceMapper.getFilePath(req)
   const requestUri = (await ldp.resourceMapper.mapFileToUrl({ path: filename, hostname: req.hostname })).url
```
RubenVerborgh: So what is happening above seems to be answering the question “what is the full request URL”? Moreover, that question seems to be answered multiple times. Let's create a method for it.

lib/handlers/put.js
```diff
@@ -36,8 +36,9 @@ async function putStream (req, res, next, stream = req) {
 async function putAcl (req, res, next) {
   const ldp = req.app.locals.ldp
   const contentType = req.get('content-type')
-  const path = ldp.resourceMapper.getFullPath(req)
-  const requestUri = await ldp.resourceMapper.resolveUrl(req.hostname, path)
+  const filename = ldp.resourceMapper.getFilePath(req) // @@ Note: weird to go through filename @@
+  const requestUri = (await ldp.resourceMapper.mapFileToUrl({ path: filename, hostname: req.hostname })).url
```
RubenVerborgh: Idem.

lib/resource-mapper.js
```diff
@@ -44,8 +44,8 @@ class ResourceMapper {
   // Maps the request for a given resource and representation format to a server file
   // When searchIndex is true and the URL ends with a '/', and contentType includes 'text/html' indexFilename will be matched.
   async mapUrlToFile ({ url, contentType, createIfNotExists, searchIndex = true }) {
-    let fullPath = this.getFullPath(url)
-    let isIndex = searchIndex && fullPath.endsWith('/')
+    let fullFilePath = this.getFilePath(url)
```
RubenVerborgh: Indeed; let's not have variables named just `path` anywhere in the code.

lib/resource-mapper.js
```diff
@@ -54,21 +54,21 @@ class ResourceMapper {
...
       }
       if (contentType && contentType.includes(this._indexContentType)) {
-        fullPath += this._indexFilename
+        fullFilePath += this._indexFilename
       }
     }
     // Create the path for a new file
     if (createIfNotExists) {
-      path = fullPath
+      path = fullFilePath
```
RubenVerborgh: Probably want to also give `path` a different name here then.

lib/resource-mapper.js
```diff
@@ -54,21 +54,21 @@ class ResourceMapper {
...
       if (searchIndex && this._getContentTypeByExtension(path) !== contentType) {
         path += `$${contentType in extensions ? `.${extensions[contentType][0]}` : '.unknown'}`
       }
     // Determine the path of an existing file
     } else {
       // Read all files in the corresponding folder
-      const filename = fullPath.substr(fullPath.lastIndexOf('/') + 1)
-      const folder = fullPath.substr(0, fullPath.length - filename.length)
+      const filename = fullFilePath.substr(fullFilePath.lastIndexOf('/') + 1)
+      const folder = fullFilePath.substr(0, fullFilePath.length - filename.length)
```
RubenVerborgh: Not a big fan of the arithmetic; we probably just want a regex here.

lib/resource-mapper.js
```diff
@@ -79,7 +79,7 @@ class ResourceMapper {
         if (isIndex) {
           match = ''
         } else {
-          throw new HTTPError(404, `File not found: ${fullPath}`)
+          throw new HTTPError(404, `File not found: ${fullFilePath}`)
```
RubenVerborgh: We probably do not want to expose the internals! Should be the URL.

lib/resource-mapper.js
```diff
@@ -141,13 +141,13 @@ class ResourceMapper {
   }
 
   // Determine the full file path corresponding to a URL
-  getFullPath (url) {
+  getFilePath (url) {
```
RubenVerborgh: I'm going to try to eliminate this method from the public API.

package-lock.json
```diff
@@ -436,15 +436,6 @@
       "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
       "dev": true
     },
-    "JSONStream": {
```
RubenVerborgh: Should not be here.

RubenVerborgh:
@megoth Too much knowledge about the URL/file mapping was leaking in other components, leading to bugs such as #1211, which #643 was created to prevent.

Unit tests run, but have not checked anything else. In particular, you want to look at:
- [allow.js](https://github.com/solid/node-solid-server/pull/1213/files#diff-b11433f1fb19c85e170915d27502d490)
- [globbing](https://github.com/solid/node-solid-server/pull/1213/files#diff-7356ba1cd5bf2e897c784cad5f7b59a7R134) (which now no longer is a hack, but has been reduced in functionality to only folder globs)
- a [skipped test](https://github.com/solid/node-solid-server/pull/1213/files#diff-1ff4cf3c842f350c18fd7d16420f32f9R166) which I think was incorrect

kjetilk:
@RubenVerborgh perhaps you can hit merge on this when you're happy?

lib/resource-mapper.js
```diff
@@ -37,49 +44,87 @@ class ResourceMapper {
     }
   }
 
-  get rootPath () {
-    return this._rootPath
+  // Returns the URL of the given HTTP request
+  getRequestUrl (req) {
```
kjetilk: Allright, but then I don't quite understand the purpose of this PR, as wasn't the intention that it would bring it back to the original interface?

RubenVerborgh: Yes, but this one has been ruled harmless too.

The idea is that this class will be used for all tasks URL and file, and "obtain the full URL" was used in a couple of places, so we better want that done correctly.

RubenVerborgh:
Thanks all!

michielbdejong:
@RubenVerborgh so did https://github.com/solid/node-solid-server/issues/1226 start happening when this PR was merged, or was that a coincidence?

RubenVerborgh:
Yes; some of the knowledge that needed to be inside of `ResourceMapper` was encoded externally, and in a different way. Relying again on `ResourceMapper` brought to light the incorrect mapping.

PR #1215, opened 2019-06-04
Another Sharing Patch

jaxoncreed:
Included in this one:

 - [X] I feel like an idiot for forgetting subdomains AGAIN
 - [X] This fix for Tim's static webId worked on localhost but it looks like it's not working in production 😞  so I'll fix that
 - [ ] Major security patch for `ws`. (Complete but not pushed to npm)
 - [ ] Guess who's back, back again. The infinite loop bug in the mother tracker's back. Tell a friend.


jaxoncreed:
Oh dear this shouldn't have been merged yet

kjetilk:
Uh-oh... But it didn't have WiP-tag, and it looked to me to be fairly self-contained... :-)

PR #1217, opened 2019-06-06
A small error fix

megoth:


kjetilk:
Apparently fixed in #1219 too.

Issue #1226, opened 2019-06-12
Unexpected 409 response on PATCH


michielbdejong:
* open a private browsing window
* visit https://generator.inrupt.com/
* log in with either solid.community (which is on NSS version 4) or inrupt.net (which is on NSS version 5)
* edit something on your profile (for instance your name or your role)
* refresh the page

if you logged in with solid.community, you'll see the change was successful
if you logged in with inrupt.net you'll see the change failed.

In the developer tools you'll see that the PATCH request was responded to with a 409 status, which doesn't make a lot of sense.

jaxoncreed:
Testing now

james-martin-jd:
In the past I've received 409 errors when the value/node to delete during a PATCH can't be found. If I recall, some of the file names changed recently, is it possible the PATCH can't find the file with the delete and so it's throwing 409?

jaxoncreed:
@RubenVerborgh In reference to https://github.com/solid/node-solid-server/commit/111e63ba52e88023f38ef871e147536f38403827#diff-93af968f3e26d51a3f13dc503ce4b347R69

`this._rootpath` is `/Users/m/inrupt/node-solid-server/data` and the file location is `/Users/m/inrupt/node-solid-server/data/jackson.localhost/profile/card$.ttl`

This means that when that line of code runs, there's still a "jackson.localhost" before profile/card which is where that's coming from. Is this how it should work?

jaxoncreed:
For more context when it's 409ing it creates the error `The patch could not be applied. Could not find to delete: <https://jackson.localhost:8443/profile/card#me> <http://xmlns.com/foaf/0.1/name> "jackson"`

RubenVerborgh:
@jaxoncreed So what you mean is that, when calling `mapFileToUrl` with `{ path: '/Users/m/inrupt/node-solid-server/data/jackson.localhost/profile/card$.ttl', hostname: 'jackson.localhost' }`, it gives back `https://jackson.localhost:8443/jackson.localhost/profile/card`? (took that from somewhere else)

That would be incorrect behavior indeed.

Weird that this never popped up; corresponding test is wrong (https://github.com/solid/node-solid-server/blob/b7b5bffd1b94f4976c04c3cf845854efecc81168/test/unit/resource-mapper-test.js#L536) and should be adjusted, then the code can be fixed.

jaxoncreed:
Okay will do

RubenVerborgh:
[...]

PR #1227, opened 2019-06-12
Patch fix

jaxoncreed:
Fixes https://github.com/solid/node-solid-server/issues/1226

### DIFF
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -69,7 +69,6 @@ class ResourceMapper {
       path = path.substring(this._rootPath.length)
     }
     if (this._includeHost) {
-      console.log(path)
       if (!path.startsWith(`/${hostname}`)) {
         throw new Error(`Path must start with hostname (/${hostname})`)
       }
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -65,6 +65,7 @@ class ResourceMapper {
   // Maps a given server file to a URL
   async mapFileToUrl ({ path, hostname }) {
     // Remove the root path if specified
+    path = path.replace(/\\/g, '/')
     if (path.startsWith(this._rootPath)) {
       path = path.substring(this._rootPath.length)
     }
@@ -77,7 +78,7 @@ class ResourceMapper {
 
     // Determine the URL by chopping off everything after the dollar sign
     const pathname = this._removeDollarExtension(path)
-    const url = `${this.resolveUrl(hostname)}${encodeURI(pathname.replace(/\\/g, '/'))}`
+    const url = `${this.resolveUrl(hostname)}${encodeURI(pathname)}`
     return { url, contentType: this._getContentTypeByExtension(path) }
   }
 
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -69,10 +69,10 @@ class ResourceMapper {
       path = path.substring(this._rootPath.length)
     }
     if (this._includeHost) {
-      if (!path.startsWith(`/${hostname}`)) {
+      if (!path.startsWith(`/${hostname}/`)) {
         throw new Error(`Path must start with hostname (/${hostname})`)
       }
-      path = path.substring(`/${hostname}`.length)
+      path = path.substring(hostname.length + 1)
     }
 
     // Determine the URL by chopping off everything after the dollar sign
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -68,6 +68,13 @@ class ResourceMapper {
     if (path.startsWith(this._rootPath)) {
       path = path.substring(this._rootPath.length)
     }
+    if (this._includeHost) {
+      console.log(path)
+      if (!path.startsWith(`/${hostname}`)) {
+        throw new Error(`Path must start with hostname (/${hostname})`)
+      }
+      path = path.substring(`/${hostname}`.length)
+    }
 
     // Determine the URL by chopping off everything after the dollar sign
     const pathname = this._removeDollarExtension(path)
--- a/test/resources/accounts-scenario/alice/profile/card$.ttl
+++ b/test/resources/accounts-scenario/alice/profile/card$.ttl
@@ -1,13 +1,10 @@
 @prefix : <#>.
 @prefix acl: <http://www.w3.org/ns/auth/acl#>.
-@prefix c: <https://localhost:8443/profile/card#>.
 
 :me
-    acl:trustedApp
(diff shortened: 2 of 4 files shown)


lib/resource-mapper.js
```diff
@@ -68,6 +68,13 @@ class ResourceMapper {
     if (path.startsWith(this._rootPath)) {
       path = path.substring(this._rootPath.length)
     }
+    if (this._includeHost) {
+      console.log(path)
```
RubenVerborgh: 😉

jaxoncreed: Sorry about that

lib/resource-mapper.js
```diff
@@ -68,6 +68,12 @@ class ResourceMapper {
     if (path.startsWith(this._rootPath)) {
       path = path.substring(this._rootPath.length)
     }
+    if (this._includeHost) {
+      if (!path.startsWith(`/${hostname}`)) {
```
RubenVerborgh: Needs to start with `/${hostname}/` (otherwise, hostname prefixes would also be allowed).

lib/resource-mapper.js
```diff
@@ -68,6 +68,12 @@ class ResourceMapper {
...
+        throw new Error(`Path must start with hostname (/${hostname})`)
+      }
+      path = path.substring(`/${hostname}`.length)
```
RubenVerborgh: No need to create a new string just for length; let' take `hostname.length + 1`.

lib/resource-mapper.js
```diff
@@ -68,6 +68,12 @@ class ResourceMapper {
     if (path.startsWith(this._rootPath)) {
```
RubenVerborgh: I wonder if we should be concerned about backslashes in paths (Windows) at the start of the function. There is another line in this file dealing with backslashes.

lib/resource-mapper.js
RubenVerborgh: Might want to add a negative test for the slashless case.

test/resources/accounts-strict-origin-off/alice/profile/card$.ttl
```diff
@@ -1,12 +1,11 @@
...
-@prefix c: <https://localhost:7010/profile/card#>.
-@prefix c0: <https://localhost:8443/profile/card#>.
+@prefix c: <https://localhost:8443/profile/card#>.
 
-c:me
+:me
     acl:trustedApp
     [ acl:mode acl:Read, acl:Write; acl:origin <https://app.example.com> ].
-c0:me
+c:me
```
RubenVerborgh: Two different `me`s probably means trouble. Should be both the same, so both `:` / `<#>`.

test/unit/resource-mapper-test.js
```diff
@@ -604,7 +606,7 @@ describe('ResourceMapper', () => {
 
     itMapsFile(mapper, 'a file with the port',
       {
-        path: `${rootPath}space/foo.html`,
+        path: `${rootPath}example.org/space/foo.html`,
         hostname: 'example.org'
       },
       {
```
RubenVerborgh: Does not have a negative test that checks the error when file path does not start with hostname.

RubenVerborgh: We also might want a negative test in the case `includeHost` is false (first block of tests), that it does not strip of any hostname.

lib/resource-mapper.js
jaxoncreed: That's taken care of later in the file

test/unit/resource-mapper-test.js
jaxoncreed: There are negative tests that don't strip off the hostname. I'll add a new one that purposely throws the error.

RubenVerborgh: 👍

lib/resource-mapper.js
RubenVerborgh: Yes, but what if this method is called with the Windows file path `\bla\example.org\foo\bar`?

jaxoncreed:
@RubenVerborgh Everything look good?

test/resources/accounts-scenario/alice/profile/card$.ttl
```diff
@@ -1,12 +1,11 @@
...
-@prefix c: <https://localhost:7000/profile/card#>.
-@prefix c0: <https://localhost:8443/profile/card#>.
+@prefix c: <https://localhost:8443/profile/card#>.
 
-c:me
+:me
     acl:trustedApp
     [ acl:mode acl:Read, acl:Write; acl:origin <https://app.example.com> ].
-c0:me
+c:me
```
RubenVerborgh: I still think these are wrong. Both should be `:me`. Somehow, the RDF serializer is getting the wrong URL.

test/resources/accounts-strict-origin-off/alice/profile/card$.ttl
```diff
@@ -1,14 +1,10 @@
...
-    [ acl:mode acl:Read, acl:Write; acl:origin <https://app.example.com> ].
-c0:me
+:me
     acl:trustedApp
             [
                 acl:mode acl:Append, acl:Control, acl:Read, acl:Write;
                 acl:origin <https://trusted.app>
-            ].
+            ],
+        [ acl:mode acl:Read, acl:Write; acl:origin <https://app.example.com> ].
```
RubenVerborgh: Here it's fine.

test/unit/resource-mapper-test.js
```diff
@@ -604,14 +606,26 @@ describe('ResourceMapper', () => {
...
+
+  describe('A ResourceMapper instance for an HTTPS host with non-default port in a multi-host setup', () => {
+    const mapper = new ResourceMapper({ rootUrl: 'https://localhost:81/', rootPath, includeHost: true })
+
+    it('throws an error when there is an improper file path', (done) => {
+      mapper.mapFileToUrl({
+        path: `${rootPath}example.orgspace/foo.html`,
+        hostname: 'example.org'
+      }).then(() => done('Did not throw error'))
+      .catch(() => done())
```
RubenVerborgh: Nit: you can just write that as

```javascript
    it('throws an error when there is an improper file path', () => {
      return expect(mapper.mapFileToUrl({
        path: `${rootPath}example.orgspace/foo.html`,
        hostname: 'example.org'
      })).to.be.rejectedWith(new Error('error message'))
    })
```

and simultaneously test the error message. (`return` necessary because it's a promise)

jaxoncreed:
@RubenVerborgh can you please approve. We need to get this in asap

RubenVerborgh:
@jaxoncreed All good, excellent work on this!

Issue #1228, opened 2019-06-15
Paths for new resources are incorrectly generated by LDP POST


Otto-AA:
When trying to create a new folder via the solid-filemanager I get a `500 - Path must start with hostname (/otman.inrupt.net)` with the new version on inrupt.net

I've made a POST to https://otman.inrupt.net/public with `slug: my-folder`, `link: <http://www.w3.org/ns/ldp#BasicContainer>; rel="type"` and `Content-Type: text/turtle`.
It responded with the status 500 and the statusText `Path must start with hostname (/otman.inrupt.net)`.

Is this something we should change in the request setup or an error inside NSS?

Here are the complete request headers (except for bearer token and connect.sid cookie):
```
POST /public HTTP/1.1
Host: otman.inrupt.net
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0
Accept: */*
Accept-Language: en,en-US;q=0.7,de;q=0.3
Accept-Encoding: gzip, deflate, br
link: <http://www.w3.org/ns/ldp#BasicContainer>; rel="type"
slug: my-folder
Content-Type: text/turtle
authorization: Bearer  {...}
Origin: https://solid-filemanager.solid.community
Content-Length: 0
DNT: 1
Connection: keep-alive
Referer: https://solid-filemanager.solid.community/public/nightly/?url=https://otman.inrupt.net/public
Cookie: connect.sid=s%3 { ... }
Pragma: no-cache
Cache-Control: no-cache
```

And the response headers:
```
HTTP/1.1 500 Internal Server Error
X-Powered-By: solid-server/5.1.5
Access-Control-Allow-Origin: https://solid-filemanager.solid.community
Vary: Accept, Authorization, Origin
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate
Allow: OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE
Link: <public.acl>; rel="acl", <public.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
Content-Type: text/plain; charset=utf-8
Content-Length: 50
ETag: W/"32-fmcIjz1GGt/H6AvpCi6o3nXP7kw"
[...]

PR #1241, opened 2019-06-21
POST does not error in multiuser mode

jaxoncreed:
Fix for #1228

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -434,11 +434,11 @@ class LDP {
 
   async getAvailableUrl (hostname, containerURI, { slug = uuid.v1(), extension }) {
     let requestUrl = this.resourceMapper.resolveUrl(hostname, containerURI)
-    requestUrl.replace(/\/*$/, '/')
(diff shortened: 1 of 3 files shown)


jaxoncreed:
@RubenVerborgh I took your suggestions into account and refactored

lib/ldp.js
```diff
@@ -432,19 +432,20 @@ class LDP {
     }
   }
 
-  getAvailablePath (host, containerURI, { slug = uuid.v1(), extension }) {
-    const path = slug + extension
+  async getAvailableUrl (hostname, containerURI, { slug = uuid.v1(), extension }) {
+    const fileName = slug + extension
```
RubenVerborgh: This is going to give encoding problems.

Here's what we will need to do. We need to do all operations fully in URL space (expect for the one "does the file exist already?" check, for which we convert to file space).

1. Construct the full URL of the target resource.
2. Map it to a filename with `mapUrlToFile`.
3. Check if the file exists.
4. If it exists, generate a new full URL (with `uuid`)
5. Return the URL.

PR #1252, opened 2019-07-03
fix: check git status before using git version tag

linonetwo:
This fixes the problem described in https://github.com/solid/node-solid-server/pull/1112

Otherwise, it will display version likes `v0.33.0`

### DIFF
--- a/bin/lib/cli.js
+++ b/bin/lib/cli.js
@@ -25,8 +25,9 @@ function getVersion () {
     // Obtain version from git
     const options = { cwd: __dirname, encoding: 'utf8' }
     const { stdout } = spawnSync('git', ['describe', '--tags'], options)
+    const { stdout: gitStatusStdout } = spawnSync('git', ['status'], options)
     const version = stdout.trim()
-    if (version === '') {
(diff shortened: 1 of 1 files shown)


gklyne:
That certainly seems to be the cause of the problem.

But I worry where this "fix" is going if `git describe` behaves differently on different systems (which criticism applies also to my previous fix).  It feels like a lurking problem waiting to bite.

An online search wasn't massively helpful, but did turn up:

- https://stackoverflow.com/questions/2180270/check-if-current-directory-is-a-git-repository
- https://stackoverflow.com/questions/2044574/determine-if-directory-is-under-git-control

On my system, `git rev-parse` exits silently with status `0` in a got repo, or with "fatal: Not a git repository (or any of the parent directories): .git" and status `128` if not.  There seems to be a little consensus behind appropriate `git rev-parse` in this way - I guess because it dosn't actually do much with the repo itself?

linonetwo:
@gklyne  I thought it is fine, because many people on Windows will install "Git for Windows", which behaves just like in Linux. And I tested it in MacOS, it is the same as Linux. So it might be a good enough solution.

Issue #1256, opened 2019-07-06
Unexpected behavior when client omits Content-Type on resource creation


bourgeoa:
@RubenVerborgh

There is a code error in the new ressource-mapper.js 

> line 109    path += `$${contentType in extensions ? `.${extensions[contentType][0]}` : '.unknown'}`

should be (only one $) : 

> line 109    path += `${contentType in extensions ? `.${extensions[contentType][0]}` : '.unknown'}`

The consequences are quite bad for all pods under NSS 5.1.6 (solid.community and dev.inrupt and all others) : 
- files are created with `$.unknown` at the end :
 
-- `ressource test.txt => file test.txt$.unknown` 
-- or `ressource index.html => file index.html$.unknown` (app not working)

- repair of the filesystem is needed, because if you only edit ressource-mapper-js, then when the bad files are edited the 2 above exemples are returned as ressources and files : `test.txt.unknown` and `index.html.unknown` (not usable)

- Strangely the error never appear under the databrowser : ressource creation has the correct extension (tested with the actual ressource-mapper.js and a corrected one)

RubenVerborgh:
The `$` is [by design](https://www.w3.org/DesignIssues/HTTPFilenameMapping.html) and this behavior is explicitly asserted in unit tests.

The `unknown` signals the lack of a `Content-Type`.

bourgeoa:
Ok, I misunderstood that it was indeed by design.

Then it should have been specified and largely commented, so as to check consequences, like when change in webId file was introduced in NSS 5.0.0

I found that it revealed consequences (bugs ) when 'content-type' is not set :
- POST is now buggy :
  - fail with 500 on create file in node (not in browser) (that's why I looked at the filesystem on my test server under docker)
- header returns 'plain/text'
- .html ressource are not understood by the pod web server.
- ...

Work on the new solid-file-client has stopped, because it doesn't pass node tests anymore.

For the above points can you explain where the solution lies, in NSS or in APP like solid-file-client and what to do for the pod web server.

Hope it helps

RubenVerborgh:
Thanks for reporting.
Does everything work when you set a Content-Type (and is anything
preventing you from doing so)? If you rely on extensions, you might want to
use a MIME lookup library.

The previous behavior where the content type was guessed from the extension
was unsafe. Returning text/plain ensures that no weird interpretations are
given to content (but no correct ones either though).

bourgeoa:
Actually solid-file-client is doing the following with POST : 
[...]

Issue #1245, opened 2019-06-29
Default content type is text/plain


jeff-zucker:
## Please describe what you did in reproducible steps
Put foo and foo.ttl with no content-type specified.  In both cases the server serves them as "text/plain"/

## How did it work with 4.x series servers?
Unless I'm mis-remembering, it used to default to "text/turtle" for put of resources.

## What happened when you tried the same with the 5.x series server?
It now defaults to "text/plain".

## Any material that will help, logs, error messages, etc.
```const auth = require('solid-auth-cli')
const base = "https://jeffz.solid.community/public/test/test-file"

async function main(){
  await auth.login()
  let res = await auth.fetch( base, {method:"PUT"} )
  console.log( res.headers.get("content-type") )
  res = await auth.fetch( base + ".ttl", {method:"PUT"} )
  console.log( res.headers.get("content-type") )
}
main()
// output :                                                                     
//                                                                              
// text/plain; charset=utf-8                                                    
// text/plain; charset=utf-8                                                    
```

jeff-zucker:
I believe that 4.x guessed content-type from file extensions.  I can see wanting to move away from that, but I would definitely vote for "text/turtle" as the default content-type for file creation, otherwise lots of apps will break when they expect foo.ttl to be Turtle.

jaxoncreed:
Okay that makes sense. Jeff, would you like to become more involved in maintaining NSS?

jeff-zucker:
On the default content-type.  Since posting this I read this rather authoritative comment: https://github.com/solid/node-solid-server/issues/1165#issuecomment-486229117 in which Tim says PUT without content-type is not valid.  So switching to a 500 on missing content type as POST now does is probably the way to go,  But if we keep a default, it should be Turtle.

On helping with NSS : I'm glad to help anywhere I can be useful.

csarven:
I haven't backtracked this/related issue, .. but the 500 just caught my eye. Why not 415?

jeff-zucker:
Yes, 415 or 400 would be better, as also suggested here: https://github.com/solid/node-solid-server/issues/1246#issuecomment-507361526

kjetilk:
Indeed, it is a client error, so a 4xx is appropriate.

RubenVerborgh:
Just to clarify what is actually happening: the server does _not_ think that files with an unknown content type _are_ text/plain.

Rather:
- the server stores that the content type is unknown (through a `$.unknown` extension on disk)
[...]

PR #1258, opened 2019-07-08
Make application/octet-stream the default Content-Type

RubenVerborgh:
See:
- https://github.com/solid/node-solid-server/issues/1245#issuecomment-509390003
- https://github.com/solid/node-solid-server/issues/1256#issuecomment-509385407

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -247,13 +247,6 @@ function _asyncReadfile (filename) {
  * @return {string} A content type string
  */
 function getContentType (headers) {
-  const headerValue = headers.get ? headers.get('content-type') : headers['content-type']
-
-  // Default content type as stated by RFC 822
-  if (!headerValue) {
-    return 'text/plain'
-  }
-
-  // Remove charset suffix
-  return headerValue.split(';')[0]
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -247,13 +247,6 @@ function _asyncReadfile (filename) {
  * @return {string} A content type string
  */
 function getContentType (headers) {
-  const headerValue = headers.get ? headers.get('content-type') : headers['content-type']
-
-  // Default content type as stated by RFC 822
```
RubenVerborgh: RFC 822 doesn't even contain the string "text/plain" and predated HTTP, so completely bogus.

TallTed: Yes. For completeness' sake, [RFC 1521](https://tools.ietf.org/html/rfc1521) (since obsoleted) would have been the appropriate reference --

> `When a mail reader encounters mail with an unknown Content-type`
> `value, it should generally treat it as equivalent to`
> `"application/octet-stream", as described later in this document.`

michielbdejong:
Thanks!

PR #1260, opened 2019-07-11
Resource-mapper now considers content type when inserting index.html

jaxoncreed:
Fix for #1171 again

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -324,7 +324,7 @@ class LDP {
   async get (options, searchIndex = true) {
     let path, contentType, stats
     try {
-      ({ path, contentType } = await this.resourceMapper.mapUrlToFile({ url: options, searchIndex }))
+      ({ path, contentType } = await this.resourceMapper.mapUrlToFile({ url: options, contentType: options.contentType, searchIndex }))
       stats = await this.stat(path)
     } catch (err) {
       throw error(404, 'Can\'t find file requested: ' + options)
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -122,7 +122,7 @@ class ResourceMapper {
         if (!isIndex) {
           match = files.find(f => this._removeDollarExtension(f) === filename)
         // Check if the index file exists
-        } else if (files.includes(this._indexFilename)) {
+        } else if (files.includes(this._indexFilename) && contentType && contentType.includes(this._indexContentType)) {
           match = this._indexFilename
         }
       }


RubenVerborgh:
How does this address #1171?

jaxoncreed:
Oops mislabeled it. It's #1120

lib/resource-mapper.js
```diff
@@ -122,7 +122,7 @@ class ResourceMapper {
         if (!isIndex) {
           match = files.find(f => this._removeDollarExtension(f) === filename)
         // Check if the index file exists
-        } else if (files.includes(this._indexFilename)) {
+        } else if (files.includes(this._indexFilename) && contentType && contentType.includes(this._indexContentType)) {
```
RubenVerborgh: Hmm. First of all, this can't be correct; the last condition will be true with `text/htmlxyz` whereas we probably want it to fail. So why the choice for `includes` here? Is `contentType` expected to be an `Accept` header? That would not be a correct use of the parameter, which is intended for a single content type. (Namely in the case of a `PUT`, where the client sends a `Content-Type`.)

However, I don't understand the intuition behind this line. What is is supposed to mean? The index file can only be served if the user agent "sends" a content type, that just so happens to be the content type of the index file?

lib/ldp.js
```diff
@@ -324,7 +324,7 @@ class LDP {
   async get (options, searchIndex = true) {
     let path, contentType, stats
     try {
-      ({ path, contentType } = await this.resourceMapper.mapUrlToFile({ url: options, searchIndex }))
+      ({ path, contentType } = await this.resourceMapper.mapUrlToFile({ url: options, contentType: options.contentType, searchIndex }))
```
RubenVerborgh: (See comment below first.)

Is `options.contentType` the value of the `Accept` header? Then it's not a content type, so can't be passed. But in any case, passing the content type to the resource mapper is only meaningful in the case of `PUT` or `POST`, when a new URL needs to be created; the mapper does not do content negotiation.

It seems to me that the desired logic should be implemented at this level, perhaps as follows:
- Set `searchIndex` to `false` if the `Accept` header does not explicitly ask for `text/html`.
- Then call `mapUrlToFile`.

lib/resource-mapper.js
Ryuno-Ki: Perhaps it is, because the contentType can have more values than „text/html” (IIRC a charset parameter or something). That's why an accept match isn't possible.
I had a „starts with” comparison in one of my projects, but a check against MIME types?
https://www.npmjs.com/package/@goa/content-type came after a short search for „RFC 7231”.

RubenVerborgh: I'd be surprised if a `GET` could be reduced to a single content type (it cannot), but indeed, content types can have arguments.

lib/ldp.js
jaxoncreed: Ah okay thanks. I'll do that

jaxoncreed:
Turns out this is much more complicated than setting https://github.com/solid/node-solid-server/blob/master/lib/handlers/get.js#L48 to
```
ret = await ldp.get(options, req.accepts('html'))
```

`searchIndex` is not only used to return HTML files, but is also used to translate /profile/card into /profile/card$.ttl. So there will probably need to be another flag for "return HTML" or something.

I will come back to this one later.

RubenVerborgh:
> `searchIndex` is not only used to return HTML files

We should probably document what exactly it does 🙂
I thought its purpose was to return the `index.html` when present; if switched off, it will only allow exact path matches.

> but is also used to translate /profile/card into /profile/card$.ttl

That is independent of `searchIndex`.

> So there will probably need to be another flag for "return HTML" or something.

We should avoid that; ResourceMapper is content-type-agnostic at the moment.

jaxoncreed:
> but is also used to translate /profile/card into /profile/card$.ttl

I might be interpreting this incorrectly, but I believe that https://github.com/solid/node-solid-server/blob/master/lib/resource-mapper.js#L119 says that in order to find files with the $ extension, search index must be true.

Is this the way it should work?

Issue #1267, opened 2019-07-22
Remove External WebID from account creation


megoth:
Many users do not understand the use of this field - we should move the field to a more specialized form, where we can explain the purpose of the field.

michielbdejong:
Was this resolved?

In general, just a heads-up that we'll switch developer efforts from this repo to the new TypeScript code at https://github.com/inrupt/pod-server soon.

megoth:
This was resolved in https://github.com/solid/node-solid-server/pull/1273

PR #1270, opened 2019-07-22
Removing externalWebId from account creation form

megoth:
Partially solves https://github.com/solid/node-solid-server/issues/1267 - need to implement an API that allows users to configure their externalWebID elsewhere.

We'll discuss this with rest of the team tomorrow. Might discard this PR until we have an API and interface that allows for editing externalWebID.

### DIFF
--- a/default-views/account/register-form.hbs
+++ b/default-views/account/register-form.hbs
@@ -63,14 +63,6 @@
             <span class="help-block">Your email will only be used for account recovery</span>
           </div>
 
-          <div class="form-group">
-            <label class="control-label" for="externalWebId">External WebID:</label>
-            <input type="text" class="form-control" name="externalWebId" id="externalWebId"/>
-            <span class="help-block">
-                We will generate a Web ID when you register, but if you already have a Web ID hosted elsewhere that you'd prefer to use to log in, enter it here
-            </span>
-          </div>
-
           {{#if enforceToc}}
             {{#if tocUri}}
               <div class="checkbox">


timbl:
An alternative to removing it completely would be maybe three things
1-  Hiding it behind an "advanced" section button
2- Explaining exactly what the option is: You have an existing id, and now you just want some more storage for it
3- instead of getting them to type the webid, get them to log in adn grab the webid.  That will save mistypings and misunderstandings

timbl:
The siyte admins also  might not want to enable hosting of pods for existing users.

megoth:
Reg "2- Explaining exactly what the option is: You have an existing id, and now you just want some more storage for it"

This has proven not to work - people simply scan the fields and ignore descriptions.

timbl:
"This has proven not to work"  I don't remember ever seeing a server with a complete explanation.  Is there one?   We have to do ALL three things by the way. 
Are you saying users will never ever understand the concept of having  new pod for their old id, no matter what we write ?  Can you point to the explantion which was given and did not work?

megoth:
Sorry, I should've been more specific; I've been working with forms in my previous work, and all practice shows that many, if not most, people will ignore descriptions as part of a field in a form.

Opt 1 and 3 that you mentioned are things we could try though. I think I'll try opt 1 in a sep branch tomorrow.

michielbdejong:
Can this be closed in favour of #1273?

megoth:
@michielbdejong Yes, I think we'll be going for https://github.com/solid/node-solid-server/pull/1273 instead.

Closing this PR, but might reopen it later in case @timbl disagrees.

PR #1273, opened 2019-07-23
External web id as opt in

megoth:
This gives an alternative fix to https://github.com/solid/node-solid-server/pull/1270 by keeping externalWebID in form, but presenting it as an opt-in part of the form that must be "opened" by checking a checkbox.

I've also labelled the checkbox with *Advanced feature*, which would probably deter most users who don't know what the feature involves. But I've also changed the description a bit, so that those who do want to understand the externalWebID field better might do so.

### DIFF
--- a/default-views/account/register-form.hbs
+++ b/default-views/account/register-form.hbs
@@ -66,7 +66,7 @@
           <div class="checkbox">
             <label>
               <input type="checkbox" name="connectExternalWebId" value="true" id="ConnectExternalWebId" {{#if connectExternalWebId}}checked{{/if}}/>
-              Connect to External WebID (<strong>Advanced feature</strong>
+              Connect to External WebID (<strong>Advanced feature</strong>)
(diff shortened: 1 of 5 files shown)


megoth:
This solves https://github.com/solid/node-solid-server/issues/1267

default-views/account/register-form.hbs
```diff
@@ -114,10 +121,27 @@
 <script>
   var username = document.getElementById('username');
   username.onkeyup = function() {
-  var list = document.getElementsByClassName('editable-username');
-  for (let item of list) {
-    item.innerHTML = username.value.toLowerCase()
+    var list = document.getElementsByClassName('editable-username');
+      for (let item of list) {
+      item.innerHTML = username.value.toLowerCase()
+    }
```
jaxoncreed: I think there might be some accidental indentation here.

Issue #1271, opened 2019-07-22
Redirect users to their Pod after logging in via the server's homepage


megoth:
Today we simply replace the login button with a logout button. Most users would probably expect to be redirected to their Pod.

michielbdejong:
Was this resolved?

In general, just a heads-up that we'll switch developer efforts from this repo to the new TypeScript code at https://github.com/inrupt/pod-server soon.

megoth:
This was resolved in https://github.com/solid/node-solid-server/pull/1274

PR #1274, opened 2019-07-23
Login redirect

megoth:
Offers a solution for https://github.com/solid/node-solid-server/issues/1271 - this has not been tested, and might have unforeseen consequences.

### DIFF
--- a/common/js/auth-buttons.js
+++ b/common/js/auth-buttons.js
@@ -43,7 +43,6 @@
   // Redirect to the registration page
   function register () {
     const registration = new URL('/register', location)
-    registration.searchParams.set('returnToUrl', location)
     location.href = registration
   }
 })(solid)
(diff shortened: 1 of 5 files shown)


common/js/auth-buttons.js
```diff
@@ -30,7 +30,7 @@
         await auth.logout()
       }
       // Now that we have a cookie, reload to display the authenticated page
-      location.reload()
+      location.href = new URL(session.webId).origin
```
jaxoncreed: If I remember correctly, this ended up being overzealous as you would rather have the same page reload rather than redirecting to the origin.

jaxoncreed:
@megoth Was this still something that you'd want merged?

megoth:
No, please don't merge this, it brings with it some really annoying behavior.

Didn't realize it wasn't merged yet, will close it now ^_^

Issue #1275, opened 2019-07-23
Remove the generated index.html for newly created accounts


megoth:
The strategy to generate an index.html for the root has several disadvantages, like keeping it up to date as things change. I propose to remove this "homepage" in advantage of a dashboard pane that the data browser will only show when people navigate to the root of a Pod.

mikeadams1:
I don’t think I agree with this at all! It appears to be in contravention of a pod owner/user capacity to do whatever they want with their homepage root index.HTML, Unless I’m misunderstanding something, it appears you intend to take away the index.HTML in return for some dashboard for the user, is this correct?

megoth:
For posterity, I'll answer @mikeadams1 here as well ([answered earlier on Gitter](https://gitter.im/solid/node-solid-server?at=5d36fe83d61887416422a378)). Please continue this thread here.

Do note that you can still edit an index.html for a Pods root, enabling you to create a custom homepage for that Pod. You will have to create an index.html though, which you can do by using the PUT method. If you want us to enable that programmatically through the data browser, I suggest you create an feature request in the [solid-panes repo](https://github.com/solid/solid-panes/issues/new), as that is where such a feature probably will be implemented.

Vinnl:
> It appears to be in contravention of a pod owner/user capacity to do whatever they want with their homepage root index.HTML, Unless I’m misunderstanding something, it appears you intend to take away the index.HTML in return for some dashboard for the user, is this correct?

I think there's some misunderstanding here. This issue is not about removing the `index.html` that users currently have in their Pods - it's about not placing it there in the first place when someone creates a new Pod. People who already have it in their Pod will keep it, as far as this Issue is concerned.

mikeadams1:
why not have a user.dev.inrupt.net/home page.

Vinnl:
The URLs currently represent data in a user's Pod. There is no data in `/home`.

Do note that what you currently see at `dev.inrupt.net` is not the final state (hence the `dev.`).

mikeadams1:
so essenitaly your stating that you want to create a user dashboard and serve it up on index.html of a user pod, is that correct?

Vinnl:
Not exactly, I think. It basically works the same way it does today: any resource in your Pod that is not an HTML file, will give you the Data Browser when visited with your Web browser. The only difference is that we would like to no longer write an HTML file to people's Pods when they create an account. The functionality currently provided by it can also be provided by the Data Browser.

[...]

PR #1276, opened 2019-07-23
Removing the default index.html for new accounts

megoth:
This solves https://github.com/solid/node-solid-server/issues/1275 .

Before releasing this onto production, we need to make sure that we have a functioning dashboard for the data browser. The latest iteration of what will probably come is in https://github.com/solid/solid-panes/pull/131, but we might change that depending on what is ready before first-coming Thursday.


mikeadams1:
I don’t think I agree with this at all! It appears to be in contravention of a pod owner/user capacity to do whatever they want with their homepage root index.HTML, Unless I’m misunderstanding something, it appears you intend to take away the index.HTML in return for some dashboard for the user, is this correct?

megoth:
Answered in https://github.com/solid/node-solid-server/issues/1275#issuecomment-514191783, please continue discussion there.

PR #1280, opened 2019-08-02
fix 1120 1256 from NSS5.1.6  ressource-mapper

bourgeoa:
@RubenVerborgh @jaxoncreed 
fix issues #1120 #1256 from NSS5.1.6  `ressource-mapper` (should replace PR #1260 )

- ressource-mapper `function mapUrlToFile` :
( it seems that for some reason using `searchIndex = true`  in function definition mapUrlToFile causes problem. searchIndex is not updated when calling directly mapUrlToFile - but updated if called through an other function like ldp.get or get.js ) 
-- createIfNotExists = true : replace isIndex by isFolder (I suppose not needed use of searchIndex = true)
-- createIfNotExists = false : reorganization of tests around `isFolder` and `isIndex` in lieu of `searchIndex` and `isIndex`
- `get.js` modified along PR #1260 (use case not clear to me)
- `ldp.js` modified to avoid double extension when using `slug with extension`

Tested on a pod (created with docker) and solid-Ide.

There is a remaining problem : Pod connexion to root needs login (??  something with allow.js)


lib/resource-mapper.js
```diff
@@ -7,7 +7,7 @@ const HTTPError = require('./http-error')
 
 /*
  * A ResourceMapper maintains the mapping between HTTP URLs and server filenames,
- * following the principles of the “sweet spot” discussed in
+ * following the principles of the “sweet spot� discussed in
```
RubenVerborgh: encoding issue

lib/resource-mapper.js
```diff
@@ -84,28 +84,29 @@ class ResourceMapper {
...
-  async mapUrlToFile ({ url, contentType, createIfNotExists, searchIndex = true }) {
+  async mapUrlToFile ({ url, contentType, createIfNotExists, searchIndex }) {
     // Parse the URL and find the base file path
     const { pathname, hostname } = this._parseUrl(url)
     const filePath = this.resolveFilePath(hostname, decodeURIComponent(pathname))
     if (filePath.indexOf('/..') >= 0) {
       throw new Error('Disallowed /.. segment in URL')
     }
-    let isIndex = searchIndex && filePath.endsWith('/')
+    let isFolder = filePath.endsWith('/')
```
RubenVerborgh: Is this actually an assumption we can make? Or should it be passed in explicitly as a parameter?

lib/resource-mapper.js
```diff
@@ -116,23 +117,15 @@ class ResourceMapper {
...
-          throw new HTTPError(404, `Resource not found: ${pathname}`)
-        }
+      const files = await this._readdir(folder)
+      // Search for files with the same name (disregarding a dollar extension)
+      if (!isFolder) {
+        match = files.find(f => this._removeDollarExtension(f) === filename)
+        if (match === undefined) {throw new HTTPError(404, `Resource not found: ${pathname}`)}
+
+      // Check if the index file exists
+      } else if (isIndex && files.includes(this._indexFilename)) {
```
RubenVerborgh: `isIndex` should be `searchIndex` here

lib/resource-mapper.js
RubenVerborgh: `const`?

lib/resource-mapper.js
```diff
@@ -84,28 +84,29 @@ class ResourceMapper {
...
+  async mapUrlToFile ({ url, contentType, createIfNotExists, searchIndex }) {
     // Parse the URL and find the base file path
     const { pathname, hostname } = this._parseUrl(url)
     const filePath = this.resolveFilePath(hostname, decodeURIComponent(pathname))
     if (filePath.indexOf('/..') >= 0) {
       throw new Error('Disallowed /.. segment in URL')
     }
-    let isIndex = searchIndex && filePath.endsWith('/')
+    let isFolder = filePath.endsWith('/')
+    let isIndex = searchIndex && isFolder
```
RubenVerborgh: I think this variable can be dropped (as per my comment below).

bourgeoa: Not being a coder. I need help.

- L127 OK `searchIndex` and drop of `isIndex` in L95
- L94 do you want to replace `isFolder` by `filePath.endsWith('/') `  everywhere it's used or to use an other variable name or to pass a parameter (this test was already used)
(The code allready used the endsWith('/') I would like to avoid at this stage to add a new parameter and push a test on all `mapUrlToFile` calls)

- I suppose I have to modify the tests that errored N°5 and 6 in travis due to my modifications. It may relates to explicitely set searchIndex value to true.

How can I send my modifications : as a new PR or something else (I have never done it)
Sorry for my deeply need of help

RubenVerborgh: > Not being a coder.

This is good stuff though 👍

L94, okay, leave as is.

> How can I send my modifications

You just push another commit to the branch you have created. So just keep pushing to your own `master` branch, and they will end up here. Don't hesitate to ask if you're uncertain, happy to explain!

bourgeoa: **ressource-mapper tests**
I have a problem with that test : 

>     itMapsUrl(mapper, 'a URL ending with a slash to a folder when index is skipped',
      {
        url: 'http://localhost/space/',
        contentType: 'application/octet-stream',
        createIfNotExists: true,
        searchIndex: false
      },
      {
        path: `${rootPath}space/`,
        contentType: 'application/octet-stream'
      })

Is it a use case ? 
It creates a folder with contentType.
In all tests cases it is the only case where a folder is created. I missed this behaviour.
If yes I need to slightly modify my code and I did not found where it is used (PUT, PATCH, POST)

Can you confirm.

RubenVerborgh: Is tthere no code calling it with `createIfNotExists: true` and `searchIndex: false`? Because this code might be used to create a folder.

bourgeoa:
fix finally Ok.
The tests case where a bit complex to elaborate. I tried to make them clean enough to understand. I reintroduced `isIndex` to make it clearer. I can suppress it if you prefer as it is only used once 
- all `ressource-mapper` tests passed.
- And finally <podroot> opens normally on  <podroot>index.html as expected

lib/resource-mapper.js
```diff
@@ -91,22 +91,28 @@ class ResourceMapper {
...
-        }
-        path += this._indexFilename
-      }
-      // If the extension is not correct for the content type, append the correct extension
-      if (searchIndex && this._getContentTypeByExtension(path) !== contentType) {
-        path += `$${contentType in extensions ? `.${extensions[contentType][0]}` : '.unknown'}`
+      // create a new folder
+      if (!searchIndex && isFolder) {
+  	  // else create a new file
+      } else {
```
RubenVerborgh: `if` statement with empty body is to be avoided

lib/resource-mapper.js
```diff
@@ -91,22 +91,28 @@ class ResourceMapper {
...
+	      // Append index filename if needed
+	      if (isIndex) {
```
RubenVerborgh: Since `isIndex` only appears once (and I find its name confusing), I suggest to replace this by `searchIndex && isFolder`

lib/resource-mapper.js
bourgeoa: I shall replace with

> if ((!searchIndex && isFolder) === false) {

What shall I do with the tests ? Are the errors related to the fix ?

lib/resource-mapper.js
bourgeoa: OK

lib/resource-mapper.js
RubenVerborgh: Or just `if (searchIndex !! !isFolder)`.

Tests are indeed related; some behavior is changed, so should be updated. I think the behavior is correct though.

bourgeoa: I suppose you meant `if (searchIndex || !isFolder)` which do not cover `searchIndex: false` and `isFolder: false` which would allow to create a file without contentType control or read without check it exists.

I think there are missing tests that should cover the case : 

     itMapsUrl(mapper, "a URL with a bogus extension that doesn't match the content type",
      {
        url: 'http://localhost/space/foo.ttl',
        contentType: 'text/html',
        createIfNotExists: true,
        searchIndex: false
      },
      {
        path: `${rootPath}space/foo.ttl$.html`,
        contentType: 'text/html'
      })

and cases without `createIfNotExists`.

bourgeoa:
requested changes made.
fails tests relates to ACL timeout, all other tests passed on node 8.

RubenVerborgh:
But are the timeouts causes by the changes, or where they present before?

lib/resource-mapper.js
RubenVerborgh: > I suppose you meant `if (searchIndex || !isFolder)`

No, I meant that `if ((!searchIndex && isFolder) === false) {` is written more simply as `if (searchIndex || !isFolder)`.

Case looks good.

bourgeoa:
I close the pull request to begin a new one to check travis at every step
I think I bugged something

Issue #1289, opened 2019-08-19
Do not allow root ACL to be deleted


megoth:
> The root container of a user's account MUST have an ACL resource specified.

From https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm

Deleting a root ACL will corrupt the Pod. Deleting the root ACL should only be done as part of removing the Pod altogether.

PR #1290, opened 2019-08-19
This helps the data browser know whether or not a ACL can be deleted

megoth:
Root ACLs should never be deleted. (https://github.com/solid/node-solid-server/issues/1289)

This hack is working with solid-ui to indicate whether a ACL resource is working toward the root container of a Pod. (https://github.com/solid/solid-ui/pull/96)

### DIFF
--- a/default-templates/new-account/.acl
+++ b/default-templates/new-account/.acl
@@ -2,6 +2,9 @@
 @prefix acl: <http://www.w3.org/ns/auth/acl#>.
 @prefix foaf: <http://xmlns.com/foaf/0.1/>.
 
+# Indicates that root is the root of the Pod
+</> a sp:Storage.
+
 # The homepage is readable by the public
 <#public>
     a acl:Authorization;


michielbdejong:
@megoth NSS is now EOL, so we would have to make this change in https://github.com/inrupt/wac-ldp/blob/master/src/lib/authorization/AclManager.ts#L102-L109 instead.

But apart from that, this would create an undocumented link between the databrowser (a Solid app) and the inrupt pod-server (a Solid storage server). If you think all pod servers should have this triple in their root ACL then you should add it to https://github.com/solid/specification first, and then from there it will get added to the test-suite and to the implementations. Does that make sense?

megoth:
> But apart from that, this would create an undocumented link between the databrowser (a Solid app) and the inrupt pod-server (a Solid storage server). If you think all pod servers should have this triple in their root ACL then you should add it to https://github.com/solid/specification first, and then from there it will get added to the test-suite and to the implementations. Does that make sense?

I agree, it's not good to introduce hacky interim-solutions. I've [added an issue](https://github.com/solid/data-interoperability-panel/issues/10) to the [Data Interoperability Panel](https://github.com/solid/data-interoperability-panel/) that will hopefully find some standardized way of allowing this.

megoth:
I'll close this PR unless someone thinks it should be added anyway, albeit being a undocumented hack.

PR #1292, opened 2019-08-22
PR # 1282 with minimized rewrite on ressource-mapper.js

bourgeoa:
@michielbdejong @jaxoncreed @RubenVerborgh 

This PR # 1282 has been rewritten to minimize any change other than strictly needed by the fact that `index.html` is created/checked only if explicitly asked on `get` accept header

- change made by @jaxoncreed on get.js in PR # 1260 + the consequence on resource-mapper.js

-- that introduce the fact that for resource-mapper.js called by get.js the searchIndex is usually false and  resource-mapper.js uses it only in conjonction with a container (no meaning for a file)

 -- the tests on searchIndex needed to be replaced by a check that we have a file (!isFolder) and if we have a folder the we check if we a looking for an index.html file (in the resource-mapper original code the test where made  falsely on searchIndex that was always true by default and not set in get.js)
there is a fake if just for diff indentation (to be deleted)

- resource-mapper.js introduced the obligation to have a contentType that implies that ldp.post added an extension to all slugs including .acl, .meta and were created by ldp.post as .acl.ttl and .meta.acl, and also that any slug containing an extension was creating a file with a double extension (foo.ttl slug gave foo.tll.tll resource). That is the change introduced to ldp.js

Sorry to insist but I hope that these explanations avoid any reasonable doubt to push this PR or the original PR # 1282 and to make a NSSv5.1.8 to dev.inrupt.net and then to solid.community

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -45,7 +45,7 @@ async function handler (req, res, next) {
 
   let ret
   try {
-    ret = await ldp.get(options)
+    ret = await ldp.get(options, req.accepts('html'))
   } catch (err) {
     // use globHandler if magic is detected
     if (err.status === 404 && glob.hasMagic(path)) {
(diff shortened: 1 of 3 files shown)


bourgeoa:
Tests passes on travis except on node 10 as before
tests same as before on my server under NNSv5.1.7+PR
NSSv5.1.7 + PR works as expected with the new mashlib (manuel tests on create/delete/edit) and with solid-ide under solid-file-client.

Issue #1264, opened 2019-07-18
GET on ANY container with an index.html returns HTML, not Turtle


jeff-zucker:
## Please describe what you did in reproducible steps

Performed a GET on https://jeffz.solid.community/ (5.x) and on https://jeffz.solid.openlinksw.com:8444/ (4.x)
## How did it work with 4.x series servers?
The GET returned Turtle

## What happened when you tried the same with the 5.x series server?

The GET returns HTML

dmitrizagidulin:
@jeff-zucker What was the `Accept:` header, on that GET, in each case?

megoth:
I believe this might be connected to a hack we did earlier on NSS - https://github.com/solid/node-solid-server/blob/master/lib/handlers/allow.js#L56

IIRC the hack was there to mitigate the fact that the root ACL was preventing index.html to be served, which made all home pages for pods private by default. This is not a problem for newer pods, but was a problem for older pods (again, IIRC).

jeff-zucker:
I tried without specifying an Accept header (which, should return Turtle anyway, I believe), and also with Accept "text/turtle".  And got HTML in both cases.

mikeadams1:
I agree with @megoth, I remember bringing this problem up but, the fix was not supposed to create this problem.

jeff-zucker:
This looks to be the same issue as https://github.com/solid/node-solid-server/issues/1261

mikeadams1:
I guess my question would be, what has changed in and assess that would cause this to occur? When did it change, And who changed it? Well I guess that’s a few questions that I have.

mikeadams1:
Talk to text on the cell phone really stinks I did not mean to say asses I meant to say NSS 5

jeff-zucker:
I have confirmed this is a problem on all containers, not just the root container.  A GET on .../anyContainer/ returns HTML if there is an index.html file even if Accept "text/turtle" is specified.  This effectively means there is no way to find out the contents of containers that have an index.html.  Which breaks lots of things ...

mikeadams1:
Hopefully it all gets figured out. Thanks for the update Jeff.

fletcher91:
I set up a local server (dev branch) from which I learned it's backed by a literal filesystem, so trying a DELETE to `https://fletcher91.solid.community/index.html` succeeded with a 200 (the home page now has the data browser).

The turtle representation works fine now, but trying to fetch that resource as turtle from my app results in a 403 however (I can't find the acl settings).

[...]

PR #1294, opened 2019-08-24
Fix false response despite accept headers for containers

ludwigschubi:
I think I found a simple fix for #1120 & #1264

please review @jaxoncreed

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -45,7 +45,7 @@ async function handler (req, res, next) {
 
   let ret
   try {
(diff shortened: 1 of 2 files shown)


ludwigschubi:
In case you want to quickly test this you can do a get request to: 

https://ludwig.owntech.de/

My browser shows my root index.html and my rest client gives me the turtle resource at / when sending the Accept: text/turtle header.

ludwigschubi:
Just saw that there are already pull requests to fix these issues. Sorry.

Issue #1291, opened 2019-08-22
Server home page should link to pod when user is logged in.


timbl:
![image](https://user-images.githubusercontent.com/1254848/63527571-5c75a900-c4cf-11e9-9261-34814b71a15e.png)

When you are at the server home page, and you are logged in, instead of saying " If you have not already done so, please create an account.", it should give a big button to link to your pod.  (and maybe also to your profile).

mikeadams1:
And if logged in, and you register another pod, it could act as slave pod.

jaxoncreed:
This should probably not be in the Data Browser project, but in the NSS project.

PR #1296, opened 2019-08-30
Reverting a fix that redirects user to Pod to agressively

megoth:
I tried a quick and dirty fix earlier, but it is way too aggressive to make it useful. I'm reverting the "solution", and will work on a fix for https://github.com/solid/node-solid-server/issues/1291 instead.


michielbdejong:
@megoth we're no longer maintaining this code, let's test this with pod-server instead. I'll help you, ping me tomorrow!

jaxoncreed:
Will be merged with https://github.com/solid/node-solid-server/pull/1311

PR #1311, opened 2019-09-27
Dev

jaxoncreed:


michielbdejong:
Or just merge as-is, that's actually fine as well. :) and then tag a version, deploy to dev.inrupt.net and then plan deploy to inrupt.net and solid.community on Monday 7 Oct.

PR #1313, opened 2019-10-03
Allow apps to access the MS-Author-Via header

### LINKED ISSUE #1314: Solid apps cannot detect that NSS is a Solid server
If I've understood this correctly, there are two ways for a client app to know whether it can update a resource:

- The resource includes the header `Accept-Patch: application/sparql-update`
- The resource includes the header `MS-Author-Via: SPARQL`

As far as I can see, NSS returns the latter, though not the former. This would be fine, except that client-side code running on a different server cannot access that header unless the server also includes it in the CORS header `Access-Control-Expose-Headers: MS-Author-Via`. Currently, NSS does not do that.

Unfortunately, since rdflib 0.20.3, it is now checking for these two headers before updating a Document. Since it cannot find them, it assumes NSS is not a Solid server, and will error out without updating the Document.

Vinnl:
The `MS-Author-Via` header is used by rdflib to detect whether it can send PUT requests.
This used to not be a problem since rdflib would assume it would work, this assumption has now been removed (see https://github.com/linkeddata/rdflib.js/issues/359), so it will fail until it can read this header.

Fixes #1314.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -31,7 +31,7 @@ const corsSettings = cors({
   methods: [
     'OPTIONS', 'HEAD', 'GET', 'PATCH', 'POST', 'PUT', 'DELETE'
   ],
-  exposedHeaders: 'Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate',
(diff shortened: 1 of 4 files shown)


Vinnl:
@jaxoncreed What would be needed to get this merged and released?

jaxoncreed:
I'll merge it in now and it'll be in the next release

Issue #1316, opened 2019-10-07
POST with empty body fails


michielbdejong:
in one window:
curl -k -X POST https://michielbdejong.localhost:8443/public/public-write/

server logs:
solid:handlers POST -- On parent: /public/public-write/ +0ms
solid:handlers POST -- Will create at: https://michielbdejong.localhost:8443/public/public-write/5c9a88d0-e8ef-11e9-b308-a754342b2519 +0ms
solid:server Error page because of: TypeError: Cannot read property 'includes' of undefined
    at LDP.post (/Users/michiel/gh/solid/node-solid-server/lib/ldp.js:161:21) +0ms

dmitrizagidulin:
Out of curiosity, what's the expected behavior of a POST with an empty body? Like, spec-wise?

michielbdejong:
I would have expected it to create an empty document :)

jaxoncreed:
Yep, should fix

csarven:
Assuming that the server assigned the LDPC interaction model on https://michielbdejong.localhost:8443/public/public-write/ , with no `Content-Type` and no payload could result in a `201`. The server could handle the request as if `text/turtle` or `application/octet-stream`. An empty payload is valid for both. What may need more attention is that server is persistent with the interaction model it assigns on the newly created resource.

U *Can* Touch This.

jaxoncreed:
@csarven Could you confirm what should happen in the following scenarios. Is the following true?

 - No content type, no body: 201, empty text/turtle document is made
 - No content type, RDF body: 201, rdf document is made
 - No content type, non-RDF body: 415 unsupported media type
 - Content type something other the RDF, no body: 201, empty file of the given content type is made

csarven:
I'll leave "what should happen" to solid/specification :)

All of the behaviours that you have listed are acceptable.

Requests with no `Content-Type` fall on server's decision to either a) reject the request or b) determine data's type - an implementation detail - and carry on.

Requests with non-RDF `Content-Type` and without payload can result in new resources. Whether an empty resource for a given format is allowed, useful or otherwise is out of scope.

jaxoncreed:
If that's all an implementation detail, then I'd say that rejecting requests without content types seems to be the simplest and most consistent option.

csarven:
https://github.com/solid/specification/issues/70

PR #1320, opened 2019-10-18
Fix empty content type and body problem

jaxoncreed:
Fix for #1316


jaxoncreed:
Right, I can also do that, but after looking at how the code functioned, it turned out to work this way. Based on my interpretation of @csarven 's comment, it is also okay to allow it to create a file. Though, I'll go ahead and have it reject.

jaxoncreed:
Closing this as we should be merging into dev before master.

Continued here: https://github.com/solid/node-solid-server/pull/1323

Issue #1324, opened 2019-10-22
Colons in filenames must be escaped in URIs


timbl:
For example at the moment

`Blog discovery for the future? · Issue #96 · scripting:Scripting-News · GitHub.pdf`

is sent from the server to the client in the directory listing as 

`        <Blog%20discovery%20for%20the%20future?%20%C2%B7%20Issue%20#96%20%C2%B7%20scripting:Scripting-News%20%C2%B7%20GitHub.pdf>,
`

TallTed:
> Colons in filenames must be escaped in URIs

Colons in the first path segment of relative URIs must be percent-escaped (to `%3A`).

Unescaped colons are valid in all subsequent path segments of relative URIs, and in all path segments of absolute URIs.

For the relative URI in the example, the colon must indeed be percent-escaped, so it should be

`<Blog%20discovery%20for%20the%20future?%20%C2%B7%20Issue%20#96%20%C2%B7%20scripting%3AScripting-News%20%C2%B7%20GitHub.pdf>`

However, these would be valid URIs, without the escape -- and escaping the colon in these cases would be considered an error.

`<http://example.com/foo/bar/Blog%20discovery%20for%20the%20future?%20%C2%B7%20Issue%20#96%20%C2%B7%20scripting:Scripting-News%20%C2%B7%20GitHub.pdf>`

`<foo/bar/Blog%20discovery%20for%20the%20future?%20%C2%B7%20Issue%20#96%20%C2%B7%20scripting:Scripting-News%20%C2%B7%20GitHub.pdf>`

timbl:
This should be a relative URI, but because of the unescaped colon it becomes an absolute URI, so it doesn't get the URI of the containing directory appended, which causes other problems down the line.

For an example le test data:  at `https://timbl.com/timbl/Public/Test/Characters/`

timbl:
![Screen Shot 2019-10-22 at 18 05 59](https://user-images.githubusercontent.com/1254848/67338363-f3260c80-f4f6-11e9-87c6-34a2da77701e.png)

PR #1327, opened 2019-10-23
Resource mapper properly encodes special characters in the uri

jaxoncreed:
Fix for #1324


lib/resource-mapper.js
```diff
@@ -78,7 +78,10 @@ class ResourceMapper {
 
     // Determine the URL by chopping off everything after the dollar sign
     const pathname = this._removeDollarExtension(path)
-    const url = `${this.resolveUrl(hostname)}${encodeURI(pathname)}`
+    console.log(pathname)
```
RubenVerborgh: ping

lib/resource-mapper.js
```diff
@@ -78,7 +78,10 @@ class ResourceMapper {
...
+    const url = `${this.resolveUrl(hostname)}${
+      pathname.split('/').map((component) => encodeURIComponent(component)).join('/')
+    }`
```
RubenVerborgh: I would split on the filesystem path separator char [`path.sep`](https://nodejs.org/docs/latest/api/path.html#path_path_sep).

jaxoncreed: Gotcha

jaxoncreed:
Continued in https://github.com/solid/node-solid-server/pull/1337

RubenVerborgh:
Changed base of this PR instead.

PR #1328, opened 2019-10-23
Do not append an identical file extension

### LINKED ISSUE #1318: .mp3 files uploaded with dataBrowser as .mp3$.mp3
see https://github.com/solid/folder-pane/issues/5 from @timbl 

resource-mapper is creating the file `foo.mp3` as `foo.mp3$.mp3`
contentType of .mp3 is audio/mpeg but iana extension for audio/mpeg is mpga.

So the check is a `false wrong match` and resource-mapper is creating foo.mp3$.mp3
I think it should be foo.mp3

It can be generalized so that resource-mapper.mapUrlToFile returns foo.ext and not foo.ext$.ext

bourgeoa:
Fixes #1318

Problem identified in https://github.com/jshttp/mime-types/issues/23

### DIFF
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -33,7 +33,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._types = { ...types, ...overrideTypes }
     this._indexFilename = indexFilename
-    this._indexContentType = this._getContentTypeByExtension(indexFilename)
+    this._indexContentType = this._getContentTypeFromExtension(indexFilename)
 
     // If the host needs to be replaced on every call, pre-split the root URL
     if (includeHost) {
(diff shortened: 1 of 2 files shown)


bourgeoa:
@RubenVerborgh
it solves foo.ext$.ext that should be kept as foo.ext
this will help resolve issue #1318 (see comments in - https://github.com/solid/folder-pane/issues/5)

lib/resource-mapper.js
```diff
@@ -162,6 +162,15 @@ class ResourceMapper {
     return extension && this._types[extension[1].toLowerCase()] || this._defaultContentType
   }
 
+  // hack to fix file.ext$.ext --> file.ext
```
RubenVerborgh: Let's not do hacks in this class that's actually in good shape 😉

jaxoncreed:
Continued in https://github.com/solid/node-solid-server/pull/1336

PR #1333, opened 2019-10-29
Fix tls tests

jaxoncreed:
The tests for WebID TLS didn't actually test NSS. I have updated the tests properly which show that TLS is broken on NSS. Yet, because TLS will soon be depricated I will not fix.


test/integration/acl-tls-test.js
```diff
@@ -50,22 +53,32 @@ var userCredentials = {
   }
 }
 
-describe('ACL with WebID+TLS', function () {
+// TODO Remove skip. TLS is currently broken, but is not a priority to fix
```
michielbdejong: maybe add:
... since the current Solid spec does not require supporting webid-tls on the resource server. The current spec only requires the resource server to support webid-oidc, and it requires the IDP to support webid-tls as a log in method, so that users of a webid-tls client certificate can still use their certificate (and not a username/password pair or other login method) to "bridge" from webid-tls to webid-oidc.

michielbdejong: or something to that extent. Would be even better if you link to a spec issue where this is stated, will post here if i find one.

michielbdejong: ah, found it: https://github.com/solid/solid-spec/pull/171#issuecomment-506350514

michielbdejong:
Impressive work, Jackson! Sorry you did all this work to restore the tests for webid-tls on the resource server, only to find out that it's broken and wont-fix, but it's definitely good to have these tests working, even if they are currently skipped.

PR #1336, opened 2019-11-04
Fix/1318  .mp3 files uploaded with dataBrowser as .mp3$.mp3

### LINKED ISSUE #1318: .mp3 files uploaded with dataBrowser as .mp3$.mp3

jaxoncreed:
Fix #1318

### DIFF
--- a/lib/resource-mapper.js
+++ b/lib/resource-mapper.js
@@ -33,7 +33,7 @@ class ResourceMapper {
     this._defaultContentType = defaultContentType
     this._types = { ...types, ...overrideTypes }
     this._indexFilename = indexFilename
-    this._indexContentType = this._getContentTypeByExtension(indexFilename)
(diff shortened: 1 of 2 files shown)


jaxoncreed:
Continuation of  https://github.com/solid/node-solid-server/pull/1328

RubenVerborgh:
Mmm why?

RubenVerborgh:
Can't we merge #1328 instead?

PR #1337, opened 2019-11-04
Fix/1324

jaxoncreed:


RubenVerborgh:
Changed base of https://github.com/solid/node-solid-server/pull/1327 instead.

Issue #1339, opened 2019-11-05
PUT do not store body for .json files


bourgeoa:
There are different problems related to creation of a .json file with PUT : 
You need a valid JSON body but not only : 
- text body return error 400
- JSON body like "{}" or `{"foo":"test"}` return 200 with a `void body` (filesize = 0) whatever the JSON body content

[...]

PR #1346, opened 2019-11-07
PUT copy body for contentType application/json

bourgeoa:
@jaxoncreed 
move the steam hack inside from before call to ldp.put, to inside ldp.put.
So it is used by POST and PUT.
Checked on a docker server
npm test passed before and after commit

should solve issue #1339


bourgeoa:
@jaxoncreed Solved the conflict

Issue #1361, opened 2019-11-16
"js" container gets mime type


Otto-AA:
When the name of a container is a file extension (e.g. js), the parent of it will list it with a mime-type.

For instance if I make this request: 
```javascript
SolidAuthClient.fetch('https://otman.solid.community/public/', {
  method: 'POST',
  headers: {
    link: '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"',
    slug: 'js',
    'content-type': 'text/turtle' }
})
```
Then `https://otman.solid.community/public/` will contain following:
```text/turtle
js:
    a j:Resource, ldp:BasicContainer, ldp:Container, ldp:Resource;
    terms:modified "2019-11-16T12:48:43Z"^^XML:dateTime;
    st:mtime 1573908523.907;
    st:size 4096.
```
So it says, that the container is a javascript resource. It also works for css and highly likely other file extensions.

jaxoncreed:
Oh, interesting. Good catch. I welcome a pull request for a fix for this, but if not, I will eventually get around to fixing it.

jaxoncreed:
Fixed in #1363

PR #1363, opened 2019-11-19
mime-type named container gets mime-type

bourgeoa:
@jaxoncreed 
should solve issue #1361
checked on a docker NSS 5.2.2 + this fix


jaxoncreed:
Fantastic! Thanks so much for helping out.

Could you write a test for this use case?

bourgeoa:
I checked `ldp-test.js` and there is no example relating to rdf for folder, they all relate to files.
I don't think I can produce a test. I know it is the way to go when you correct a bug.

jaxoncreed:
Okay then. I'll have to manually review it. Thanks so much for your help!

PR #1373, opened 2019-12-10
Updated default node version to 13.2.0

### LINKED ISSUE #1330: NSS needs node.js > 8
We lately tried to install NSS with node.js 8, which resulted in a server which doesn't talk at all on it's public port. In [the docs on the inrupt page](https://solid.inrupt.com/docs/installing-running-nss) it's mentioned in the first sentence that node.js 8 would be sufficient. Maybe it works with 10, but we haven't tried.

I haven't found the sources for that text on github, fell free to move this issue to the appropriate place.

jaxoncreed:
Fix #1330


y0va:
Dunno, but better to put the LTS version here, what would be `v12.13.1` currently?

Issue #1400, opened 2020-01-30
Disallow writing of broken .acl files


jeff-zucker:
If the user attempts to put/post an .acl file using the wrong content-type or if they have a syntax error in the .acl, the user then loses control of the .acl and whatever it points to.  Only a server admin can fix the broken .acl, the user can't edit or delete it.

I think a minimum fix is that the server should cowardly refuse to write an .acl file that has broken syntax or the wrong content-type - allowing users to do this has no benefits and quite terrible consequences - it guarantees that no one except the server admin can access the .acl or the resources it points to.

jeff-zucker:
Looking at https://forum.solidproject.org/t/is-my-pod-dead, you can see that several users have managed to make their entire pods inaccessible because of this problem.

scenaristeur:
An example https://github.com/solid/node-solid-server/issues/1376

jeff-zucker:
This may be too tricky, but the server should also disallow creation of a .acl if it does not grant Control to anyone.  As with broken syntax and wrong content-type, an .acl that does not grant Control to anyone becomes unalterable by anyone but a sysadmin.

kjetilk:
We have a generic issue on this topic in the spec work: https://github.com/solid/specification/issues/67
However, we added RDF validation of ACL files quite some time back, it is surprising if that isn't working anymore.

melvincarvalho:
> Only a server admin can fix the broken .acl, the user can't edit or delete it.

This is not quite correct.  Anyone with access to the file system should be able to edit it or delete.

This should actually be all users.  The idea of a centralized server admin should that we would like to make less common, both in thought process, and in reality

This is a practical point as well as theoretical.  For most of my Pods I will have access to the file system.  Using solid any other way, is for me, a poor mans solid, and IMHO not really giving too much advantage

This can be applied to both the single user and multi user server.  For example you can give each solid user a unix user id and login via SSH.  This has a number of advantages including allowing you to set a file quota.  I do this with node solid server today, and I would hate to use a pod that didnt offer that feature/service at some point.  It would be the equivalent of getting a VPS and only access via FTP

scenaristeur:
>  Anyone with access to the file system should be able to edit it or delete

So who can unlock my POD ?

bourgeoa:
[...]

PR #1407, opened 2020-03-04
reject .acl with bad contentType and related tests

bourgeoa:
issue #1400 relating to avoid being locked out by ACL with bad contentType
This pull rejects .acl files created with POST or PUT when contentType is not `text/turtle`

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -143,6 +143,12 @@ class LDP {
       if (slug.match(/\/|\||:/)) {
         throw error(400, 'The name of new file POSTed may not contain : | or /')
       }
+      // not to break pod ACL must have text/turtle contentType
+      if (slug.endsWith(this.suffixAcl) || extension === this.suffixAcl) {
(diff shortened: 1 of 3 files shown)


bourgeoa:
@RubenVerborgh @kjetilk @jaxoncreed
> This is just on disk. It means: if you see a .acl extension on disk, the thing is Turtle. You can still conneg.

Apart from `text/turtle` are there any conneg that are implemented in the actual NSS for ACL.
The PR can check for all RDF but then the risk is bad content ACL and eventual folder/file access denied.

RubenVerborgh:
> Apart from `text/turtle` are there any conneg that are implemented in the actual NSS for ACL.

Not sure, but I'm fine with restricting it on write.

melvincarvalho:
@kjetilk @michielbdejong @jaxoncreed thanks for looking at this

Unfortunately, this is causing current productions issues in the community pod, on a daily basis

Is there any way this could be expedited?

Or is there some way that we can help?

PR #1410, opened 2020-03-05
delete container with file.acl

bourgeoa:
This allow to delete folder containing orphan file.acl or file.meta
NSS do not do any atomic delete. This should help solve the problem encountered by users not to be able to delete folder where all files have been deleted but not the related file.acl file.meta

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -410,7 +410,7 @@ class LDP {
     }
 
     // Ensure the container is empty (we ignore .meta and .acl)
-    if (list.some(file => file !== this.suffixMeta && file !== this.suffixAcl)) {
+    if (list.some(file => !file.endsWith(this.suffixMeta) && !file.endsWith(this.suffixAcl))) {
       throw error(409, 'Container is not empty')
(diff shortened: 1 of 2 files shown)


jaxoncreed:
Thanks for the PR, but unfortunately recursive delete is something that's still under discussion (https://github.com/solid/specification/issues/132). I'd be happy to add it if it were decided on in the spec, but for now, all we can do is wait.

bourgeoa:
@jaxoncreed
It has nothing to do with recursive. It only authorize folder delete when the folder only contains .acl and does not contain any resource (files) or containers (folders)

kjetilk:
Right, I think this could be reviewed on its own merits, the problem isn't recursive delete, where there isn't a clear consensus, but the lifecycle of an ACL resource, where the [consensus](https://github.com/solid/specification/issues/58#issuecomment-561729455) is that:

>  Deleting a resource must remove associated ACL resource.

So, if it implements this consensus (as it seems to do), I think we're good to merge it.

PR #1414, opened 2020-03-18
Add clarification for configuring an nginx reverse proxy.

tjhowse:


ewingson:
the blockrule for nginx http to https
```
# redirect all http to https
server {
	listen 80 default_server;
	listen [::]:80 default_server;
	server_name _;
	return 301 https://$host$request_uri;
}
```
the exact configuration for reverse proxy is not known by me, tho I' m autodidact

tjhowse:
To my understanding the configuration snippit @ewingson provided will upgrade incoming http connections from the internet to https via a 301 redirect. That's not what this PR is referring to. The `proxy_pass` setting controls how nginx handles routing connections it has already established with an internet client after building the TLS connection.

PR #1415, opened 2020-03-25
delete file.acl when deleting file resource (atomic delete)

bourgeoa:
following emerging consensus and this is an update of pull #1180 and comments from @RubenVerborgh  and @kjetilk

### DIFF
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -392,7 +392,7 @@ class LDP {
     if (stats.isDirectory()) {
       return this.deleteContainer(path)
     } else {
-      return this.deleteResource(path)
+      return this.deleteDocument(path)
     }
   }
 
@@ -422,8 +422,8 @@ class LDP {
     }
   }
 
-  // delete resource with acl and meta links
-  async deleteResource (path) {
+  // delete document (resource with acl and meta links)
+  async deleteDocument (path) {
     const linkPath = this.resourceMapper._removeDollarExtension(path)
     try {
       // first delete file, then links with write permission only (atomic delete)
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -410,7 +410,7 @@ class LDP {
     }
 
     // Ensure the container is empty (we ignore .meta and .acl)
-    if (list.some(file => file !== this.suffixMeta && file !== this.suffixAcl)) {
+    if (list.some(file => !file.endsWith(this.suffixMeta) && !file.endsWith(this.suffixAcl))) {
       throw error(409, 'Container is not empty')
     }
 
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -422,9 +422,21 @@ class LDP {
     }
   }
 
+  // delete resource with acl and meta links
   async deleteResource (path) {
+    const linkPath = this.resourceMapper._removeDollarExtension(path)
     try {
-      return await withLock(path, { mustExist: false }, () => promisify(fs.unlink)(path))
+      // first delete file, then links with write permission only (atomic delete)
+      await withLock(path, { mustExist: false }, () => promisify(fs.unlink)(path))
+
+      const aclPath = `${linkPath}${this.suffixAcl}`
+      if (await promisify(fs.exists)(aclPath)) {
+        await withLock(aclPath, { mustExist: false }, () => promisify(fs.unlink)(aclPath))
+      }
+      const metaPath = `${linkPath}${this.suffixMeta}`
+      if (await promisify(fs.exists)(metaPath)) {
+        await withLock(metaPath, { mustExist: false }, () => promisify(fs.unlink)(metaPath))
+      }
     } catch (err) {
       debug.container('DELETE -- unlink() error: ' + err)
       throw error(err, 'Failed to delete resource')
--- a/test/integration/http-test.js
+++ b/test/integration/http-test.js
@@ -506,9 +506,12 @@ describe('HTTP APIs', function () {
       // Ensure all these are finished before running tests
       return Promise.all([
         rm('/false-file-48484848'),
-//        createTestContainer('delete-test-empty-container'),
-        createTestResource('/delete-test-empty-container/test.txt.acl'),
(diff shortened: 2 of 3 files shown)


bourgeoa:
Sorry for the problems encountered with git. The pull shall be OK now

jaxoncreed:
@RubenVerborgh

lib/ldp.js
```diff
@@ -422,9 +422,21 @@ class LDP {
     }
   }
 
+  // delete resource with acl and meta links
   async deleteResource (path) {
+    const linkPath = this.resourceMapper._removeDollarExtension(path)
     try {
-      return await withLock(path, { mustExist: false }, () => promisify(fs.unlink)(path))
+      // first delete file, then links with write permission only (atomic delete)
+      await withLock(path, { mustExist: false }, () => promisify(fs.unlink)(path))
```
RubenVerborgh, line 430: If this is a folder, it seems that the ACL/meta deletion would fail?
I would expect this statement to come last?

lib/ldp.js
```diff
@@ -422,9 +422,21 @@ class LDP {
...
+
+      const aclPath = `${linkPath}${this.suffixAcl}`
```
RubenVerborgh, line 432: I would actually expect a listing of _all_ files in the folder, and those should be deleted. As per the above check, they should all be `.meta` and `.acl` files. Although, since we didn't lock from that point to here, we don't know that 100% certain, so better to delete all.

bourgeoa: This is already done in `async deleteContainer(directory)`, where `rimraf` is used and deletes all files which are expected to only be `.acl` or `.meta` per the above the check.

lib/ldp.js
bourgeoa: This is never a folder. 

`async delete (url)` is calling : 
- `await deleteContainer (directory)`
- or` await deleteResource (path)`

The PR changes only  `async deleteResource (path)` and `path` is always a `file`

RubenVerborgh: OK, shall we change the name of this method then to be `deleteDocument` so that this is clear?

bourgeoa:
Everything is between line 381-445
I don't how to paste with line numbering

```
async delete (url) {
    // First check if the path points to a valid file
    let path, stats
    try {
      ({ path } = await this.resourceMapper.mapUrlToFile({ url }))
      stats = await this.stat(path)
    } catch (err) {
      throw error(404, "Can't find " + err)
    }

    // If so, delete the directory or file
    if (stats.isDirectory()) {
      return this.deleteContainer(path)
    } else {
      return this.deleteResource(path)
    }
  }

  async deleteContainer (directory) {
    if (directory[ directory.length - 1 ] !== '/') {
      directory += '/'
    }

    // Ensure the container exists
    let list
    try {
      list = await promisify(fs.readdir)(directory)
    } catch (err) {
      throw error(404, 'The container does not exist')
    }

    // Ensure the container is empty (we ignore .meta and .acl)
    if (list.some(file => !file.endsWith(this.suffixMeta) && !file.endsWith(this.suffixAcl))) {
      throw error(409, 'Container is not empty')
    }

    // Delete the directory recursively
    try {
      await promisify(rimraf)(directory)
    } catch (err) {
      throw error(err, 'Failed to delete the container')
    }
  }

  // delete resource with acl and meta links
  async deleteResource (path) {
    const linkPath = this.resourceMapper._removeDollarExtension(path)
    try {
      // first delete file, then links with write permission only (atomic delete)
      await withLock(path, { mustExist: false }, () => promisify(fs.unlink)(path))

      const aclPath = `${linkPath}${this.suffixAcl}`
      if (await promisify(fs.exists)(aclPath)) {
        await withLock(aclPath, { mustExist: false }, () => promisify(fs.unlink)(aclPath))
      }
      const metaPath = `${linkPath}${this.suffixMeta}`
      if (await promisify(fs.exists)(metaPath)) {
        await withLock(metaPath, { mustExist: false }, () => promisify(fs.unlink)(metaPath))
      }
    } catch (err) {
      debug.container('DELETE -- unlink() error: ' + err)
      throw error(err, 'Failed to delete resource')
    }
  }

```

Ryuno-Ki:
> I don't how to paste with line numbering
That's described in https://github.blog/changelog/2019-10-01-multi-line-references-and-comments-beta/

When you then paste the link (`https://github.com/solid/node-solid-server/blob/9e45888f223b470ac143ef091f6ecdcb6b3a55d5/lib/ldp.js#L381-L423`), GitHub will turn it into:

https://github.com/solid/node-solid-server/blob/9e45888f223b470ac143ef091f6ecdcb6b3a55d5/lib/ldp.js#L381-L423

lib/ldp.js
bourgeoa: May be a good idea `document` can be understood as `resource and links`.

bourgeoa:
@RubenVerborgh @jaxoncreed
I renamed `deleteResource()` to `deleteDocument()` 

One test fails and only on `node.js/lts` (the 3 others are OK), it is not related to `delete`, but `POST (multipart)`
```
1 failing
  1) HTTP APIs
       POST (multipart)
         should create as many files as the ones passed in multipart:
     Error: Either the size (remote/local) don't match or files are not stored
      at Test.<anonymous> (test/integration/http-test.js:853:27)
      at Test.assert (node_modules/supertest/lib/test.js:181:6)
      at Server.localAssert (node_modules/supertest/lib/test.js:131:12)
      at emitCloseNT (net.js:1653:8)
      at processTicksAndRejections (internal/process/task_queues.js:83:21)
```

bourgeoa:
@jaxoncreed @RubenVerborgh @kjetilk 
I saw new version 5.2.4 

This PR having been reviewed and amended I thought he could have been included. It is linked to PR #1410 which has been merged. This bug caused a lot of harm

RubenVerborgh:
> I thought he could have been included.

There's things such as urgent security patches 🙂
I'm sure @jaxoncreed will put it in ASAP.

Issue #1418, opened 2020-04-09
Users with only acl:Append permission are able to POST ACL files


angelo-v:
## Actual behaviour

Given a container where I have `acl:Append` permission, but no `acl:Control` permission
When I POST a file to the container using a Slug header ending with `.acl`
Then the resource gets created and is considered as a valid ACL

## Expected behaviour

Creating ACL resources is only possible with `acl:Control` permission

## Example

Container at path `/foo` with public write permission containing a resource `bar.ttl`

Request to create an ACL file for `bar.ttl`: 

```
POST /foo HTTP/1.1
Content-Type: text/turtle
Link: <http://www.w3.org/ns/ldp#Resource>; rel="type"
Slug: bar.ttl.acl

# ACL content
```

csarven:
It is only "considered as a valid ACL" because of the automatic relation with a URI template that's in use (which happens to match the created resource) ie. id acl id.acl.

If bar.ttl exists and has rel=acl to bar.ttl.acl, I agree that this is an NSS bug and it should reject the request to create bar.ttl.acl from user with acl:Append - if it didn't do anything about the Slug suggestion. If it did, then it'll just rename to something "safe".

If however bar.ttl doesn't exist, bar.ttl.acl should be handled as any RDF source - not to be "considered as a valid ACL" unless of course if NSS reserves .acl suffix in resource names for "valid ACL"s.

justinwb:
> If however bar.ttl doesn't exist, bar.ttl.acl should be handled as any RDF source - not to be "considered as a valid ACL" unless of course if NSS reserves .acl suffix in resource names for "valid ACL"s.

I don't think it _should_ be handled as any RDF source. The association of a given resource R to an acl resource A in NSS is based on the presence of '.acl' at the end of a A's name, when the part before the .acl in A's name is the same name as another resource. If A's name is just '.acl', it would apply to the container that A is a member of.

Consider the creation of a resource named 'foo' in container C. If NSS doesn't ensure that there isn't a foo.acl in C first, anyone that can predict that a file named foo will be created at some point can set the permissions in advance by making foo.acl, even if they don't have control permission on container C.

IMO - the simplest path would be for NSS to reserve the .acl suffix to avoid this. Note this is an NSS problem, not a Solid problem. In Solid you can make any resource names you want, and the server maintains the association of associated acls. It is _how_ NSS maintains that association that leads to this potential issue.

csarven:
>..**unless**..
[...]

PR #1419, opened 2020-04-12
Post acl fix

jaxoncreed:


csarven:
Great!

Closes #1418 . Thanks @angelo-v for raising the issue.

melvincarvalho:
Was this change reviewed?

We are having issues in production since the upgrade.

test/integration/acl-oidc-test.js
```diff
@@ -545,6 +545,16 @@ describe('ACL with WebID+OIDC over HTTP', function () {
         done()
       })
     })
+    it('user2 should not be able able to post an acl file', function (done) {
```
angelo-v, line 548: This test should actuallly be like this:
(do a POST with a Slug Header, not PUT)

```javascript
it('user2 should not be able able to post an acl file', function (done) {
      var options = createOptions('/append-acl/', 'user2', 'text/turtle')
      options.headers.Slug = 'some.acl'
      options.body = '<a> <b> <c> .\n'
      request.post(options, function (error, response, body) {
        assert.equal(error, null)
        assert.equal(response.statusCode, 403)
        assert.equal(response.statusMessage, 'User Unauthorized')
        done()
      })
    })
```

PR #1422, opened 2020-04-15
Allowed acls to be created via slugs if there is control access

jaxoncreed:


jeff-zucker:
It's not solid-file-client.  I get a 500 error on 5.2.4 with either a plain vanilla solid.auth POST or a PUT of a file with a .acl extension regardless of the permissions or name of the container when identical requests succeed in 5.2.3.  I'm afraid that I can't tell from looking at your changes if they fix this.  I'm glad to test more.

jeff-zucker:
Yes, I believe this will solve the issue, though I didn't install the server to test.  Can you clarify something for me - what is the expected behavior of a slug on PUT?  In Solid-File-Client we are sending one with PUT, should we stop doing that?

RubenVerborgh:
@jeff-zucker: with PUT, don’t use a slug. PUT directly to the URL where you want the thing to be.

bourgeoa:
Solves the regression from 5.2.4
Tested with PUT and POST under my server : node-solid-server on docker instance running current master + PR #1422 + PR #1415 (atomic delete) and PR #1408 (update dockerFile to node LTS))

csarven:
>Can you clarify something for me - what is the expected behavior of a slug on PUT? In Solid-File-Client we are sending one with PUT, should we stop doing that?

RFC defines Slug and used alongside POST. Slug in PUT is not meaningful. Server should ignore Slug if it encounters it in PUT (and PATCH for that matter).

jeff-zucker:
Thanks, that was my understand too.  Recent versions of Solid-file-client have a slug in the PUT due to a cut-and-paste error.  We're removing it.

angelo-v:
Hi jaxon thanks for the fix. Can you add my two tests here an ensure they are green?

https://github.com/angelo-v/node-solid-server/pull/2/files#diff-5ec9298c6982e0d2454df58b243cc5dd

bourgeoa:
@angelo-v 
Just added your 2 tests : they are not passing

```
  1) ACL with WebID+OIDC over HTTP
       Append-only
         user1 (owner) can use POST to append an acl:
     Uncaught AssertionError: expected 403 to equal 201
      at Request._callback (test/integration/acl-oidc-test.js:563:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
      at endReadableNT (_stream_readable.js:1184:12)
      at processTicksAndRejections (internal/process/task_queues.js:80:21)

  2) ACL with WebID+OIDC over HTTP
       Append-only
         user2 (with append permission) cannot use POST to append an acl:
     Uncaught AssertionError: expected 'Forbidden' to equal 'User Unauthorized'
      at Request._callback (test/integration/acl-oidc-test.js:575:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
      at endReadableNT (_stream_readable.js:1184:12)
      at processTicksAndRejections (internal/process/task_queues.js:80:21)

```

jaxoncreed:
@angelo-v @bourgeoa Yep. I've pulled it down and confirmed they aren't working. I'm working on a solution now.

angelo-v:
oh, sorry @jaxoncreed might be false alert, I missed pointing out that I also added a `.acl` to the test resources here: https://github.com/angelo-v/node-solid-server/pull/2/files#diff-0dea854414621609e90e10eddda0daa5

This is needed to set the correct preconditions

The wording 'Forbidden' vs 'User Unauthorized' does not seam to matter much, maybe we can change the expectation, not sure were I got mine from.

jaxoncreed:
@angelo-v There were a few problems with the test, but they illuminated other problems with our implementation that I'm currently working through. So, I'll get back to ya'll with the solution 😄

jaxoncreed:
Hello everyone. Thank you so much for your participation in this spec issue as it really helped raise an interesting security consideration, which also is related to interoperability. 

We have after much consideration concluded the best approach is for [NSS v5.2.4](https://github.com/solid/node-solid-server/pull/1419) to remain patched and move forward. 

NSS no longer allows ACLs created with POST and a Slug header. Given the options, we found it makes the most sense to proceed with this change because it helps ensure data safety as well as future interoperability. This decision took some time to arrive because we wanted to investigate exactly why the new NSS version will break some applications ([see conversation](https://github.com/solid/node-solid-server/issues/1418#issuecomment-613144819) about the @scenaristeur application) and whether that could be avoided.  

In this case there were three issues. The first arose when an application assumed that an ACL to describe access control for a resource used that resource name and appended ‘.acl’ to the filename. This is an expectation of specific behavior for a NSS implementation version and should not be expected to work with any other Solid server implementation. 

The second issue was where an application assumed a server would use a slug with POST in an expected way to name the new resource. However, the [specification](https://www.w3.org/TR/ldp/#h-ldpc-post-slug) makes slug use optional and this also should not be expected to work with any other Solid server implementation.

Third and finally, we looked at how the server uses a header to inform the client of the URI for the associated ACL when a resource is created. An application could use PUT or PATCH to make changes to an ACL or even create an ACL if it does not yet exist (as would happen when an inherited ACL is being used). However, if an application instead used POST to create an ACL then it is based on assumptions about the structure of the URI, and depends on the slug being accepted by the server. The outcome of a POST request could be a normal resource using the slug as part of URI, or could be a normal resource where a slug is ignored or not used. Again this should not be expected to work with any other Solid server implementation.

Applications that depend upon knowing very specific implementation details unique to one Solid server could introduce serious compatibility risks when expected to work with another Solid server. Also ACLs are fundamental to trust in the Solid server security model and only should be created using a safe URI provided by the server. Therefore given these three issues that impact interoperability and security, the best and safest way forward with NSS was decided to be disabling creation of ACL with POST using a slug. 

The issue @scenaristeur created has been updated, and includes suggestions on ways to create ACLs [here](https://github.com/solid/node-solid-server/issues/1418#issuecomment-618049602).

This issue also further supported and clarified the spec requirement where agents need to have Control access privilege on the resource in order to create its ACL, as well as ensuring that a resource exists prior to the directly associated ACL resource. There are additional checks being considered to ensure a client’s intention to create an ACL resource as well as enhanced server validation being considered. If you’re interested in discussing methods of ACL creation, @csarven is continuing the conversation [here](https://github.com/solid/specification/issues/42#issuecomment-616688848).

michielbdejong:
@jaxoncreed so we can close this PR, then?

michielbdejong:
I think we don't want to add the option of creating ACL docs through POST+Slug, so I'll descope and close this. Please reopen if I misunderstood and/or if you still want this PR to be merged.

PR #1454, opened 2020-07-24
Update dependencies

michielbdejong:


michielbdejong:
500 on register:
```sh
curl 'https://localhost:8443/register' \
   -H 'content-type: application/json' \
   -H 'Origin: https://localhost:8443' \
   -H 'Referer: https://localhost:8443/common/popup.html' \
   --data-binary '{"issuer":"https://localhost:8443/","grant_types":["implicit"],"redirect_uris":["https://localhost:8443/.well-known/solid/login"],"response_types":["id_token token"],"scope":"openid profile"}' \
   --insecure

michielbdejong:
This is fixed by adding isomorphic-webcrypto as a workspace in mashlib-dev. This error also makes the tests of oidc-op fail. Basically, oidc-op requires isomorphic-webcrypto to create a key, then jose requires another instance of isomorphic-webcrypto to use the key. Before use, isomorphic-webcrypto checks whether the key is an instance of its own CryptoKey class, and that check then fails. Maybe the solution is if jose re-exports isomorphic-webcrypto, so all code uses the same instance.

michielbdejong:
Now looking into https://github.com/solid/node-solid-server/issues/1455

michielbdejong:
As discussed in https://solidos.solid.community/Team/SolidOs%20team%20chat/index.ttl#this

Issue #1460, opened 2020-07-31
Replace locking with proper in-memory RW locking


timbl:
The current locking system has many problems, which impact the performance of the server and other things.

- A file-based locking system is used,  -- bug [#1397](https://github.com/solid/node-solid-server/issues/1372)
 - There is only one lock for read and write operations

The last problem means that people cannot simultaneously read files, even when no writing is happening.  If two people want to watch the same movie, they have to do it one after another!!

A locking system need to have the following properties:

- It is memory based
- It allows concurrent reads of the same file
- It only allows one request at a time to write the file
- While a request has write access to the file, no others can **read or write** it.

This a relatively common problem

NoelDeMartin:
I think this is also the cause of some problems [I've been discussing in the solid forum](https://forum.solidproject.org/t/state-of-the-art-for-querying-large-containers/3320).

In particular, making a GET request to a container causes the `purl:modified` and `posix:mtime` properties to be modified. This is annoying because I'm using them to invalidate my application cache, and given that reads change them the cache becomes useless.

bourgeoa:
I googled around and propose to try to build a lock system around : npm node-cache (https://www.npmjs.com/package/node-cache)and async-lock (https://www.npmjs.com/package/node-cache)
- async-lock will be used to create/remove the key (file path) when some write mode is entered/ended
- and node-cache to retain all keys used in write mode
- when the key exists in cache no one else can read or write.

@timbl is it what you are expecting ?

Can anyone help to elaborate a bit more or propose an other strategy to solve this bug ? @RubenVerborgh @michielbdejong @jaxoncreed @jeff-zucker @Vinnl

I found lock.js in NSS to contain the actual lock/unlock mechanism (it uses proper-lockfile)

RubenVerborgh:
Solved by https://github.com/solid/node-solid-server/pull/1475

PR #1464, opened 2020-08-12
in memory lock

bourgeoa:
This is first part of issue #1460 replacing file locking by in memory locking.

Can reviewers give hints on what is expected for the rest.
I was thinking of having a dual in memory lock for Read and Write (googling 2 semaphore ?) : 
options.mustExist = false for Write so I propose : 
- when Read to have in memory cache ( shall we check for a size limit) @timbl is it what you expected
- when Write the actual solution will do

### DIFF
--- a/lib/cache-provider.js
+++ b/lib/cache-provider.js
@@ -1,0 +1,12 @@
+const NodeCache = require('node-cache')
+let cache = null
+
+exports.start = function (done) {
+  if (cache) return done()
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
Proposing https://github.com/solid/node-solid-server/pull/1475 instead, as it relies on an existing, well-tested locking mechanism.

Issue #1443, opened 2020-07-09
Performance loss when using webId from a different server


bourgeoa:
see issue #1439 
@megoth writed `The reduced performance might be related to linkeddata/rdflib.js#419`

I made some more tests adding webId (acl:agent) to acl rule (so that a rule as more than one webId - acl:agent) 
- the performance loss arises allready in https://solid.community running on NSS 5.2.3 (I don't know the rdflib version)
- there is no performance loss if all acl:agent are on the same server

I practice it is not usable. It takes more than one minute, and some time timeout on reading acl doc.
It does not seem to be be related to mashlib (to be confirmed) because I also have the problem using solid-ide.

michielbdejong:
I'm also seeing a huge delay in test/integration/header-test.js, I'm having to set the mocha timeout to 10s there now in the update-deps branch (added a FIXME comment that links to this issue).

michielbdejong:
Confirmed that when we removed all external webid's from https://solidos.solid.community/.acl, https://solidos.solid.community/ became like a hundred times faster.

timbl:
The external webids are [i think] only damaging when they have Control access.   See operational issue https://gitlab.com/solid.community/support/-/issues/15

michielbdejong:
I reproduce this locally in my mashlib-dev now:
```sh
[...]

PR #1467, opened 2020-08-19
Add temporary cache in lib/acl-checker.js

michielbdejong:


michielbdejong:
Cache ACL documents and profiles of owners therein for 5 minutes.
Brings the test I did in https://github.com/solid/node-solid-server/issues/1443#issuecomment-675928820 down from 1 minute to 1 second.

angelo-v:
That means, if the cache is not hit, it still takes 5 minutes?

michielbdejong:
> That means, if the cache is not hit, it still takes 5 minutes?

No, if the cache is not hit, it will still fetch the document only once instead of 18 times, so it will still be something like 18 times faster.
If the cache is hit it will be up to 1000 times faster, depending on the ratio between CPU and network speed.

michielbdejong:
> where to declare this server-wide variable.

Yes, good question. It doesn't need to be a singleton though, it doesn't matter if it exists twice, once per execution thread, or even once per http request, just that the less unique it is, the less effective it is. Having it as an in-memory global in the lib/acl-checker.js file feels ok to me now, we can always move it around if our insights change.

> Looks good to me. Lots of other cache strategies are possible, but let's put this in now to get over the performance problem we have now.

Thanks!

PR #1473, opened 2020-09-03
Fix Docker instructions in the main README 

vemonet:
Remove outdated instructions about running using Docker from README.md, and redirect the user to the docker-images/README documentation page

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ $ solid start --help
     --support-email [value]       The support email you provide for your users (not required)
     -q, --quiet                   Do not print the logs to console
     -h, --help                    output usage information
- ```
+```
 
 Instead of using flags, these same options can also be configured via environment variables taking the form of `SOLID_` followed by the `SNAKE_CASE` of the flag. For example `--api-apps` can be set via the `SOLID_API_APPS`environment variable, and `--serverUri` can be set with `SOLID_SERVER_URI`.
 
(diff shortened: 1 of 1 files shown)


angelo-v (approved):

bourgeoa:
There are 2 ways to use docker :
- with the docker-image using some env parameters
- using the Dockerfile and a config.json with extensive parameters

I personally only use the Dockerfile which can also be used to test NSS modifications.

The documentation should reflect that, and not only the docker-image way.

angelo-v:
You could mount a config.json to the production image as well. The dockerfile is good for development while the image is production ready.

bourgeoa:
I don't think the documentation relates only to production ready server. That's why both approach should be maintained.

angelo-v:
You are right @bourgeoa the readme currently mixes both and we should reflect that for docker as well, just separate both variants clearly by subsections.

README.md
```diff
@@ -228,47 +228,8 @@ Configuring Solid via the config file can be a concise and convenient method and
...
-
-```bash
-docker pull nodesolidserver/node-solid-server
-```
-
-Run with:
-```bash
-docker run -p 8443:8443 --name solid nodesolidserver/node-solid-server
-```
+See the [documentation to run Solid using docker and docker-compose](https://github.com/solid/node-solid-server/tree/master/docker-image).
```
angelo-v, line 231:
```suggestion

### Production usage

See the [documentation to run Solid using docker and docker-compose](https://github.com/solid/node-solid-server/tree/master/docker-image).
```

README.md
```diff
@@ -228,47 +228,8 @@ Configuring Solid via the config file can be a concise and convenient method and
 
 ## Use Docker
 
-Pull with:
-
```
angelo-v, line 232: Let's restore this section under the headline `development usage`

README.md
```diff
@@ -228,47 +228,8 @@ Configuring Solid via the config file can be a concise and convenient method and
...
-```bash
-docker pull nodesolidserver/node-solid-server
```
angelo-v, line 234: Instead of pulling the prod image this should instruct to build the Dockerfile

angelo-v (changes requested):

vemonet:
Thanks for the feedback @angelo-v 

@bourgeoa I put back the [previous documentation](https://github.com/vemonet/node-solid-server/tree/fix/docker-docs#development-usage) describing the use of the Dockerfile at the root of the repository under the `Development usage` section

bourgeoa:
OK for me

angelo-v (commented):
Nice!

PR #1475, opened 2020-09-07
Replace on-disk lock by in-memory lock

### LINKED ISSUE #1372: Stop creating lock files in the filestore for each read operation
The server currently, to read a file,  creates a lock file (like /`favicon.ico.lock` directory) reads the file,(like `/favicon.ico`) and then delets the lock file.

This is a bad idea for more than reason. 
- The server won't run on a read-only file system.  
- Files can't be read if the server user isn't allowed access to the the directory above them ( eg read access to say /favicon.ico needs write access to / )
- The file system, depending on its technology, may wear down the device, reduce its life.
- Much the speedup one gets from layers of caching is pointless as the server has to do the round tripe s with the file system itself.

Suggested fixes: Both:

- Switch the file-based lock to an in-memory JS  variable lock.
- Check the access of a file, and if no write access is allowed by anyone, then assume file will not change. In that case, remove the lock completely on read, so many threads can stream in parallel .

RubenVerborgh:
### Background

Prior to https://github.com/solid/node-solid-server/pull/1116, two simultaneous `PATCH` requests could corrupt files. The solution in https://github.com/solid/node-solid-server/pull/1116 was built with the design goals:
- to function in a multi-worker environment
- to not require external services (like Redis)

The file-based locking system—although meeting the above two goals—did cause some annoyances (notably https://github.com/solid/node-solid-server/issues/1372).

### What this PR does

This PR switches the file-based locking to in-memory locking.

Closes #1372
Obsoletes #1464

#### Advantages
- no more gratuitous writes to disk
- uses an external library that is well-tested
- still no dependencies on external services

#### Disadvantages
- does not function with multiple workers or threads
  - …but neither do a a lot of other parts of NSS, so multi-worker support is future work anyway

### What this PR does not
- read/write locking (https://github.com/solid/node-solid-server/issues/1460)
  - so still can't watch videos with multiple people (unless you use HTTP ranges)
  - we would need to implement https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock for that
  - …but I'm not touching that myself until NSS has a layered system where reads and writes happen in a single place, because they are now all over the handlers, which would make building and testing a nightmare

### DIFF
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -64,7 +64,7 @@ async function patchHandler (req, res, next) {
     }
 
     // Patch the graph and write it back to the file
-    const result = await withLock(path, { mustExist: false }, async () => {
+    const result = await withLock(path, async () => {
       const graph = await readGraph(resource)
       await applyPatch(patchObject, graph, url)
       return writeGraph(graph, resource, ldp.resourceMapper.resolveFilePath(req.hostname), ldp.serverUri)
--- a/lib/ldp.js
+++ b/lib/ldp.js
@@ -251,7 +251,7 @@ class LDP {
     }
 
     // Directory created, now write the file
-    return withLock(path, { mustExist: false }, () => new Promise((resolve, reject) => {
+    return withLock(path, () => new Promise((resolve, reject) => {
       // HACK: the middleware in webid-oidc.js uses body-parser, thus ending the stream of data
       // for JSON bodies. So, the stream needs to be reset
       if (contentType.includes('application/json')) {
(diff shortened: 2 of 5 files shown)


michielbdejong:
Nice! Is it covered by a unit test?

@bourgeoa what do you think? Especially given your remark in https://github.com/solid/node-solid-server/pull/1464/files#diff-42068aebd9eb6d9326293ef88bb61b3dR32 - why did you discard`async-lock` originally?

I don't have a strong preference either way, as long as it works :) Eventually I want to implement a proper caching layer with write-through so we will no longer need locking (writes to the cache will be synchronous).

RubenVerborgh:
> Nice! Is it covered by a unit test?

Yes, all the LDP logic covers it.

bourgeoa:
@michielbdejong 
I did find this solution and thought it could be used as a read cache for resources with text/turtle contentType, if the idea was wellcomed.

RubenVerborgh:
It's probably too early for a cache, before we have the single read/write layer. Plus, we still need to solve the locking then with regard to disk writing.

michielbdejong (approved):

RubenVerborgh:
I think your SDD will be very happy, @timbl.

PR #1480, opened 2020-09-27
Fix section hierarchy in README.

Gittenburg:
The email section belongs under the multi-user section
and not under the reverse proxy section.


angelo-v (approved):
good catch

PR #1484, opened 2020-10-05
Check read access in websocket-pubsub

michielbdejong:
Requires https://github.com/solid/node-solid-ws/pull/22


michielbdejong:
Interestingly I can't find a function that parses a given `Authorization` and `DPop` header combination.
I see `getPermissionsFor` in lib/header.js but it uses `session.userId`.
I can't see where `session.userId` gets set from the request headers.

michielbdejong:
Hm, could it be related to https://github.com/solid/node-solid-server/blob/master/lib/api/authn/webid-oidc.js#L121? Maybe that sets the userId in the express session?

PR #1494, opened 2020-10-20
Fix patch mkdir -p

### LINKED ISSUE #1493: Failing CRUD test: Create non-container › in an non-existing container › using PATCH › emits websockets-pubsub on the resource

michielbdejong:
I can't reproduce it in isolation, but apparently there is a glitch that makes `fx.mkdirSync` fail. Using `fx.mkdir` instead fixes #1493.


michielbdejong:
Ah, reproduced it in isolation now, see https://github.com/hex7c0/mkdir-recursive/issues/7

Issue #1491, opened 2020-10-15
Path must start with hostname


highlevellogic:
I just got a pod server set-up - v5.5.2 on Windows 10, Node v12.18.3
When the registration processed is finished, it automatically sends me to my page.
(from https://solid.example.com/register to https://rogerfgay.solid.example.com/) // fake domain
Then I get the error message in the browser and nothing else.

Path must start with hostname (/rogerfgay.solid.example.com)

config: "root": "C:\\programs\\solid\\data"
directory C:\programs\solid\data\rogerfgay.solid.example.com exists and has a lot of normal looking stuff in it

Server-side:
solid:get / on rogerfgay.solid.example.com +0ms
solid:get GET -- Error: undefined Path must start with hostname (/rogerfgay.solid.example.com) +1ms
solid:server Error page because of: Error: Path must start with hostname (/rogerfgay.solid.example.com)
at ResourceMapper.mapFileToUrl (C:\Users\roger\AppData\Roaming\npm\node_modules\solid-server\lib\resource-mapper.js:77:15)
at LDP.get (C:\Users\roger\AppData\Roaming\npm\node_modules\solid-server\lib\ldp.js:358:10)

note:
resource-mapper.js, line 77
path is C:/programs/solid/data/rogerfgay.solid.example.com/
/${hostname} is /rogerfgay.solid.example.com

!path.startsWith(`/${hostname}/`)

SOLUTION?  (This is my first time in, but it looks something like starting comparison at the length of the root?)
!path.startsWith(`/${hostname}/`, length of root)

bourgeoa:
You cannot use a sub domain `solid.exemple.com` as a server name. In your case server name is `example.com`.

highlevellogic:
Be gentle. I'm a newbie and feeling a bit confused.
Setting up a Pod server instructions - https://solidproject.org/for-developers/pod-server
Section "Install Node Solid Server" - they're showing subdomains 2 levels deep.
Am I missing the point entirely?

bourgeoa:
Can you detail a bit what you are trying to achieve :
 . localhost
 . real server name

I can help on a server with a real server name. I never used localhost.

The easiest way is to prepare a `config.json` file.

You can begin with `config.json.default` that you can find in the folder where you `npm install` node-solid-server.
You can also find it https://github.com/solid/node-solid-server/blob/master/config.json-default
From there I shall try to help you.

[...]

PR #1499, opened 2020-10-24
Normalize rootPath for Windows (forward slashes)

highlevellogic:
Normalize rootPath for Windows (forward slashes)

This change should have no effect on Linux.

issue: Path must start with hostname #1491
one change in file lib/resource-mapper.js

"path" variable uses forward slashes. "rootPath" does not. IMO: all path variables should be normalized (forward slashes) ASAP to conform with URI / URL standards. Therefore, this change normalizes rootPath.

Before this change, rootPath (Windows) uses backslashes which causes comparison on line 72 (in method mapFileToUrl) to fail. Attempts to register an account throws error (line 77). Note that the "path" variable for individual resources has already been explicitely normalized. (line 71)

Unlike "path", rootPath is constant and this correction takes place only once in the constructor. This has nearly zero impact on performance. Tested on Windows. This should have no impact on other operating systems such as Linux.

TEST performed on Windows: Ran node-solid-server on Windows. Registered new account. Logged in.


lib/resource-mapper.js
```diff
@@ -69,7 +69,7 @@ class ResourceMapper {
   async mapFileToUrl ({ path, hostname }) {
     // Remove the root path if specified
     path = path.replace(/\\/g, '/')
-    if (path.startsWith(this._rootPath)) {
+	if (path.startsWith(this._rootPath)) {
```
michielbdejong, line 72: please undo this indent, it was correct with the `i` of if under the `p` of path

michielbdejong:
Failing in https://travis-ci.com/github/solid/node-solid-server/jobs/407056730#L287

lib/resource-mapper.js
highlevellogic: Done.

bourgeoa: You added a ";". Lint impose a very strict style. There should be no ;

highlevellogic: OK, I didn't realize. I've removed the ;

bourgeoa:
@michielbdejong 
Apparently all tests passes but there is `continuous integration expected`. what does it mean ?
I found this https://travis-ci.community/t/pr-build-status-stuck-at-continuous-integration-travis-ci-expected-waiting-for-status-to-be-reported/7220

highlevellogic:
Continuous integration generally means frequent merges. It looks like this is just waiting for review.

michielbdejong:
yes the build is green now

michielbdejong (commented):
I see your explanation of why this change is necessary makes sense but it fixes a bug that I haven't been able to reproduce yet.
And this change looks quite scary because it's in the resource mapper.

highlevellogic:
Officially, all testing is on Linux - which is presumably why you can't reproduce the problem. 
I'd like to suggest that the change isn't so scarey if it tests on Linux in the way described - i.e. has no effect.
No harm. No foul.
As for Windows - I think you can see what's happening just by looking at it. 
Two things are being compared: path and rootPath.
path was already being normalized with .replace(/\\/g, '/')
now rootPath is being normalize in exactly the same way.

The actual change isn't so scarey.

highlevellogic:
Or you could do a simulation of the problem on Linux by transforming both path and rootPath to backward slashes and then see what happens when only one is changed to forward slashes.

highlevellogic:
If you think it would help (to document or whatever) I can set up a separate simple demonstration for node.js ???

highlevellogic:
Reproducing the problem

To emulate the problem on Linux, change line 33 from 
    this._rootPath = this._removeTrailingSlash(rootPath)
to 
    this._rootPath = this._removeTrailingSlash(rootPath).replace(/\//g,'\\')

Test
then add the fix
 this._rootPath = this._removeTrailingSlash(rootPath).replace(/\//g,'\\').replace(/\\/g, '/')

Actual new code (just the fix):
this._rootPath = this._removeTrailingSlash(rootPath).replace(/\\/g, '/')

highlevellogic:
Would be nice to get this merged. Looks like more fixes may be coming in the same file.

highlevellogic:
With one more fix to this file, it looks like the data browser is now working well on Windows.
I will submit a new PR with both fixes included.

PR #1512, opened 2020-11-10
Add test/surface/

michielbdejong:


michielbdejong:
The nyc tests still seem to be picking up config from the Dockerfile

https://travis-ci.com/github/solid/node-solid-server/jobs/434304691#L260

michielbdejong:
The problem with this was mocha got confused by the new files under test/*. Fixed that by updating what mocha looks for.

Issue #1497, opened 2020-10-20
Emits websockets-pubsub on the parent containers


michielbdejong:
When a parent container changes, it's not picked up by https://github.com/solid/node-solid-ws/blob/master/index.js because that code only looks at the URL of the request and has no way of knowing whether any parent containers were created in the process. We probably need to call https://github.com/solid/node-solid-server/blob/0542fa24b240936eac3a8051f9ca01a8e93e1912/lib/create-server.js#L103 explicitly when adding/removing entries in (parent) containers

jaxoncreed:
I'm having this problem too. I might devote some time to fixing it.

RubenVerborgh:
I was talking to @timbl about this yesterday in the context of https://github.com/solid/community-server/pull/339

Given a resource R and a container C.

Out of that discussion came:
- upon `PATCH` to R: pub R
- upon `PUT` to R: pub R
- upon `DELETE` to R: pub C, pub R
- upon `POST` to C, pub C, pub R' (where R' is the newly created resource)

The Community Solid Server will implement the above.

jaxoncreed:
I will work on updating NSS to do the above.

PR #1523, opened 2020-11-20
Triggers websocket when a folder is recursively created

jaxoncreed:
Solves https://github.com/solid/node-solid-server/issues/1497

### DIFF
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -4,7 +4,7 @@ module.exports = handler
 
 const bodyParser = require('body-parser')
 const fs = require('fs')
-const fx = require('mkdir-recursive')
+const mkdirp = require('fs-extra').mkdirp
 const debug = require('../debug').handlers
 const error = require('../http-error')
 const $rdf = require('rdflib')
@@ -12,6 +12,8 @@ const crypto = require('crypto')
 const overQuota = require('../utils').overQuota
 const getContentType = require('../utils').getContentType
 const withLock = require('../lock')
+const { promisify } = require('util')
+const pathUtility = require('path')
 
 // Patch parsers by request body content type
 const PATCH_PARSERS = {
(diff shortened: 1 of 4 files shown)


lib/ldp.js
```diff
@@ -229,8 +229,10 @@ class LDP {
 
     // First check if we are above quota
     let isOverQuota
+    // For some reason someone thought it was a bright idea to make url actually a
+    // req sometimes but not all the time 🙄. So Now I have to account for that
```
michielbdejong, line 233: Can you specify who "I" is, or just rewrite this so it's more "us" and "we" or "the code needs to ..."

michielbdejong (approved):
Can you also unskip the test for this in the nss-skips branch of solid-crud-tests please?

lib/handlers/patch.js
```diff
@@ -60,7 +62,19 @@ async function patchHandler (req, res, next) {
     const pathSplit = path.split('/')
     const directory = pathSplit.slice(0, pathSplit.length - 1).join('/')
     if (!fs.existsSync(directory)) {
-      fx.mkdirSync(directory, { recursive: true })
+      const firstDirectoryRecursivelyCreated = await promisify(mkdirp)(directory)
```
timbl, line 65: The mkdirp function https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md doesn't seem to return the directory, only an error code. Are you below  in 69 assuming it firstDirectoryRecursivelyCreated will be the directory name?

jaxoncreed: I removed fs extra from node solid server as this seemed to be the only place it was used. I replaced it with https://www.npmjs.com/package/mkdirp which does return the directory.

lib/ldp.js
TallTed, line 233:
```suggestion
    // Someone had a reason to make url actually a req sometimes but not
    // all the time. So now we have to account for that, as done below.
```

PR #1526, opened 2020-11-26
PUT/POST with no content-type return 400

bourgeoa:


bourgeoa:
one test as been added for PUT
all tests passes locally

I could have modified `utils/getContentType()` that is returning `application/octet-stream` as default. But it is used elsewhere and I am not sure of the consequences, even if it also passes all tests locally.

jeff-zucker:
Great, thanks @bourgeoa.  Minor note about your error messages: you need an "s" on "requires"  because "request" is a singular noun.

POST request requires a valid content type via the content-type header

csarven:
Be careful with:
* respond with the `400` status code when there is no `Content-Type` header (or an invalid field-value) in the request, as per Solid specification and RFC 7231.
* respond with the `415` status code when the `Content-Type` header of the request includes an unsupported `media-type`, as per RFC 7231.

bourgeoa:
@csarven 
Thanks so I suppose I must update the issue title to 400 and make the code change.
It is a change with the actual NSS and actual tests that uses 415 for not existing content-type

NSS use `application/content-type` as a fallback for unsupported media-types so should not return 415 in such situation.

Should I consider this as a consensus ?

Thanks @jeff-zucker It was a copy/paste

jeff-zucker:
So is this the way it should be for POST, PUT, and PATCH? (I don't see PATCH in your PR, in the case where it is updating an existing resource, the existing content-type could be used, I suppose, but in the case where it is creating a resource, it would need a content-type as well)

* missing or empty content-type : fails with 400
* unsupported content-type : succeeds, setting content-type to application/octet-stream

I guess I would prefer unsupported to fail with 415 - otherwise, a simple typo could change the content-type and the user wouldn't even be aware unless they checked.  But if there's a reason for defaulting to octet-stream, I can live with it.

csarven:
If `media-type` is unsupported by the server, `415`. Otherwise, `2xx`.

Unless server can generate a representation of the resource that's equivalent to the representation that was used to create the resource, stick to using the original `media-type` when responding to a `GET` request. Do not change/set/fix the resource's representation to `application/octet-stream`, especially if the write operation wasn't rejected in the first place. `application/octet-stream` is not meaningfully or usefully equivalent to an unsupported `media-type`.

While a server may not be capable of doing anything with the resource besides "storing it as is", a client has the possibility to use the representation if the original `media-type` is echoed back. Setting `application/octet-stream` will mess that up.

If the request used `application/octet-stream` to create the resource and server accepted it, it is up to the server if it wants to echo the `media-type` back or determine the `media-type` and use that instead. Do note the last paragraph of this section: https://tools.ietf.org/html/rfc7231#section-3.1.1.5

michielbdejong:
Copying this branch to here so we get Travis to run.
Ah, but:
Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.
:(
Will run the tests on my local machine.

bourgeoa:
@TallTed Can you review this PR I don't know why the tests failed

TallTed:
@bourgeoa - I don't think my review will help in that regard. I'm not much in the coding department; more in the overall architecture and the human-facing language, UI, UX...

bourgeoa:
@michielbdejong 
The tests fail on the solid-test-suite, not the NSS tests.

For me the test-suite give a wrong result and the expected result should not even be `application/octet-stream` but 400
I had the problem due to be able to make a distinction between `no` contentType parameter and `non-valid` contentType in the mimeType.

That's why I had to change the PUT handler to directly look in the `req.get('content-type')` and not indirectly through a function returning `undefined`. I suppose it is what you get. so the PUT succeed and returns a not knowned contentType 'application/octet-stream'
```
FAIL test/surface/create-non-container.test.ts (11.34 s)
  ● Create non-container › in an existing container › using PUT › creates the resource

    expect(received).toContain(expected) // indexOf

    Expected substring: "text/plain"
    Received string:    "application/octet-stream"

      133 |         expect(responseCodeGroup(result.status)).toEqual("2xx");
      134 |         expect(await result.text()).toEqual("Hello World");
    > 135 |         expect(result.headers.get("Content-Type")).toContain("text/plain");
          |                                                    ^
      136 |       });
      137 |       it("adds the resource in the container listing", async () => {
      138 |         const containerListing = await getContainerMembers(

      at test/surface/create-non-container.test.ts:135:52
      at step (test/surface/create-non-container.test.ts:33:23)
      at Object.next (test/surface/create-non-container.test.ts:14:53)
      at fulfilled (test/surface/create-non-container.test.ts:5:58)

  ● Create non-container › in a non-existing container › using PUT › creates the resource

    expect(received).toContain(expected) // indexOf

    Expected substring: "text/plain"
    Received string:    "application/octet-stream"

      289 |         expect(responseCodeGroup(result.status)).toEqual("2xx");
      290 |         expect(await result.text()).toEqual("Hello World");
    > 291 |         expect(result.headers.get("Content-Type")).toContain("text/plain");
          |                                                    ^
      292 |       });
      293 |       it("adds the resource in the container listing", async () => {
      294 |         const containerListing = await getContainerMembers(

      at test/surface/create-non-container.test.ts:291:52
      at step (test/surface/create-non-container.test.ts:33:23)
      at Object.next (test/surface/create-non-container.test.ts:14:53)
      at fulfilled (test/surface/create-non-container.test.ts:5:58)
```

Issue #1524, opened 2020-11-24
NSS should not allow deletion of the content of the root


scenaristeur:
_should not allow deletion of the root folder and some other that I call system folders like (profile, inbox, settings, .well-known)_
https://github.com/solid/solidcommunity.net/issues/27

bourgeoa:
>When a DELETE request targets storage’s root container or its associated ACL resource, the server MUST respond with the 405 status code. Server MUST exclude the DELETE method in the HTTP response header Allow in response to safe method requests [RFC7231]. [Source]
https://solid.github.io/specification/#deleting-resources

bourgeoa:
soved with PR #1554

PR #1527, opened 2020-11-28
delete root should return 405

bourgeoa:
issue #1524 solid


bourgeoa:
Added 2 tests for root and rootAcl.
All test passes locally on my machine.

Issue #1529, opened 2020-11-29
POST/PUT in NSS allow multiple resource with same url and different content-type


bourgeoa:
You can create resources : 
1. https://pod/foo.txt with content-type `text/plain` => creates file : `/foo.txt`
2. https://pod/foo.txt with content-type `text/turtle` => creates file `/foo.txt$.ttl`
3. https://pod/foo.txt/ as a container => creates folder `/foo.txt`

The first 2 have the same url and there are 2 different files created.
All 3 have the same visual representation in SolidOs.

The first 2 should not occur and I think this is a bug
For the third one I remember @timbl writing it should not occur, but did not find where.
There was also a redirect discussion between foo/bar/ and foo/bar. His this something to do in GET ?

Is there any solid/specification, I did not find anything ? @csarven @michielbdejong

csarven:
No issues with either one of those. They will create/replace resource given representation depending on request semantics. If /foo.txt exists, /foo.txt/ should not, and vice-versa. Redirect is optional.

bourgeoa:
@csarven thanks
I shall work on correcting to only have one representation for 1. 2. 3. and let open redirect for more information

csarven:
Hmm, not exactly sure what you are after.. but there is no limit on how many or what representations can be made available for a resource per request. If you are considering whether to store one or more representations and have them available at request, that's entirely your (NSS's) call. You'll have to take care when updating of course.

jeff-zucker:
Is this correct?
```
/foo.txt exists                                                                 
  PUT /foo.txt   with any content type, overwrites                              
  PUT /foo.txt/  fails                                                          
  POST /foo.txt  creates a new file with a different location                   
  POST /foo.txt/ fails ??? or creates new folder with different location        
/foo.txt/ exists                                                                
  PUT /foo.txt   fails                                                          
  PUT /foo.txt/  fails                                                          
  POST /foo.txt  creates a new file with a different location                   
  POST /foo.txt/ creates new folder with different location                     
```

angelo-v:
> The first 2 should not occur and I think this is a bug

I don't think it is a bug as long as the correct representation is returned given the respective Accept header on GET

bourgeoa:
@angelo-v 
I do think it is a bug because it is not a question of data representation but of `2 different sets of data`.
For example it can be an **image and a text**, or **an RDF or any non RDF data**

If it where different representations of the data, do you know how to replace the data ?
[...]

PR #1537, opened 2020-12-08
no resource and container with same name

bourgeoa:
- NSS should not allow these 2 URI's to coexist /text.txt/ and /text.txt
POST return error 404
PUT return error 200
see https://github.com/solid/node-solid-server/issues/1529#issuecomment-735709503

- an URI cannot be related to 2 different paths.
PUT /text.txt with 'text/plain' shall replace path /text.txt$.ttl, actually PUT creates path /text.txt without deleting /text.txt$.ttl

- tests have been added and are all passing locally


TallTed:
I think you meant "`/txt.txt/` and `/txt.txt`", not "`/txt.txt/` and `/text.txt`"?

bourgeoa:
Thanks updated

PR #1549, opened 2021-01-14
Use solid-auth-fetcher

michielbdejong:
Continued from #1488


michielbdejong:
Error on main welcome page:
```
Uncaught ReferenceError: solidClientAuthentication is not defined at auth-buttons.js:82
```
That's in https://github.com/solid/node-solid-server/blob/use-solid-auth-fetcher/common/js/auth-buttons.js#L82

PR #1550, opened 2021-01-14
Update options.js

xpat:
This is related to a pull request discussed with @Vinnl and approved by @ewingson on January 14, 2020.  Re: [solid/solidproject.org] Update 2019-01-01-00_overview.md (#495)

In this new pull request, I only changed one thing, as it already looks to be partly in line with the approved changes in #495.

In this new pull request, I only suggest that we remove (for example, /etc/solid-server).

### DIFF
--- a/bin/lib/options.js
+++ b/bin/lib/options.js
@@ -50,7 +50,7 @@ module.exports = [
   },
   {
     name: 'config-path',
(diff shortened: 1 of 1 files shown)


michielbdejong (approved):
LGTM

michielbdejong:
Thanks for your contribution @xpat !

PR #1551, opened 2021-01-15
docker image: disable cors-proxy by default

angelo-v:
The proxy should only be turned on deliberately, because it can be abused by spammers


bourgeoa:
@angelo-v ready to merge. I just want understand.
Does your patch mean that with proxy = false, you let solid is take care of CORS with the trusted app.

angelo-v:
No, if this is not false, but any url path, NSS serves a proxy on that path that can proxy any other site. I think this was meant to be able to proxy requests that get CORS errors, but I never saw it working

bourgeoa (approved):

PR #1552, opened 2021-01-15
Expose X-Powered-By header

NoelDeMartin:
This header is not accessible from web applications, and I think that's unintended.


michielbdejong (approved):
Thanks a lot!

PR #1554, opened 2021-01-17
make NSS test-suite compliant

bourgeoa:
- clearAclCache(aclUrl) in ldp.put and ldp.delete
- implement accessTo Append and default Write for resource creation.
NSS used only default Write)
This is a regression issue. Because some there is more control. In practice may be none. There should not be a lot of container .acl without accessTo without acl mode Append and default Write.

There is one remaining point see issue for PATCH https://github.com/solid/specification/issues/220

### DIFF
--- a/lib/acl-checker.js
+++ b/lib/acl-checker.js
@@ -1,6 +1,7 @@
 'use strict'
 /* eslint-disable node/no-deprecated-api */
 
+const { dirname } = require('path')
 const rdf = require('rdflib')
 const debug = require('./debug').ACL
 const debugCache = require('./debug').cache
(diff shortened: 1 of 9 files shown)


lib/acl-checker.js
```diff
@@ -84,7 +85,19 @@ class ACLChecker {
       // FIXME: https://github.com/solid/acl-check/issues/23
       // console.error(e.message)
     }
-    const accessDenied = aclCheck.accessDenied(acl.graph, resource, directory, aclFile, agent, modes, agentOrigin, trustedOrigins, originTrustedModes)
+    let accessDenied = aclCheck.accessDenied(acl.graph, resource, directory, aclFile, agent, modes, agentOrigin, trustedOrigins, originTrustedModes)
+
+    // For create and update HTTP methods
+    if ((method === 'PUT' || method === 'PATCH' || method === 'COPY') && directory) {
+      // if resource and acl have same parent container,
+      // and resource do not exists then accessTo Append from parent is required
```
TallTed, line 93:
```suggestion
      // and resource does not exist then accessTo Append from parent is required
```

Issue #1465, opened 2020-08-17
Support creating a Container using PUT


Vinnl:
[According to the Solid spec](https://solid.github.io/specification/#write):

> Servers MUST create a container with URI path ending /{id}/ in container / for requests including the HTTP Link header with rel="type" targeting a valid LDP container type.

However, when I send a `PUT` request to a Container URL with a Link header `<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"`, NSS responds with:

> Can't write file: PUT not supported on containers, use POST instead

bourgeoa:
This is the new spec. Does it support also creation of missing containers with PUT ?

Vinnl:
If you mean "would the workaround of creating a dummy file inside the Container and then deleting it again" still work, then yes:

> Servers MUST create intermediate containers and include corresponding containment triples in container representations derived from the URI path component of PUT, POST and PATCH requests.

bourgeoa:
No, I was meaning : if the container URI contains missing container do you create the container and all missing ones or do you return an error.

csarven:
@Vinnl The requirement you've originally quoted in this issue is in context of `POST`. I can update the spec paragraph to clarify if that helps.

NSS doesn't support PUT to create a container (when target resource is a container). Background/proposal to support creating containers with PUT: https://github.com/solid/node-solid-server/issues/1108 and https://github.com/solid/node-solid-server/pull/513

Spec going forward clarifies slash semantics and hierarchical containment, PUT /path/to/container/ will create all intermediate containers. Client doesn't need include the Link header with a valid LDP container type for PUT. Server is not expected to observe it.

bourgeoa:
@csarven So if I read correctly your comment : Why the new spec does not state that PUT to a container URI (ending with /) is not allowed and should return error (405 or 400) ?

csarven:
@bourgeoa Can you please rephrase your question or comment?

It would be great if NSS can accept requests with PUT to create a container, as well as intermediate containers if they don't exist.

Note slash semantics requirement:

>The slash character in the URI path indicates hierarchical relationship segments, and enables relative referencing. The semantics of the slash character is shared by servers and clients. Paths ending with a slash denote a container resource.

NSS may also do the following:

>Servers MUST respond with the 405 status code to requests using HTTP methods that are not supported by the target resource.

bourgeoa:
@csarven
[...]

PR #1557, opened 2021-01-19
Make NSS test-suite and spec compliant

bourgeoa:
With this PR all tests from the test-suite are passing. Considering the following upgrades to be made :
- except the 2 ones relating to PATCH without Write. (see issue https://github.com/solid/specification/issues/220)
- Test suite needs to be upgraded with https://github.com/solid/web-access-control-tests/pull/19 https://github.com/solid/web-access-control-tests/pull/20
- rdflib needs to be upgraded to properly parse JSON-LD https://github.com/linkeddata/rdflib.js/pull/469
- mashlib needs to be upgraded solid/folder-pane https://github.com/solid/folder-pane/pull/17 (in the mean time there is a hack in (ldp.put() to add a content-type)

This is a bit large upgrade. I am sorry for it.

The main changes are : 
- creation of container with PUT (changes in ldp.js)
- resource creation needs acl:accessTo Append and acl:default Write (changes in acl-checker.js, handlers/allow.js, handlers/patch.js)
- delete do not need acl check on folder (changes in ldp-middleware.js)
- content-type is needed (except for create container with PUT. There is no default value.)
- pubsub notification for folder is now send in line with test-suite
 
Some more changes to comply with spec : 
- resource and container cannot only differ with / (changes in handlers/patch.js and ldp.js)

NSS implementation changes : 
- aclCache implement clearing cache with clearAclCache(url) in ldp.delete and ldp.put (changes in ldp.js and acl-checker.js)
- delete of root and root.acl are forbidden
- use the same functions in ldp.js and handlers/patch to create intermediate directories and resource and container doe not use same name differing with /

This PR + NSS 5.6.0 is running on https://solidcommunity.net:8443 test server with mashlib 1.6.5 latest published version working for long chat.


Vinnl:
It looks like this would also fix #1465, is that correct?

bourgeoa:
@Vinnl yes this solves #1465 however header link is not needed following @csarven comment in https://gitter.im/solid/test-suite
>  btw, that Link header is not necessary for PUT because the effective request URI already has a trailing slash

lib/acl-checker.js
```diff
@@ -40,7 +41,7 @@ class ACLChecker {
 
   // Returns a fulfilled promise when the user can access the resource
   // in the given mode, or rejects with an HTTP error otherwise
```
TallTed, line 43:
```suggestion
  // in the given mode; otherwise, rejects with an HTTP error
```

lib/acl-checker.js
```diff
@@ -84,7 +85,19 @@ class ACLChecker {
       // FIXME: https://github.com/solid/acl-check/issues/23
       // console.error(e.message)
     }
-    const accessDenied = aclCheck.accessDenied(acl.graph, resource, directory, aclFile, agent, modes, agentOrigin, trustedOrigins, originTrustedModes)
+    let accessDenied = aclCheck.accessDenied(acl.graph, resource, directory, aclFile, agent, modes, agentOrigin, trustedOrigins, originTrustedModes)
+
+    // For create and update HTTP methods
+    if ((method === 'PUT' || method === 'PATCH' || method === 'COPY') && directory) {
+      // if resource and acl have same parent container,
+      // and resource do not exists then accessTo Append from parent is required
```
TallTed, line 93:
```suggestion
      // and resource does not exist, then accessTo Append from parent is required
```

michielbdejong (approved):
Sounds great!

Issue #1548, opened 2021-01-14
non-turtle .meta files corrupt a folder


angelo-v:
*Actual behaviour*:

Given a folder contains a `.meta` file with invalid turtle
When the folder is requested
Then the server fails with status 500
And the browser shows `Can't parse container`

*Expected behaviour:*

NSS should reject to create `.meta` files that contain invalid turtle.

Further it should ignore existing `.meta` files or at least not try to parse them and fail.

It is essential that a folder containing such a `.meta` file for any reason can still be used and the `.meta` file can be fixed or deleted manually.

See also: https://github.com/solid/solidcommunity.net/issues/33

bourgeoa:
> NSS should reject to create `.meta` files that contain invalid turtle.

done 
> Further it should ignore existing `.meta` files or at least not try to parse them and fail.

Still to be discussed 
> It is essential that a folder containing such a `.meta` file for any reason can still be used and the `.meta` file can be fixed or deleted manually

Can be done using a direct link to .meta

bourgeoa:
https://github.com/solid/solidos/issues/129

bourgeoa:
#1677

PR #1563, opened 2021-02-02
.meta and remove mashlib hack

bourgeoa:
- Check that .meta has text/turtle content to not break containement triples. issue #1548 
- remove mashlib hack not needed anymore resolved in mashlib@1.6.8 PR https://github.com/solid/folder-pane/pull/17


lib/handlers/put.js
```diff
@@ -5,15 +5,16 @@ const debug = require('debug')('solid:put')
...
 
 async function handler (req, res, next) {
   debug(req.originalUrl)
   res.header('MS-Author-Via', 'SPARQL')
 
   const contentType = req.get('content-type')
-  if (LDP.mimeTypeIsRdf(contentType) && isAclFile(req)) {
-    return bodyParser.text({ type: () => true })(req, res, () => putAcl(req, res, next))
+  if (isAuxiliary(req)) {
+    if (contentType === 'text/turtle') {
```
michielbdejong, line 15: Why not use `LDP.mimeTypeIsRdf(contentType)` like above?

michielbdejong (approved):
LGTM

lib/handlers/put.js
bourgeoa: It would be certainly better, but I understand that in NSS .meta can only be text/turtle see ldp.list parsing meta content.
For acl it has been stated, I don't remember where. Maybe in solid/acl-check or acl-checker.

But I may misunderstand the code.

PR #1569, opened 2021-02-09
replace node-webid

bourgeoa:
Following a security issue on node-forge <0.10.0 consumed in solid/node-webid published as npm webid

imported node-webid linted code in lib/webid


michielbdejong (approved):
LGTM

PR #1577, opened 2021-03-03
[WiP] Add hasPaid(agent) function to ACL check

michielbdejong:


michielbdejong:
Linked to https://github.com/solid/acl-check/pull/38

michielbdejong:
Now seeing:
```sh
solid:ACL accessDenied: modeURIorReasons: ["User Unauthorized","Paying Would Help"] +0ms
```
in the logs so that's good progress so far.

michielbdejong:
I'm thinking the ACL checks are getting too complicated if we add this the way I originally thought we could.

So instead:
* run the ACL check as normal, but keep track of whether any of the authorisations that did not apply would have applied if either the user would (authenticate and) pay (i.e. was the `hasPaid(agent)` callback called at least once).
* if so and access was denied (401 or 403) then re-run the ACL check *as if* the agent had (authenticated and) paid. Careful not to cause any side-effects there.
* if so, respond with 402 instead of 401 / 403.
* if not (e.g. paying would give read access but the attempt is to control), then still give the 401/403 we would originally.

michielbdejong:
Got an approximation now where if hasPaid was checked and it was empty, and the response otherwise is a 401 or a 403, then it becomes a 402. This is too often, because it could be that you pay to get more access, but then that access will still not be enough to do what you were trying to do. But it's a reasonable first approximation.

I can now look at the integration of koa-ilp first, and the resolve this small inaccuracy last.

michielbdejong:
Hm, seems that https://github.com/interledgerjs/ilp-fetch supports both psk2 and stream (ILP/STREAM being the newer one) but https://github.com/interledgerjs/koa-ilp/blob/master/index.js only does psk2

michielbdejong:
https://interledger.org/rfcs/0025-pre-shared-key-2/ even explicitly says PSK2 is deprecated. So that's no good, then. Will see if I can find some code that creates a `Pay` header with stream.

michielbdejong:
Not sure how to add the `Pay` header into the response. https://github.com/solid/node-solid-server/blob/v5.6.2/lib/handlers/allow.js#L73 seems to be where the response status is passed on from the acl check, but not sure if we could add additional headers through that too

michielbdejong:
https://expressjs.com/en/guide/error-handling.html seems relevant.

michielbdejong:
Failing test from monetization-tests:
● Read-Paying › Gives a 402 with accessTo Read access on non-container resource

michielbdejong:
It's now making the call to the oracle and checking if the response body equals 'ok'.
The monetization-tests seem to be making some calls with a null agent, which I don't understand.
Should also maybe return the payment details from the oracle?

michielbdejong:
This PR goes together with https://github.com/solid/acl-check/pull/38 and https://github.com/solid/monetization-tests

PR #1604, opened 2021-06-14
add owner to acl authorization

bourgeoa:
owner is defined as the subject in /.meta.
This is pre-existing in NSS

### DIFF
--- a/lib/handlers/allow.js
+++ b/lib/handlers/allow.js
@@ -4,7 +4,6 @@ module.exports = allow
 const ACL = require('../acl-checker')
 const debug = require('../debug.js').ACL
 // const error = require('../http-error')
-const { getOwner } = require('../utils')
 
 function allow (mode) {
   return async function allowHandler (req, res, next) {
@@ -42,14 +41,17 @@ function allow (mode) {
     }
     // Obtain and store the ACL of the requested resource
     const resourceUrl = rootUrl + resourcePath
-    req.acl = ACL.createFromLDPAndRequest(resourceUrl, ldp, req)
-
     // Ensure the user has the required permission
     const userId = req.session.userId
-    const isAllowed = await req.acl.can(userId, mode, req.method, stat)
-    if (isAllowed) {
-      return next()
-    }
+    try {
+      req.acl = ACL.createFromLDPAndRequest(resourceUrl, ldp, req)
+
+      // if (resourceUrl.endsWith('.acl')) mode = 'Control'
+      const isAllowed = await req.acl.can(userId, mode, req.method, stat)
+      if (isAllowed) {
+        return next()
+      }
+    } catch (error) { next(error) }
     if (mode === 'Read' && (resourcePath === '' || resourcePath === '/')) {
       // This is a hack to make NSS check the ACL for representation that is served for root (if any)
       // See https://github.com/solid/node-solid-server/issues/1063 for more info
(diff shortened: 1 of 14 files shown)


lib/handlers/allow.js
```diff
@@ -68,6 +69,21 @@ function allow (mode) {
         }
       }
     }
+
+    // check user is owner. Find owner from /meta
+    let graph
+    try {
+      // do not check for permission. Owner has all rights.
```
csarven, line 76: Just  a heads-up that the owner has implicit Control privilege on all resources in the Storage, not necessarily Read-Write on the resource itself.

lib/handlers/allow.js
```diff
@@ -68,6 +69,21 @@ function allow (mode) {
...
+      // TODO check permissions on .meta ???
+      graph = await ldp.getGraph(rootUrl + '/.meta')
+      const SOLID = $rdf.Namespace('http://www.w3.org/ns/solid/terms#')
+      const owner = await graph.any(null, SOLID('account'), $rdf.sym(rootUrl + '/'))
```
csarven, line 80: Not sure about the role that `solid:account` plays towards the role that `solid:owner` plays. `solid:account` typically refers to the root container / Storage

csarven (commented):
Will the discovery of the solid:owner link relation from the root container be a separate PR?

lib/handlers/allow.js
timbl: `solid:account` seems to be what NSS is already using for this.  So this fix can piggyb ack on that.

lib/handlers/allow.js
timbl: Agree, only control, not necessarily read or write.

timbl: So something like
```
    if (mode === 'Control') {}
```
around this bit seems it would do that

timbl (changes requested):

bourgeoa:
> Agree, only control, not necessarily read or write.

This is not clear to me.
here is an ACL where users have only `Read` real example on solid-test running locahost as multiuser
https://alain.localhost:8443/public/text3.txt.acl
```
@prefix : <#>.
@prefix acl: <http://www.w3.org/ns/auth/acl#>.
@prefix foaf: <http://xmlns.com/foaf/0.1/>.
@prefix Gro: <ContactsTest/Group/Group.ttl#>.
@prefix tes: </testGroup.ttl#>.

:Read
    a acl:Authorization;
    acl:accessTo <text3.txt>;
    acl:agentClass foaf:Agent;
    acl:agentGroup Gro:this, tes:this;
    acl:mode acl:Read.```
```
If the `owner` wants to edit and change this ACL the owner in solidOs on NSS needs : 
- `Control` because it is an ACL
- `Read` to be able to GET the resource acl
- `Write` to PUT changes

 `Read` is only needed because of the display in solidOs, but `Write` is needed for PUT.

lib/handlers/allow.js
bourgeoa: was resolved with a check that resource must be an acl.
`mode = Control` for acl's is done in `acl-checker.js`

bourgeoa:
- when the webId is the owner Some test do return 200 ok if the acl is a bad acl (example with body = '')
- it used to be 403, and still is if the webId is not the owner).

Can this wait until implementation of WAC editor's draft ?
Test suite passes without any problem.

timbl (approved):
Looks good and I understand this is tested

michielbdejong (approved):
No objections from my side!

PR #1623, opened 2021-10-13
Remove the MS-Author-Via header

kjetilk:
In https://github.com/solid/specification/issues/124 it has been resolved to not bring the `MS-Author-Via` header into the spec, as it has been obsoleted by `Accept-Patch`, as it seems that NSS has implemented.

This PR removes it from the places that I could find it.

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -32,7 +32,7 @@ const corsSettings = cors({
   methods: [
     'OPTIONS', 'HEAD', 'GET', 'PATCH', 'POST', 'PUT', 'DELETE'
   ],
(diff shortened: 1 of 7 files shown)


jeff-zucker:
Not sure of all the implications, but I am not aware of problems it would cause.

jeff-zucker (approved):

timbl:
You don't just switch from one form to another.  You always have to have both for a while.   Severs in the server-interim must provide both, and clients during the client-interim accept either.    Interim periods of a year, for example, would be reasonable.  Then when it changes to remove the old one from servers, then you upgrade the major version of the protocol.  For example to 1.0.0 in the case of a protocol at 0.9

Issue #1629, opened 2021-10-29
pim:storage is missing


bourgeoa:
There is no pim:storage link header, nor a pim.storage triple in podRoot container representation
see https://solidproject.org/TR/protocol#storage

PR #1632, opened 2021-10-31
add space:Storage

bourgeoa:
add space:Storage to headers and  podRoot 'text/turtle' container representation
see issue #1629

### DIFF
--- a/lib/header.js
+++ b/lib/header.js
@@ -40,7 +40,7 @@ function addLinks (res, fileMetadata) {
     addLink(res, 'http://www.w3.org/ns/ldp#DirectContainer', 'type')
   }
   if (fileMetadata.isStorage) {
-    addLink(res, 'http://www.w3.org/ns/pim/space#Storage', 'type')
+    addLink(res, 'http://www.w3.org/ns/pim/space#storage', 'type')
   }
 }
 
@@ -106,7 +106,7 @@ function parseMetadataFromHeader (linkHeader) {
           fileMetadata.isBasicContainer = true
         } else if (parsedLinks[rel] === 'http://www.w3.org/ns/ldp#DirectContainer') {
           fileMetadata.isDirectContainer = true
-        } else if (parsedLinks[rel] === 'http://www.w3.org/ns/pim/space#Storage') {
+        } else if (parsedLinks[rel] === 'http://www.w3.org/ns/pim/space#storage') {
           fileMetadata.isStorage = true
         }
       }
--- a/lib/header.js
+++ b/lib/header.js
@@ -39,6 +39,9 @@ function addLinks (res, fileMetadata) {
   if (fileMetadata.isDirectContainer) {
     addLink(res, 'http://www.w3.org/ns/ldp#DirectContainer', 'type')
   }
+  if (fileMetadata.isStorage) {
+    addLink(res, 'http://www.w3.org/ns/pim/space#Storage', 'type')
+  }
 }
 
 async function linksHandler (req, res, next) {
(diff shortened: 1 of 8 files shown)


angelo-v (changes requested):
Please add unit tests, at least for the code you added.

lib/header.js
```diff
@@ -40,7 +40,7 @@ function addLinks (res, fileMetadata) {
     addLink(res, 'http://www.w3.org/ns/ldp#DirectContainer', 'type')
   }
   if (fileMetadata.isStorage) {
-    addLink(res, 'http://www.w3.org/ns/pim/space#Storage', 'type')
+    addLink(res, 'http://www.w3.org/ns/pim/space#storage', 'type')
```
angelo-v, line 43: Why lowercase? http://www.w3.org/ns/pim/space#storage is a property while http://www.w3.org/ns/pim/space#Storage is the class.

this is what CSS sends in the header:

```
link: <http://www.w3.org/ns/pim/space#Storage>; rel="type"
```

bourgeoa: Because NSS in profile/card uses space:storage, ESS also and the ontology also.
I think it is an error in CSS.

jeff-zucker: 1. The header on </>
```
   link:  <http://www.w3.org/ns/pim/space#Storage>; rel="type".
````
Is the same as the predicate in the turtle of </.meta>
```
   </> rdfs:type  <http://www.w3.org/ns/pim/space#Storage> .
```


2. OTOH, in the profile, the owner/server MAY have a  triple like.
```
   <#me> pim:storage </>.
```
#1 asserts that we are at the Pod root.  #2 asserts that WebId owns the pod.  Some owners may not want to have that fact advertised and there is no requirement that they do.

Comment (kjetilk) [reply lib/header.js:43]: There is a convention to have predicates begin with a lower-case letter, and classes begin with an uppercase letter, and that's what we're seeing here: `space:Storage` is a class, `space:storage` is a predicate, and both are in use. And indeed, as @jeff-zucker is saying, the `Link` header are using classes, so it should be upper case, like @angelo-v objected.

File: lib/ldp-container.js [hunk 2]
```diff
@@ -14,6 +14,14 @@ const path = require('path')
 async function addContainerStats (ldp, reqUri, filename, resourceGraph) {
   const containerStats = await ldp.stat(filename)
   addStats(resourceGraph, reqUri, containerStats, filename)
+  const storage = new URL(reqUri)
+  if (reqUri === storage.origin + '/') {
+    resourceGraph.add(
+      resourceGraph.sym(reqUri),
+      ns.rdf('type'),
+      ns.space('storage')
```
csarven, line 22:
```suggestion
      ns.space('Storage')
```

lib/header.js
```diff
@@ -102,6 +106,8 @@ function parseMetadataFromHeader (linkHeader) {
           fileMetadata.isBasicContainer = true
         } else if (parsedLinks[rel] === 'http://www.w3.org/ns/ldp#DirectContainer') {
           fileMetadata.isDirectContainer = true
+        } else if (parsedLinks[rel] === 'http://www.w3.org/ns/pim/space#storage') {
```
csarven, line 109:
```suggestion
        } else if (parsedLinks[rel] === 'http://www.w3.org/ns/pim/space#Storage') {
```

lib/header.js
csarven: Confirming that the Solid Protocol requires `http://www.w3.org/ns/pim/space#Storage` (uppercase 'S') for Link rel=type.

```suggestion
    addLink(res, 'http://www.w3.org/ns/pim/space#Storage', 'type')
```

csarven (changes requested):

kjetilk (approved):
This looks good to me now. However, it has been so long since I was into the details of NSS that it should probably receive more than my review :-)

angelo-v (approved):

PR #1633, opened 2021-11-01
Check write on container for delete

michielbdejong:
See https://github.com/solid/web-access-control-tests/issues/42
This reverts https://github.com/solid/node-solid-server/pull/1557/files#diff-296d21adeb16979ad974b59ab2cf7fa5be21fb4c898c3d5aab20fbd5081c97f3L29-R29


bourgeoa:
Also `checkPermissionsForDirectory` was removed from  lib/handlers/allow.js
Not sure that this is the solution is.

But may be around https://github.com/solid/node-solid-server/blob/ac596768d3360aeb9f4137e4025b0b478052429b/lib/acl-checker.js#L91

acl-checker was upgraded to be able to check permissions separately on resource and parent.

michielbdejong:
@bourgeoa please review

lib/acl-checker.js
```diff
@@ -98,6 +98,18 @@ class ACLChecker {
...
+
+    // For delete HTTP method
+    if ((method === 'DELETE') && directory) {
+      // if resource and acl have same parent container,
+      // then accessTo Write from parent is required
+      if (directory.value === dirname(aclFile.value) + '/') {
+        const accessDeniedAccessTo = aclCheck.accessDenied(acl.graph, directory, null, aclFile, agent, [ACL('Write')], agentOrigin, trustedOrigins, originTrustedModes)
+        const accessResult = !accessDenied && !accessDeniedAccessTo
+        accessDenied = accessResult ? false : accessDenied || accessDeniedAccessTo
+        // debugCache('accessDenied result ' + accessDenied)
```
bourgeoa, line 110: Why not make a `function accessToFromParent (aclValue)`

Issue #1387, opened 2020-01-09
Link header for pod root index.html is null


jeff-zucker:
This script prints "null"
```javascript
const auth = require('solid-auth-cli')
const root = "https://jeffz.solid.community/index.html"

auth.login().then(
    auth.fetch( root ).then( response =>
        console.log( response.headers.get("link") )
    )
)

```

jaxoncreed:
That's strange. It should return a header. I don't remember if headers.get cares about case sensitivity, but it does return a "Link" header rather than a "link" header.

csarven:
As I understand it in NSS, /index.html is a resource contained in /.

/index.html is also purposed as a representation of /.

/index.html doesn't really need an LDP interaction model if that's what you're looking for, but I think it is generally expected to behave as a generic ldp:Resource here.

jeff-zucker:
@jaxoncreed Nope, the capitalization of link/Link does not matter.  If I ask for /foo/, I get link headers either way and if I ask for /foo/index.html I get no headers either way.  
```javascript
const auth = require('solid-auth-cli')
const file = "https://jeffz.solid.community/index.html"
const path = "https://jeffz.solid.community/"

auth.login().then(
    auth.fetch( file ).then( response => {
        console.log( response.headers.get("link") )
        console.log( response.headers.get("Link") )
    }).then(
        auth.fetch( path ).then( response => {
            console.log( response.headers.get("link") )
            console.log( response.headers.get("Link") )
        })
    )
)
```
Outputs :
```
null
null
<.acl>; rel="acl", <.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
<.acl>; rel="acl", <.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
```

jeff-zucker:
@csarven I see the theoretical benefit of treating index.html as a representation, but it's going to be quite confusing if it isn't also treated as a resource.  I can currently GET index.html and expect it to have all its headers.  I'm looking at this from the context of a client app wanting to copy all contents of a pod.  Would a client app have to query / for a variety of representations to know what to copy?

melvincarvalho:
@jeff-zucker good catch!

I also deploy apps to index.html, and would like to be able to read and write to that resource, so I have upvoted this issue

If time allows I may be able to help

Do we have a desired behaviour?

PR #1643, opened 2021-12-13
use of container representation should not change container headers

bourgeoa:
resolves Issue #1387

### DIFF
--- a/test/integration/http-test.js
+++ b/test/integration/http-test.js
@@ -167,8 +167,7 @@ describe('HTTP APIs', function () {
         .end(done)
     })
 
-    // This test is probably wrong: it is not a container if there is an index page
-    it.skip('should have set Link as resource on a implicit index page', function (done) {
+    it('should have set Link as Container/BasicContainer on a implicit index page', function (done) {
       server.options('/sampleContainer/')
         .expect('Link', /<http:\/\/www.w3.org\/ns\/ldp#BasicContainer>; rel="type"/)
         .expect('Link', /<http:\/\/www.w3.org\/ns\/ldp#Container>; rel="type"/)
--- a/test/integration/http-test.js
+++ b/test/integration/http-test.js
@@ -167,7 +167,7 @@ describe('HTTP APIs', function () {
         .end(done)
     })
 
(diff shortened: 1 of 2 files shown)


test/integration/http-test.js
```diff
@@ -167,8 +167,7 @@ describe('HTTP APIs', function () {
         .end(done)
     })
 
-    // This test is probably wrong: it is not a container if there is an index page
-    it.skip('should have set Link as resource on a implicit index page', function (done) {
+    it('should have set Link as Container/BasicContainer on a implicit index page', function (done) {
```
TallTed, line 170:
```suggestion
    it('should have set Link as Container/BasicContainer on an implicit index page', function (done) {
```

jeff-zucker (commented):
This makes sense for URLs in the form https://user.example.com/ but not for https://example.com/.  The second one is not a pim:Storage, is it?

jeff-zucker:
Also, wouldn't this violate the semantics of pim:Storage?  AFAIK you can never have one Storage inside another Storage.  If the server root is a Storage than all the user Storages are inside it.

bourgeoa:
> This makes sense for URLs in the form https://user.example.com/ but not for https://example.com/. The second one is not a pim:Storage, is it?

Agreed I opened an issue. I did not had a solution to find the provider Uri.

> Also, wouldn't this violate the semantics of pim:Storage? AFAIK you can never have one Storage inside another Storage. If the server root is a Storage than all the user Storages are inside it.

No because https://user.example.com/ is not a container of https://example.com

bourgeoa:
@jeff-zucker  Anyway this PR is not about pim:storage, but only container representation. See files changed

jeff-zucker:
> @jeff-zucker Anyway this PR is not about pim:storage, but only container representation. See files changed

Then I am afraid I do not understand the PR at all. What do you mean by container representation?  Doesn't this line mark the root folder as pim:Storage?
```
if (req.path === '/' && req.app.locals.host.serverUri.contains(req.hostname)) fileMetadata.isStorage = true
```

bourgeoa:
The resulting PR is only one line and a test

image

bourgeoa:
merged with PR #1646

PR #1644, opened 2021-12-20
Created a docker-compose file for development purpose

paulorodriguesxv:
It was created a docker-compose file for development purpose. It's facilitate server execution with no necessity of using solid-test script

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ We have automatic builds set up, so commits to master will trigger a build of ht
 
 ### Development usage
 
-If you want to use Docker in development, you can build and run the image locally with either docker-compose:
+If you want to use Docker in development, you can build and run the image locally with either docker-compose —
 
 ```bash
 git clone https://github.com/solid/node-solid-server
@@ -245,7 +245,7 @@ cd node-solid-server
 docker-compose up -d
 ```
 
-or these manual commands:
+ — or these manual commands —
 
 ```bash
 git clone https://github.com/solid/node-solid-server
--- a/README.md
+++ b/README.md
@@ -239,6 +239,15 @@ We have automatic builds set up, so commits to master will trigger a build of ht
 
 If you want to use Docker in development, then you can build it locally with:
 
+Using docker-compose
+```bash
+git clone https://github.com/solid/node-solid-server
+cd node-solid-server
+docker-compose up -d
+```
+
+or build image manually and run it:
+
 ```bash
 git clone https://github.com/solid/node-solid-server
 cd node-solid-server
--- a/README.md
+++ b/README.md
@@ -237,27 +237,24 @@ We have automatic builds set up, so commits to master will trigger a build of ht
 
 ### Development usage
 
-If you want to use Docker in development, then you can build it locally with:
+If you want to use Docker in development, you can build and run the image locally with either docker-compose:
 
-Using docker-compose
 ```bash
 git clone https://github.com/solid/node-solid-server
 cd node-solid-server
 docker-compose up -d
 ```
 
-or build image manually and run it:
+or these manual commands:
 
 ```bash
 git clone https://github.com/solid/node-solid-server
 cd node-solid-server
 docker build -t node-solid-server .
-```
 
-Run with:
-```bash
 docker run -p 8443:8443 --name solid node-solid-server
 ```
+
 
 This will enable you to login to solid on https://localhost:8443 and then create a new account
 but not yet use that account. After a new account is made you will need to create an entry for 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,10 @@
+version: "3"
+services:
+  solid-server:
+    build: .
+    container_name: solid
+    ports:
+      - "8443:8443"
+    entrypoint: npm run solid start -- --no-reject-unauthorized
+    environment:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -239,6 +239,15 @@ We have automatic builds set up, so commits to master will trigger a build of ht
 
 If you want to use Docker in development, then you can build it locally with:
 
+Using docker-compose
```
TallTed, line 242:
```suggestion
If you want to use Docker in development, you can build the image
locally with either docker-compose —
```

README.md
```diff
@@ -239,6 +239,15 @@ We have automatic builds set up, so commits to master will trigger a build of ht
...
+```bash
+git clone https://github.com/solid/node-solid-server
+cd node-solid-server
+docker-compose up -d
+```
+
+or build image manually and run it:
```
TallTed, line 249:
```suggestion
— or these manual commands —
```

TallTed (changes requested):
One additional tweak should be made to line 257 of `README.md`,
changing the existing —
```pre
Run with:
```
— to —
```pre
Once the image is built by either method, run it with —
```

README.md
paulorodriguesxv: Hi @TallTed.

Text adjusted. Thanks!

README.md
paulorodriguesxv: Hi @TallTed.

Text adjusted. Thanks!

README.md
```diff
@@ -237,28 +237,25 @@ We have automatic builds set up, so commits to master will trigger a build of ht
 
 ### Development usage
 
-If you want to use Docker in development, then you can build it locally with:
+If you want to use Docker in development, you can build and run the image locally with either docker-compose:
```
TallTed, line 240: Colon is not correct here. M-dash is.
```suggestion
If you want to use Docker in development, you can build and run the image locally with either docker-compose —
```

README.md
```diff
@@ -237,28 +237,25 @@ We have automatic builds set up, so commits to master will trigger a build of ht
...
 
-Using docker-compose
 ```bash
 git clone https://github.com/solid/node-solid-server
 cd node-solid-server
 docker-compose up -d
 ```
 
-or build image manually and run it:
+or these manual commands:
```
TallTed, line 248: M-dash leading into continued text is correct.  Closing colon is also not correct; M-dash is again.
```suggestion
— or these manual commands —
```

TallTed (changes requested):
Re-suggesting a couple of changes.

It appears to me that the `docker-compose` instruction does not launch the image.  If that is correct, then either the image launch following the manual build should be broken back out (so it applies to both build methods), or an image launch command should be added to the `docker-compose` command sequence (so both build sequences end with image launch).

paulorodriguesxv:
> Re-suggesting a couple of changes.
> 
> It appears to me that the `docker-compose` instruction does not launch the image. If that is correct, then either the image launch following the manual build should be broken back out (so it applies to both build methods), or an image launch command should be added to the `docker-compose` command sequence (so both build sequences end with image launch).

Hi @TallTed 

docker-compose comand here, in the way it was created, build the image and launch them. Once I'm doing docker-compose "up -d", it's launch the image. You could observe at the sscrip that there is a<build .>  key, it's mean that docker-compose will build the image in the case it wasn't built yet

README.md
paulorodriguesxv: Ok. I replaced colon by M-dash. Please, check it's ok now. Tks.

TallTed (approved):

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: "3"
+services:
+  solid-server:
+    build: .
+    container_name: solid
+    ports:
+      - "8443:8443"
+    entrypoint: npm run solid start -- --no-reject-unauthorized
```
angelo-v, line 8: **question** Why is `--no-reject-unauthorized` added? And why is the same(?) thing added as an env variable below as well?

paulorodriguesxv: Hi @angelo-v 

Because it's exactly same thing that bin/solid-test is doing, as you can see:

# Disable rejectUnauthorized when starting the server
if [ "$COMMAND" == "start" ]; then
  ADD_FLAGS="--no-reject-unauthorized"
  export NODE_TLS_REJECT_UNAUTHORIZED=0
fi

Respectfully speaking and excuse my sincerity, but this is simple docker-compose to help developer (like me) to running up application easily. But it is stuck in here for a long 1 month and ithas been generating so much questions that seems unbelieve to me.. Please feel free to decline.

angelo-v: Thanks for the hint @paulorodriguesxv I now understand and I think we can just merge it.

I totally understand your frustration. Unfortunately node-solid-server does not get much attention from the community anymore. Only a tiny group of volunteers keeping an eye on it more or less. I have the permission to merge and will do it, but I am not a NSS maintainer in any regard

angelo-v (approved):

angelo-v:
The build failed at "Run bash test/surface/run-solid-test-suite.sh $BRANCH_NAME" because this does not work well with forks. Since the rest of the pipeline ran successfully and the solid-test-suite is not relevant here, I wil merge any

PR #1646, opened 2021-12-23
update n3-patch to solid/protocol v0.9.0

bourgeoa:
- update n3-patch to solid/protocol v0.9.0 `solid:InsetDeletePatch` and move all tests
- keep deprecated `solid:patches` for compatibility issue with no tests


jeff-zucker:
Nice work, thanks!

PR #1668, opened 2022-02-20
Docker image: update Node.js from v10 to v16 (current LTS version)

sstieger:
As the official Docker image currently uses Node.js v10 (which has already reached its EOL), I would suggest updating to the current LTS version v16, which should also provide some performance benefits.


angelo-v (approved):

sstieger:
@angelo-v I just pushed a small update changing the test method's name, sorry for not noticing earlier :slightly_smiling_face:

PR #1676, opened 2022-03-05
Git for Windows suggestion.

chunt007:
Updated tutorial to inform Windows users of GitBash to use -verbose command. The installation process feels like its hung up because it doesn't show installation progress. Comparatively, Linux has an expected output of progress. Using this command allows us to identify any trivial matters for those using GitBash MINGW64

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,14 @@ You can install and run the server either using Node.js directly or using
 first approach, for the second approach see the section [use Docker](#use-docker)
 Section below.
 
+**Note**: If using GitBash, it is helpful to use the -verbose flag to see the progress of the install.
+
 To install, first install [Node](https://nodejs.org/en/) and then run the following
 
 ```bash
 $ npm install -g solid-server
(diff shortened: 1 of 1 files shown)


bourgeoa:
Thanks. Did you succeed to run NSS on windows.
I don't anyone really did.
I personally uses WSL  or WSL2, https://vscode.dev

chunt007:
> Thanks. Did you succeed to run NSS on windows. I don't anyone really did. I personally uses WSL or WSL2, https://vscode.dev

It seems to work fine. -verbose may produce different errors that when I have encountered, but I was able to navigate through the installation process with little tricks.

I'll throw them into a technical documentation somewhere!

bourgeoa:
Could I suggest that you do not wait and begin an has is in the wiki or in an issue as follow up.
There has been some GitHub issues raised end 2020 related to Windows. You can try a search in github issues with `windows`.

PR #1677, opened 2022-03-06
do not create container .meta by default

bourgeoa:


theRealImy (approved):
An intricate problem which was following my pod. The thought process of the fix seems logical to me.

PR #1678, opened 2022-03-06
reject url with %2F

bourgeoa:


jeff-zucker:
Hmm, is this really what we want to do?  What about a URI with an URI-encoded search string?  Should it be forbidden in the URI itsdelf but permitted in the search string?

lib/resource-mapper.js
```diff
@@ -153,11 +153,13 @@ class ResourceMapper {
   _parseUrl (url) {
     // URL specified as string
     if (typeof url === 'string') {
+      if (url.includes('%2F')) throw new Error('Url cannot contain %2F (%encoded /)')
       return URL.parse(url)
     }
     // URL specified as Express request object
     if (!url.pathname && url.path) {
       const { hostname, path } = url
+      if (path.includes('%2F')) throw new Error('Url cannot contain %2F (%encoded /)')
```
RubenVerborgh, line 162: The above implementation is broken; it only works on 2 out of 3 cases.
Rather than duplicating the code another time, it should be in a single place.

So if you must, do:

```
  // Parses a URL into hostname and pathname
  _parseUrl (url) {
    let parsed;
    // URL specified as string
    if (typeof url === 'string') {
      parsed = this._parseUURL.parse(url)
    }
    // URL specified as Express request object
    else if (!url.pathname && url.path) {
      const { hostname, path } = url
      parsed = { hostname, pathname: path.replace(/[?#].*/, '') }
    }
    // URL specified as object
    else {
        parsed = url;
    }
    // <=== do your path replacements here
    return parsed
  }
```

bourgeoa: Thanks

bourgeoa:
resolved with #1690 merged with #1695

PR #1682, opened 2022-03-22
Revert "Remove the MS-Author-Via header"

bourgeoa:
Reverts solid/node-solid-server#1623

### DIFF
--- a/lib/create-app.js
+++ b/lib/create-app.js
@@ -32,7 +32,7 @@ const corsSettings = cors({
   methods: [
     'OPTIONS', 'HEAD', 'GET', 'PATCH', 'POST', 'PUT', 'DELETE'
   ],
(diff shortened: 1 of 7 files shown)


TallTed:
Is there an issue associated with, or at least meeting minutes including some discussion of, this reversion?  I don't have a horse in the race; I just want to understand what's happening.

I'm also confused by the `obsolete` label on #1623, as in my experience that label is usually used to say the issue or PR on which it's placed is the obsolete thing (and so should be closed without being merged), not that the PR bearing that label is addressing some obsolete element in the target repo...

csarven:
Why not update mashlib/rdflib to use Accept-Patch if they aren't already?

bourgeoa:
The reason is that the latest NSS@5.7.0 is broken cannot edit card in dashboard.
Some people try to use it.

Issue has been created see #1683 and we are looking at mashilb/rdflib to find the issue.

bourgeoa:
The obsolete qualification may be misleading. The use of ms-author-via is obsolete.

theRealImy:
I can confirm this PR fixes SolidOS bug: https://github.com/solid/solidos/issues/138

theRealImy (approved):

PR #1695, opened 2022-08-03
PATCH jsonld, rdf+xml and n3

bourgeoa:


lib/resource-mapper.js
```diff
@@ -149,6 +147,31 @@ class ResourceMapper {
     return { path, contentType: contentType || this._defaultContentType }
   }
 
+  // encode/decode path except slash (/) / %encodedSlash (%2F|%2f) /ntimes%encodedSlash (%2525...2F|%2525...2f)
```
TallTed, line 150:
```suggestion
  // encode/decode path except slash (/), %encodedSlash (%2F|%2f), or ntimes%encodedSlash (%2525...2F|%2525...2f)
```

lib/handlers/patch.js
```diff
@@ -18,8 +18,17 @@ const PATCH_PARSERS = {
   'text/n3': require('./patch/n3-patch-parser.js')
 }
 
+// use mime-type as contentType for new rdf resource
```
TallTed, line 21:
```suggestion
// use media-type as contentType for new RDF resource
```
MIME types became media types some years ago. We should use current terminology whenever possible.

lib/handlers/patch.js
```diff
@@ -33,9 +42,9 @@ async function patchHandler (req, res, next) {
       // First check if the file already exists
       ({ path, contentType } = await ldp.resourceMapper.mapUrlToFile({ url: req }))
     } catch (err) {
-      // If the file doesn't exist, request one to be created with the default content type
+      // If the file doesn't exist, request one to be created with the file mime type as contentType
```
TallTed, line 45:
```suggestion
      // If the file doesn't exist, request to create one with the file media type as contentType
```

test/integration/patch-test.js
```diff
@@ -84,6 +84,17 @@ describe('PATCH', () => {
       result: '@prefix : </new.ttl#>.\n@prefix tim: </>.\n\ntim:x tim:y tim:z.\n\n'
     }))
 
+    describe('on a non-existing jsonld file', describePatch({
```
TallTed, line 87:
```suggestion
    describe('on a non-existent JSON-LD file', describePatch({
```

bourgeoa:
Should XML+RDF be spec compliant. rdfA is not ???
This questioned the default in case of new file in mashlib.
Creating a .rdf resource succeed but display a parse error.

jeff-zucker:
> Should XML+RDF be spec compliant. rdfA is not ??? This questioned the default in case of new file in mashlib. Creating a .rdf resource succeed but display a parse error.

Are you talking specifically about the content type of the file being PATCHed?  I sincerely hope that in the future RDFa will be patchable but that's a big task and not in the spec (yet) so no, I don't think we can offer to support it now.  The simple case of creating a new RDFa file with PATCH could easily work but it does not make sense to me to support it if none of the other operations can work.  As for XML+RDF - I am not familiar enough with parsing it to know for sure, but it seems to me that there would be no special issues for PATCH and therefore should be supported.

jeff-zucker:
^^ By which I mean that we should support sending a text/n3 PATCh to change an xml+rdf document, not that we should support sending an xml+rdf PATCH.

timea-solid:
I feel out of my depth to review this.

michielbdejong (approved):
LGTM! We should actually have a test for this in the test suite probably - created https://github.com/solid-contrib/test-suite/issues/148 about that

PR #1715, opened 2023-01-20
GET/PATCH HTML data island

bourgeoa:
This implements the discovery and creation/modification of HTML data island : a script HTML block containing RDF data.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -128,6 +128,9 @@ async function handler (req, res, next) {
     res.send(data)
     return next()
   } catch (err) {
+    if (err.message === '404') {
+      return next(error(404, 'HTML do not contain data island'))
+    }
     debug('error translating: ' + req.originalUrl + ' ' + contentType + ' -> ' + possibleRDFType + ' -- ' + 500 + ' ' + err.message)
     return next(error(500, 'Error translating between RDF formats'))
   }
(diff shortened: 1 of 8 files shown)


lib/handlers/get.js
```diff
@@ -128,6 +128,9 @@ async function handler (req, res, next) {
     res.send(data)
     return next()
   } catch (err) {
+    if (err.message === '404') {
+      return next(error(404, 'HTML do not contain data island'))
```
TallTed, line 132:
```suggestion
      return next(error(404, 'HTML does not contain any data islands'))
```

bourgeoa: The plural in `islands` seems to allow multiples. Actually my PR only allows one.
Not sure that more than one could be expected. The `id="data"` cannot be used more than once in an html document.

TallTed: There may be multiple islands in a single HTML doc, tho perhaps not (yet) in your environment. My error text is equivalent to saying, "I found no data island in this HTML," among other phrasings.

Another phrasing suggestion might be, `HTML does not contain a data island`.

melvincarvalho:
Thanks very much for looking at this.  I am very interested in this work.  Was it ever tried?  If not, I could perhaps try running the branch, to see what it looks like.

bourgeoa:
It has been implemented in SolidOS source-pane where you can test it.
Not in NSS. I don't remember why.

PR #1726, opened 2023-03-29
checkItmName recursively & reject resource with $.ext

bourgeoa:


lib/ldp.js
```diff
@@ -350,17 +357,27 @@ class LDP {
...
       const testUrl = container ? itemUrl.slice(0, -1) : itemUrl + '/'
       const { path: testPath } = await this.resourceMapper.mapUrlToFile({ url: testUrl })
-      // testName = fs.lstatSync(testPath).isDirectory()
       testName = container ? fs.lstatSync(testPath).isFile() : fs.lstatSync(testPath).isDirectory()
-    } catch (err) { testName = false }
-
+    } catch (err) {
+      testName = false
+
+      // item do not exist, check one level up the tree
```
TallTed, line 370:
```suggestion
      // item does not exist, check one level up the tree
```

bourgeoa:
@michielbdejong 
The new code was missing the fact that it needed to check the % encoded $
Tested locally and it is now working as expected.
I'm not sure of the error code used 400 and not 500 (server decision) ? This is not a solid specification.

Samely I am not sure with error 404 error code rejecting same name for documentResource and container, but here it is a solid protocol spec

lib/ldp.js
```diff
@@ -243,7 +243,14 @@ class LDP {
         'PUT request requires a content-type via the Content-Type header')
     }
 
+    // reject resource with % encoded $ extension
```
TallTed, line 246:
```suggestion
    // reject resource with percent-encoded $ extension
```

test/integration/http-test.js
```diff
@@ -535,13 +535,29 @@ describe('HTTP APIs', function () {
         .expect(hasHeader('acl', 'baz.ttl' + suffixAcl))
         .expect(201, done)
     })
-    it('should not create new resource if folder with same name exists', function (done) {
-      server.put('/foo/bar')
+    it('should not create a resource with % encoded $.ext', function (done) {
```
TallTed, line 538:
```suggestion
    it('should not create a resource with percent-encoded $.ext', function (done) {
```

TallTed (changes requested):
Better to use `percent-encoded` wording than `% encoded` symbol shorthand.

I'm not so sure about `$` vs `dollar-sign-extension` or `dollar-extension`, so I'm leaving those unchanged.

I'm open to embellishing by including the `$` and/or `%` symbol(s) *with* the words.

timea-solid (approved):
Why should I not be able to create a resource with same name as container?

PR #1728, opened 2023-04-15
isOwner there may be multiple owners

bourgeoa:


melvincarvalho:
multiple owners is a great idea, I've needed that for a while

PR #1738, opened 2023-09-04
Handle error when Origin header value is not IRI

csarven:
Server breaks when `Origin` header value is not an IRI. See:

https://github.com/nodeSolidServer/node-solid-server/blob/958a6ef82e67c5bfc000488aca4c8bfdd6a9bfa0/lib/acl-checker.js#L31

calling `rdf.sym` where an Error is thrown but it is not caught.

Following examples typically return 200:

```sh
$ curl -i https://example.org/
$ curl -iH 'Origin: https://example.net' https://example.org/
```

Following examples breaks the script (`bin/solid.js`). 500 returned.

```sh
$ curl -iH 'Origin: null' https://example.org/
$ curl -iH 'Origin: breaking bad' https://example.org/
```

There may be a better way to fix this issue than what's in this PR but I'll leave it in the hands of folks that can do better.


ewingson (approved):
happy to review.
Sarven has tested it as in https://app.element.io/#/room/!YCVoUIVohPTofEksKL:matrix.org/$hkj5YAplQrKLQaAxt6ZKCR4vbvUU94Y4lpwH0hm4ugg

LGTM

PR #1745, opened 2023-11-30
Append patch new document

bourgeoa:
This allow to PATCH a new document with Append only.


test/integration/acl-oidc-test.js
```diff
@@ -551,7 +551,7 @@ describe('ACL with WebID+OIDC over HTTP', function () {
         done()
       })
     })
-    it.skip('user1 should be able to PATCH a resource', function (done) {
+    it('user1 should be able to PATCH a non existing resource', function (done) {
```
TallTed, line 554:
```suggestion
    it('user1 should be able to PATCH (which CREATEs) a nonexistent resource', function (done) {
```

test/integration/acl-oidc-test.js
```diff
@@ -589,6 +589,16 @@ describe('ACL with WebID+OIDC over HTTP', function () {
         done()
       })
     })
+    it('user2 should be able to PATCH insert on a non existing resource', function (done) {
```
TallTed, line 592:
```suggestion
    it('user2 should be able to PATCH INSERT to (which CREATEs) a nonexistent resource', function (done) {
```

PR #1746, opened 2023-12-19
Append put new document

bourgeoa:


bourgeoa:
merged with https://github.com/nodeSolidServer/node-solid-server/pull/1750

PR #1750, opened 2024-01-10
5.7.9 beta

bourgeoa:
This PR seems rather big.
I should have made smaller bits.

1. It is the result of discoveries made when trying to pass the conformance tests.
These are passing entirely now.

The issues where raised in : 
- https://github.com/solid-contrib/specification-tests/issues/112
  `acl-check for DELETE` was wrong and needed access to resource ACL but also to parent resource ACL 
- https://github.com/solid-contrib/specification-tests/issues/113
  `http to https` : http not acccessible in the tested instance
- https://github.com/solid-contrib/specification-tests/issues/115 : 
   precondition `if-none-match=*` should return 412 for PUT on existing resource
- https://github.com/solid-contrib/specification-tests/issues/116
  NSS was mistakenly returning Content-Type: application/octet-stream; charset=utf-8 for HEAD on Container
  `Erroneous contentType on container`
  
Tests have been added

2. it includes also
 - APPEND for PUT on resource creation
 - control that same name do not occur on the creation of intermediate container
 
 and tests


README.md
```diff
@@ -69,6 +69,15 @@ $ solid start --root path/to/folder --port 8443 --ssl-key path/to/ssl-key.pem --
 # Solid server (solid v0.2.24) running on https://localhost:8443/
 ```
 
+By default `solid` runs in debug all mode. To stop the debug logs use `-q`, the quiet parameter.
```
TallTed, line 72:
```suggestion
By default, `solid` runs in `debug all` mode. To stop the debug logs, use `-q`, the quiet parameter.
```

README.md
```diff
@@ -69,6 +69,15 @@ $ solid start --root path/to/folder --port 8443 --ssl-key path/to/ssl-key.pem --
...
+
+```bash
+$ DEBUG="solid:*" solid start -q
+# use quiet mode and set debug to all
+# DEBUG="solid:ACL" logs only debug.ACL's
+
```
TallTed, line 78:
```suggestion
# DEBUG="solid:ACL" logs only debug.ACL's
```

TallTed (changes requested):
trivial

lib/acl-checker.js
```diff
@@ -128,19 +128,23 @@ class ACLChecker {
...
-      if (directory && this.resource.endsWith('/')) resourceAccessDenied([ACL('Read'), ACL('Write')])
-      // if resource and acl have same parent container,
-      // then Read Write from parent is required
-      else if (!directory && aclFile.value.endsWith(`/${this.suffix}`)) await accessdeniedFromParent([ACL('Read'), ACL('Write')]) // directory = rdf.sym(dirname(aclFile.value) + '/')
+      if (resourceExists) {
+        // deleting a Container
+        // without Read, the response code will reveal whether a Container is empty or not
+        if (directory && this.resource.endsWith('/')) resourceAccessDenied([ACL('Read'), ACL('Write')])
+        // if resource and acl have same parent container,
+        // then Read Write from parent is required
```
TallTed, line 136:
```suggestion
        // then both Read and Write on parent is required
```

lib/acl-checker.js
```diff
@@ -166,7 +170,9 @@ class ACLChecker {
     return `${parts.join('/')}/`
   }
 
-  // Gets the ACL that applies to the resource
+  // Gets the ACL's that applies to the resource
```
TallTed, line 173:
```suggestion
  // Gets any ACLs that apply to the resource
```

TallTed (dismissed):
minor, for humans

michielbdejong (approved):

PR #1752, opened 2024-01-30
Change translation error (500) to not acceptable (406)

csarven:
The problem is that the server returns a 500 in a situation where it tries to parse a document that it can't parse as RDF, and says "Error translating between RDF formats". This happens when the request (Accept header) includes a media type of a concrete RDF syntax, and the server attempts to parse the document using one of its RDF parsers.

This is not a 500, "unexpected condition" per se. It is an expected condition.

The server should be responding with a 406, not acceptable, because client is asking for a particular representation of the resource (using an RDF media type) and the server is unable to provide one.

I think the flow of code leading up to this can be better / refactored but I didn't want to touch it in this PR.

---

*Message to application developers*: your application can get around the current issue (500) by including `*/*` (e.g., `*/*; q=0.1`) in `Accept`, which is a reasonably useful thing to do but YMMV. Your application will then get a 200 representation (whatever the native format of the file/resource is on) instead of a 500. If this PR is merged, your application may get a 406 or 200, which is arguably better to work with than a 500.

### DIFF
--- a/lib/handlers/get.js
+++ b/lib/handlers/get.js
@@ -116,10 +116,9 @@ async function handler (req, res, next) {
 
   // If it is not in our RDFs we can't even translate,
   // Sorry, we can't help
-  if (!possibleRDFType) {
-    return next(error(406, 'Cannot serve requested type: ' + contentType))
+  if (!possibleRDFType || !RDFs.includes(requestedType)) {
+    return next(error(406, 'Cannot serve requested type: ' + requestedType))
   }
-
   try {
     // Translate from the contentType found to the possibleRDFType desired
     const data = await translate(stream, baseUri, contentType, possibleRDFType)
@@ -127,7 +126,8 @@ async function handler (req, res, next) {
     res.setHeader('Content-Type', possibleRDFType)
     res.send(data)
     return next()
-  } catch (err) {
+  }
+  catch (err) {
     debug('error translating: ' + req.originalUrl + ' ' + contentType + ' -> ' + possibleRDFType + ' -- ' + 406 + ' ' + err.message)
     return next(error(406, 'Cannot serve requested type: ' + requestedType))
   }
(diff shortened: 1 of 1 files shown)


bourgeoa:
What was your use case. Was it text/html to RDFa ?

csarven:
No, anything, that's not in that list

Try, e.g.:

`curl -iH'Accept: text/turtle' https://csarven.solidcommunity.net/public/test.rss`

500 response is given with `Content-Type: text/plain; charset=utf-8` and "Error translating between RDF formats" in response body.

This is just a case where server can't provide a Turtle representation of that resource. It should be 406.

See also:

`curl -iH'Accept: text/turtle, */*; q=0.1' https://csarven.solidcommunity.net/public/test.rss`

200 response is given with `Content-Type: application/rss+xml`

bourgeoa:
> No, anything, that's not in that list
> 
> Try, e.g.:
> 
> `curl -iH'Accept: text/turtle' https://csarven.solidcommunity.net/public/test.rss`

Then there is on other issue. 406 should already be used. See 
https://github.com/nodeSolidServer/node-solid-server/blob/83dad063955e8e42e81eb413c88710234675aa0d/lib/handlers/get.js#L117-L121

csarven:
If the selected representation's media type (what's stored on disk) is not in the list of serializable RDF media types, then no need to try to translate, and so should return 406. Only in the case when one of the RDF media types acceptable by the client that's also in the list of serializable RDF media types translation should be attempted.

bourgeoa:
That's what the actual code tries to do and for to be checked reason is not doing in L117

csarven:
What do you think of https://github.com/nodeSolidServer/node-solid-server/pull/1752/commits/65b04bbff7153d601c8b7de625d27cfc4365d336

bourgeoa:
@csarven I created a branch `translation` can you rebase your PR. Thanks

bourgeoa:
moved to https://github.com/nodeSolidServer/node-solid-server/pull/1755

PR #1754, opened 2024-02-05
Accept-Patch Accept-Post Accept-Put headers

bourgeoa:


jeff-zucker (approved):
LGTM.  Thanks for doing this.

csarven:
Are you planning to add `Accept-Put` as well (in the future)? Could it be factored into this PR? `*/*` should be fine for non-containers, and RDF types for containers (at least `text/turtle` and `application/ld+json`)

bourgeoa:
@csarven 
> Are you planning to add `Accept-Put` as well (in the future)? Could it be factored into this PR? `*/*` should be fine for non-containers, and RDF types for containers (at least `text/turtle` and `application/ld+json`)

Thanks for pointing. Just added Accept-Put
I only added `text/turtle` to non existing container (GET 404) and to `*/*` to resource documents (GET 200 or 404)
On container PUT to create do not need a contentType, and always set contentType to 'text/turtle'
Container epresentation is a different issue.

PR #1755, opened 2024-02-05
Csarven fix/translation error not acceptable

bourgeoa:


michielbdejong (approved):

michielbdejong:
Was this PR merged with failing tests? Don't we have a branch protection rule against that?

bourgeoa:
> Was this PR merged with failing tests? Don't we have a branch protection rule against that?

Yes it was.
The CI is quite unstable and prior merging I always do and did the test-suite locally (everything was OK)

Issue #1758, opened 2024-02-21
username leak. remove "user found" from login page


emmettownsend:
anyone can scrape usernames from an NSS service by using the login page, since it discloses “user found”. ([CWE - CWE-200: Exposure of Sensitive Information to an Unauthorized Actor (4.13)](https://cwe.mitre.org/data/definitions/200.html) )
![Screenshot from 2024-02-20 14-03-27](https://github.com/nodeSolidServer/node-solid-server/assets/58346537/f45ae133-6053-4329-b958-aca8f15122a9)

emmettownsend:
Davi found this defect yesterday. 
Looks like it could be fixed by just changing the error message sent back to the browser to something like...'The username and password combination is not valid'

bourgeoa:
Thanks for reporting.

PR #1759, opened 2024-02-22
fix security issue

bourgeoa:
see #1758 security issue


emmettownsend:
Thanks @bourgeoa . 
Is there going to be a new release of NSS that has the fix ? 
Cheers
Emmet

bourgeoa:
@emmettownsend 
Just published with NSS@5.7.9 see https://github.com/nodeSolidServer/node-solid-server/releases/tag/v5.7.9
Thanks for the reminder.
I was hoping to publish with next version of Mashlib, which is under review

PR #1763, opened 2024-03-05
improve consistency of the look and feel of the two buttons

ewingson:
d2024-03-05 t08:22

@bourgeoa the two buttons look different in the moment in size.
this PR improves the UI via making the size the same and consistent.


ewingson:
update after code correction:
I'm not sure why the tests fail, the code is exactly as deployed to solidweb.org where you can see a working example the buttons look the same now.
would someone with experience take care / feel responsible ?
additional : maybe line 41 should become commented out ?

zg009 (approved):
Looks like a good improvement to me.

PR #1773, opened 2024-03-15
fix issue 1770:

zg009:
changed error handling of attempting to send password reset on nonexistent user: view issue 1770

@bourgeoa , I wasn't sure what you meant by changing that one line, as I could not find a fix without changing how the accountManager worked downstream, so I simply added an error code check to if request.validate() fails it still passes through the catch.

If you have a better idea let me know and I will change + test

### DIFF
--- a/lib/requests/password-reset-email-request.js
+++ b/lib/requests/password-reset-email-request.js
@@ -93,6 +93,13 @@ class PasswordResetEmailRequest extends AuthRequest {
     return Promise.resolve()
       .then(() => request.validate())
       .then(() => request.loadUser())
+      .catch((err) => {
+        if (err.code === "ACCOUNT_MISSING") {
+          this.response.render('auth/reset-link-sent')
+        } else {
+          return err;
+        }
+      })
       .then(userAccount => request.sendResetLink(userAccount))
       .then(() => request.renderSuccess())
       .catch(error => request.error(error))
@@ -123,7 +130,9 @@ class PasswordResetEmailRequest extends AuthRequest {
     return this.accountManager.accountExists(username)
       .then(exists => {
         if (!exists) {
-          throw new Error('Account not found for that username')
+          const error = new Error("Account not found for that username")
+          error.code = "ACCOUNT_MISSING"
+          throw error
         }
 
         const userData = { username }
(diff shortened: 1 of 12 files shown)


lib/requests/password-reset-email-request.js
```diff
@@ -123,7 +130,9 @@ class PasswordResetEmailRequest extends AuthRequest {
     return this.accountManager.accountExists(username)
       .then(exists => {
         if (!exists) {
-          throw new Error('Account not found for that username')
+          const error = new Error("Account not found for that username")
+          error.code = "ACCOUNT_MISSING"
+          throw error
```
bourgeoa, line 135:
```suggestion
          // For security avoid using an error that leak information
          return renderSuccess()
```
`renderSuccess()` is may be changed to better name like resetLinkMessage()
The message itself should be changed in 
https://github.com/nodeSolidServer/node-solid-server/blob/d8290cbe3f7fe5ba7ccd46e988a1591659c6de93/default-views/auth/reset-link-sent.hbs#L17

lib/requests/password-reset-email-request.js
```diff
@@ -93,6 +93,13 @@ class PasswordResetEmailRequest extends AuthRequest {
     return Promise.resolve()
       .then(() => request.validate())
       .then(() => request.loadUser())
+      .catch((err) => {
+        if (err.code === "ACCOUNT_MISSING") {
+          this.response.render('auth/reset-link-sent')
+        } else {
+          return err;
+        }
+      })
```
bourgeoa, line 102:
```suggestion
```

bourgeoa (changes requested):

bourgeoa:
@zg009 
Do you work locally ? if yes you should  `npm run test` before push. If you need any help please ask.

Anyway after a PR it is good ppractice to come back and look a the tests. Thereafter clear any issue due to your changes.

zg009:
> @zg009 I'm sorry I have not been very clear to the changes I was thinking of.
> 
> Do you work locally ? if yes you should `npm run test` before push. If you need any help please ask.
> 
> Usually after a PR it is good practice to come back and look a the tests. Thereafter clear any issue due to your changes.

I saw, I thought I had passed the linting check but had forgotten to save, so I committed without fixing the formatting issues.

lib/requests/password-reset-email-request.js
zg009: I changed the wording and renamed it to resetLinkMessage in the latest commits

lib/requests/password-reset-email-request.js
zg009: I removed this as well

PR #1775, opened 2024-03-16
fix issue 1774

zg009:
changed account manager error checking, changed loadUser function to properly render html, updated unit test

### DIFF
--- a/lib/requests/password-reset-email-request.js
+++ b/lib/requests/password-reset-email-request.js
@@ -94,7 +94,7 @@ class PasswordResetEmailRequest extends AuthRequest {
       .then(() => request.validate())
       .then(() => request.loadUser())
       .then(userAccount => request.sendResetLink(userAccount))
-      .then(() => request.renderSuccess())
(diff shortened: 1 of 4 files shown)


bourgeoa (approved):
LGTM you can merge

edwardsph:
I think the message should say the email is sent 'to' the associated account, not from as that doesn't make sense.

bourgeoa:
Thanks. Done

lib/requests/password-reset-email-request.js
```diff
@@ -123,7 +123,10 @@ class PasswordResetEmailRequest extends AuthRequest {
     return this.accountManager.accountExists(username)
       .then(exists => {
         if (!exists) {
-          throw new Error('Account not found for that username')
+          // For security reason avoid leaking error information
```
TallTed, line 126: Merged too quickly!
```suggestion
          // For security reasons, avoid leaking error information
```

bourgeoa: Thanks

PR #1778, opened 2024-03-21
Fix/issue#1692

zg009:
See here: https://github.com/nodeSolidServer/node-solid-server/issues/1692

Added a check in put.js handler to see if a client is attempting to create an invalid resource uri

Alain, before this gets merged, let me know if there is a better way to get the RESERVED_SUFFIXES, and if the function should be moved to ldp.js instead


lib/handlers/put.js
```diff
@@ -6,12 +6,40 @@ const getContentType = require('../utils').getContentType
 const HTTPError = require('../http-error')
 const { stringToStream } = require('../utils')
 
+// TODO: ask alain a better way to get the suffix variables here
+const RESERVED_SUFFIXES = ['.acl', '.meta']
+
+/**
+ * This function is used to make sure a resource or container which contains
+ * reserved suffixes for auxiliary documents cannot be created.
```
TallTed, line 14:
```suggestion
 * This function is used to prevent creation of a resource or container
 * which contains auxiliary documents named with reserved suffixes.
```

lib/handlers/put.js
```diff
@@ -6,12 +6,40 @@ const getContentType = require('../utils').getContentType
...
+ * @param {string} path - the uri to check for invalid suffixes
+ * @returns {boolean} true is fail - if the path contains reserved suffixes
+ */
+function containsInvalidSuffixes (path) {
+  // if it is a container, no suffix so remove last slash
```
TallTed, line 19:
```suggestion
  // if it is a container, no suffix, so remove last slash
```

lib/handlers/put.js
```diff
@@ -6,12 +6,40 @@ const getContentType = require('../utils').getContentType
...
+  if (path.endsWith('/')) {
+    path = path.slice(0, -1)
+  } else {
+    // this is a resource, so it either ends with an extension, or just text
```
TallTed, line 23:
```suggestion
    // this is a resource, so it either ends with an extension, or is just text
```

TallTed (changes requested):
human-facing language tweaks for clarity

bourgeoa:
Thanks for looking at this issue.
 - Both PUT and PATCH are allowed to create intermediate Container
 - PUT and POST can create Container

The container name cannot 
1. be the same as an existing resource name
2. end with a reserved auxiliary extension, to not forbid creation of an auxiliary resource

bourgeoa:
@zg009
Not sure of POST creating a container with slug. I don't think sure such a test exist
https://github.com/nodeSolidServer/node-solid-server/blob/f5652f3dfa3ae630408125af99bcf90b7fe21c0d/lib/ldp.js#L150
Could you add a test for POST trying to create a Container with an unallowed slug

zg009:
There are tests for invalid slug on POST requests here: https://github.com/nodeSolidServer/node-solid-server/blob/f5652f3dfa3ae630408125af99bcf90b7fe21c0d/test/integration/http-test.js#L863C4-L876C7

I added the check I believe you requested

bourgeoa:
Sorry I am on a tablet. Not easy to code

- You are testing a POST creating a resource. I don't see why it should fail.
POST never fail. The name of the resource created must be checked on the location header.
Solid specification introduced an exception 409 for auxiliary resources because you cannot be unsure on the resource name in auxiliary resources

- You are not testing a POST creating a container with a slug ending with .acl
Below an exemple. Notice the link that say I want a container. This is explicitly needed because a slug cannot end with `/`
https://github.com/nodeSolidServer/node-solid-server/blob/f5652f3dfa3ae630408125af99bcf90b7fe21c0d/test/integration/http-test.js#L838

The test should create a container. The container name can be found in `location header` see next test.

Your code in `ldp.js` should do something like
```
If (container) {
  extension = ''
 If slug ends with unallowed extension remove unallowed extension from slug
}
elseif (slug) test slug to not allow creation of auxiliary --> 409
```

zg009:
Alain, I do not see how you can do what you are stating is possible.

If slug has extension and extension is .acl or .meta, 403 is thrown
```server.post('/post-resources/').set('slug', 'post-resource.acl') // throws 403```
```server.post('/post-resources/').set('slug', 'post-resource.meta') // throws 403```

If slug has no extension, it becomes container.
```server.post('/post-resources/').set('slug', 'post-resource') // creates container```

I do not know how the situation you are describing can exist.

File: test/integration/http-test.js [hunk 4]
```diff
@@ -891,12 +891,12 @@ describe('HTTP APIs', function () {
         .set('content-type', 'text/turtle')
         .expect(403, done)
     })
-    it('should error with 400 if slug contains invalid suffix', function (done) {
+    it('should not error with 400 if slug contains invalid suffix', function (done) { // TODO find better name
```
bourgeoa, line 894: TODO find a better test name

bourgeoa:
@zg009 Could you look at my changes
Basically
- the issue is only for container creation
- the check is only at the extension part (`endsWith`)
Can you 
- add a test for .meta
- TODO find a better test name on line 894

zg009:
@bourgeoa I renamed it and added some clarity on the new .meta and .acl tests

lib/ldp.js
```diff
@@ -145,18 +145,23 @@ class LDP {
...
+      extension = ''
+    }
+    // pepare slug
     if (slug) {
-      if (this.isAuxResource(slug, extension)) throw error(403, 'POST is not allowed for auxiliary resources')
       slug = decodeURIComponent(slug)
+
+      if (container) {
+        // the name of a container cannot be a valid auxiliary resource document
+        if (this._containsInvalidSuffixes(slug + '/')) slug = slug.split('.')[0]
```
bourgeoa, line 158:
```suggestion
        if (this._containsInvalidSuffixes(slug + '/')) slug = slug.split('.')[0]
```
 I made the split on first dot, it may be better to keep string before last dot or remove the invalid suffix

zg009: So that results in two outputs:

If the invalid string is ```/foo/bar.acl/test.ttl```
Then removing string before last dot wouldn't work because it would still be ```/foo/bar.acl/test.ttl```
And removing the invalid suffix would be ```foo/bar/test.ttl``` which is correct, but may cause confusion for the user since it is not what they expect. It should maybe throw an error in this case for clarity?

bourgeoa: No this code is in POST and is proposing a new valid slug. The slug is not a path, just an item name like `foo.bar.acl`

zg009: To make it recursive, can't it simply be changed from ```if (this._containsInvalidSuffixes)``` to ```while (this._containsInvalidSuffixes)``` ? as slug would stop updating when it is valid or when ```slug = ''``

zg009: > No this code is in POST and is proposing a new valid slug. The slug is not a path, just an item name like `foo.bar.acl`
> 
> An other problem is if slug is `foo.meta.acl` the code need to be recursive.
> 
> An easy solution is `slug = ''` why not this is valid.
@bourgeoa This doesn't matter because none of these conditions satisfy the container requirement in the ldp.post method.

bourgeoa: > if (container)

This is the container requirement.

zg009: @bourgeoa If you look at last 2 commits, I added example test with recursive check and example recursive code in ldp.post

lib/ldp.js
```diff
@@ -137,6 +137,7 @@ class LDP {
   }
 
   async post (hostname, containerPath, stream, { container, slug, extension, contentType }) {
+    console.log('container:', container)
```
bourgeoa, line 140:
```suggestion
```

lib/ldp.js
```diff
@@ -145,18 +146,29 @@ class LDP {
...
+    }
+    // pepare slug
     if (slug) {
-      if (this.isAuxResource(slug, extension)) throw error(403, 'POST is not allowed for auxiliary resources')
       slug = decodeURIComponent(slug)
+
+      if (container) {
+        // the name of a container cannot be a valid auxiliary resource document
+        while (this._containsInvalidSuffixes(slug + '/')) {
+          console.log('splitting slug', slug)
```
bourgeoa, line 160:
```suggestion
```

lib/ldp.js
```diff
@@ -145,18 +146,29 @@ class LDP {
...
+          // slug = slug.split('.')[0]
+          const idx = slug.lastIndexOf('.')
+          slug = slug.substr(0, idx)
+          console.log('new slug', slug)
```
bourgeoa, line 164:
```suggestion
```

bourgeoa (changes requested):
some console in lib/ldp.js to remove

bourgeoa:
Thanks LGTM just remove the 2 console.log() in the recursive

bourgeoa (approved):
LGTM

lib/ldp.js
```diff
@@ -145,18 +145,26 @@ class LDP {
...
-      if (this.isAuxResource(slug, extension)) throw error(403, 'POST is not allowed for auxiliary resources')
       slug = decodeURIComponent(slug)
+
+      if (container) {
+        // the name of a container cannot be a valid auxiliary resource document
+        while (this._containsInvalidSuffixes(slug + '/')) {
+          const idx = slug.lastIndexOf('.')
+          slug = slug.substr(0, idx)
+        }
+      } else if (this.isAuxResource(slug, extension)) throw error(403, 'POST is not allowed for auxiliary resources')
```
TallTed, line 162:
```suggestion
      } else if (this.isAuxResource(slug, extension)) throw error(403, 'POST to auxiliary resources is not allowed')
```

lib/ldp.js
```diff
@@ -145,18 +145,26 @@ class LDP {
...
+
       if (slug.match(/\/|\||:/)) {
         throw error(400, 'The name of new file POSTed may not contain : | or /')
```
TallTed, line 165:
```suggestion
        throw error(400, 'The name of a POSTed new file may not contain ":" (colon), "|" (pipe), or "/" (slash)')
```

test/integration/http-test.js
```diff
@@ -670,6 +670,23 @@ describe('HTTP APIs', function () {
           .expect(201, done)
       }
     )
+    it('should return a 400 error when trying to put a container that contains a reserved suffix',
```
TallTed, line 673:
```suggestion
    it('should return a 400 error when trying to PUT a container with a name that contains a reserved suffix',
```

test/integration/http-test.js
```diff
@@ -670,6 +670,23 @@ describe('HTTP APIs', function () {
...
+      function (done) {
+        server.put('/foo/bar.acl/test/')
+          .set('content-type', 'text/turtle')
+          .set('link', '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"')
+          .expect(400, done)
+      }
+    )
+    it('should return a 400 error when trying to put a resource that contains a reserved suffix',
```
TallTed, line 681:
```suggestion
    it('should return a 400 error when trying to PUT a resource with a name that contains a reserved suffix',
```

test/integration/http-test.js
```diff
@@ -935,19 +970,50 @@ describe('HTTP APIs', function () {
...
+        .set('slug', 'loans.acl.meta')
+        .set('link', '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"')
+        .send(postRequest2Body)
+        .expect('location', /\/post-tests\/loans\//)
+        .expect(201)
+        .end((err, res) => {
+          if (err) return done(err)
+          try {
+            postLocation = res.headers.location
+            assert(!postLocation.endsWith('.acl/') && !postLocation.endsWith('.meta/'), 'Container name should not end with .acl or .meta')
```
TallTed, line 1006:
```suggestion
            assert(!postLocation.endsWith('.acl/') && !postLocation.endsWith('.meta/'), 'Container name should not end with ".acl" or ".meta"')
```

TallTed (changes requested):
human facing, relatively small

test/integration/http-test.js
```diff
@@ -1003,7 +1003,7 @@ describe('HTTP APIs', function () {
           if (err) return done(err)
           try {
             postLocation = res.headers.location
-            assert(!postLocation.endsWith('.acl/') && !postLocation.endsWith('.meta/'), 'Container name should not end with .acl or .meta')
+            assert(!postLocation.endsWith('.acl/') && !postLocation.endsWith('.meta/'), 'Container name should not end with ".acl" or ".meta"')
```
TallTed, line 1006: Realizing this is more of a stricture, than a recommendation.
```suggestion
            assert(!postLocation.endsWith('.acl/') && !postLocation.endsWith('.meta/'), 'Container name cannot end with ".acl" or ".meta"')
```

zg009:
@TallTed If it looks good to you, can you go ahead and approve the changes and I'll merge

TallTed (approved):
OK bye me. Note that I code primarily in English, and cannot provide much if any useful input on the JavaScript or other code here.

zg009:
@TallTed That is OK. It is nice to have more input regarding clarity, when a test or code change has to be revisited later.

PR #1785, opened 2024-05-24
Update put.js send 200 or 204 depending on pre-existance of resource

jeff-zucker:
The HTTP spec says that the Server MUST return a 200 or 204 if PUT replaces and existing resource.  See second paragraph of https://httpwg.org/specs/rfc9110#PUT.

Brought up by @CxRes.

### DIFF
--- a/lib/handlers/put.js
+++ b/lib/handlers/put.js
@@ -62,7 +62,7 @@ async function putStream (req, res, next, stream = req) {
   // try {
   // Obtain details of the target resource
   let resourceExists = true
-  let returnStatus = 201;
+  let returnStatus = 201
   try {
     // First check if the file already exists
     await ldp.resourceMapper.mapUrlToFile({ url: req })
@@ -71,7 +71,7 @@ async function putStream (req, res, next, stream = req) {
       res.sendStatus(412)
       return next()
     }
-    returnStatus = 200;
+    returnStatus = 200
   } catch (err) {
     resourceExists = false
   }
--- a/lib/handlers/put.js
+++ b/lib/handlers/put.js
@@ -79,7 +79,7 @@ async function putStream (req, res, next, stream = req) {
     // Fails with Append on existing resource
     if (!req.originalUrl.endsWith('.acl')) await checkPermission(req, resourceExists)
     await ldp.put(req, stream, getContentType(req.headers))
-    res.sendStatus(returnStatus)
+    res.sendStatus(resourceExists ? 204 : 201)
     return next()
   } catch (err) {
     err.message = 'Can\'t write file/folder: ' + err.message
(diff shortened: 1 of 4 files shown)


zg009:
@jeff-zucker Just to confirm, should the successful PUT return a 200 or a 204, or the 200 and 201 as you changed?

jeff-zucker:
> @jeff-zucker Just to confirm, should the successful PUT return a 200 or a 204, or the 200 and 201 as you changed?

My understanding is that 

* when PUT creates a new file, return 201
* when PUT replaces an existing file and content is sent back, return 200
* when PUT replaces an existing file and no content is sent back, return 204

I don't know in what circumstances, if any, NSS returns content from a PUT.  If  that never happens, my PR should be changed to make 204 the response when file pre-exists.  If there are times when NSS does return content, those times should trigger a change from 204 to 200.

CxRes:
> I don't know in what circumstances, if any, NSS returns content from a PUT.

If that is the case, changing 201 to 204 is adequate. 

I would update this, but I to have @jeff-zucker do it seems easier.

jeff-zucker:
>  I would update this, but I to have @jeff-zucker do it seems easier.

Done.

zg009:
The tests are passing but I'll wait for Alain's input.

CxRes:
@zg009 Some tests are failing, but that is mostly a simple matter of changing 201 to 204 status codes. What is non-trivial here is that we need to add a few more test cases to distinguish between 201 and 204 cases (i.e. we want coverage for both), which will add a few more tests.

zg009:
@CxRes Can you tell me which tests or the output? For some reason all of mine are still passing. I need to check if there is something wrong with my local repository or if it is another issue.

bourgeoa:
@zg009 You can see this is the GitHub CI.
I suppose your local version needs a git pull

lib/handlers/put.js
```diff
@@ -70,14 +71,15 @@ async function putStream (req, res, next, stream = req) {
...
+    returnStatus = 204 // if content is sent back change to 200
   } catch (err) {
     resourceExists = false
   }
   try {
     // Fails with Append on existing resource
     if (!req.originalUrl.endsWith('.acl')) await checkPermission(req, resourceExists)
     await ldp.put(req, stream, getContentType(req.headers))
-    res.sendStatus(201)
+    res.sendStatus(returnStatus)
```
bourgeoa, line 82:
```suggestion
    res.sendStatus(resourceExists ? 204 : 201)
```
This should be a simplification. Needs to remove `returnStatus` lines 65 and 74
14 current tests to be updated from 201 to 204
Tested the test-suite and the conformance tests. It is OK

CxRes (approved):

PR #1786, opened 2024-05-31
fix: Return 201 when PATCH creates a new resource

CxRes:
A PATCH request may end up creating a new resource, in which case a 201 status code should be returned.

This is an equivalent change as that for PUT in #1785. 

This does not update or add tests, please fix that!

### DIFF
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -91,8 +91,8 @@ async function patchHandler (req, res, next) {
       return writeGraph(graph, resource, ldp.resourceMapper.resolveFilePath(req.hostname), ldp.serverUri)
     })
 
-    // Send the result to the client
-    res.sendStatus(resourceExists ? 200 : 201)
+    // Send the status and result to the client
+    res.status(resourceExists ? 200 : 201)
     res.send(result)
   } catch (err) {
     return next(err)
--- a/lib/handlers/patch.js
+++ b/lib/handlers/patch.js
@@ -92,6 +92,7 @@ async function patchHandler (req, res, next) {
     })
 
     // Send the result to the client
+    res.sendStatus(resourceExists ? 200 : 201)
     res.send(result)
   } catch (err) {
     return next(err)
--- a/test/integration/acl-oidc-test.js
+++ b/test/integration/acl-oidc-test.js
@@ -557,7 +557,7 @@ describe('ACL with WebID+OIDC over HTTP', function () {
       options.headers['content-type'] = 'application/sparql-update'
       request.patch(options, function (error, response, body) {
         assert.equal(error, null)
-        assert.equal(response.statusCode, 200)
+        assert.equal(response.statusCode, 201)
         done()
       })
     })
@@ -616,7 +616,7 @@ describe('ACL with WebID+OIDC over HTTP', function () {
       options.headers['content-type'] = 'application/sparql-update'
       request.patch(options, function (error, response, body) {
         assert.equal(error, null)
-        assert.equal(response.statusCode, 200)
+        assert.equal(response.statusCode, 201)
         done()
       })
     })
--- a/test/integration/patch-sparql-update-test.js
+++ b/test/integration/patch-sparql-update-test.js
@@ -22,7 +22,7 @@ describe('PATCH through application/sparql-update', function () {
     server.patch('/notExisting.ttl')
       .set('content-type', 'application/sparql-update')
       .send('INSERT DATA { :test  :hello 456 .}')
-      .expect(200)
+      .expect(201)
       .end(function (err, res, body) {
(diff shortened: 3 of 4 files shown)


bourgeoa:
Thanks for raising this issue.
There are some issue to resolve in /test-suite/solid-crud-tests on pub/sub tests

bourgeoa:
@CxRes 
I think we need some confirmation on specification see this discussion https://github.com/solid/specification/issues/14#issuecomment-683480525

CxRes:
I read the cited issue/comment. I think there is typo in the PATCH table, it should be 200 and 201 for exists and not exists respectively, just like the PUT table. @csarven?

csarven (commented):
tl;dr: it depends.

Generally speaking 201 better characterises the creation of a new resource but 200 is not wrong either. It does fundamentally depend on what the server prefers to respond to a particular request in the context of the resource semantics.

There are other related considerations: cases in which no Read access permission is given to the target resource, 201 would reveal that the resource did not exist before whereas 200 would not reveal it. Related: https://github.com/solid/specification/issues/311

If these distinctions are not taken into account or there is no preference/intention to be very "strict" about the status code in terms of revealing the existence of a resource, then 201 will suffice, otherwise stick to 200 (if anything but for the time being.)

CxRes:
Maybe one more clarification for this, if the resource is configured such that the user-agent had read access, would at least in those cases 201 be the appropriate code?

bourgeoa:
> I don't want to rehash https://github.com/solid/specification/issues/311 and friends, but I am unclear why revealing the prior non-existence of the resource to a user-agent that already has write access is a problem?

There has been lengthy discussions and it has been decided the following for PATCH
https://github.com/nodeSolidServer/node-solid-server/blob/843c7718cd366599272905c8e3a57f7af2ef0479/lib/handlers/patch.js#L153-L157

CxRes:
@bourgeoa Excuse my ignorance, but what is WHERE?

csarven:
I think Add would be the right activity when the container expands.

Yes, re comment in 311:

>Noting that the use cases to write-only are rare, so in practice both read and write would be given to allow a write operation. However, this is orthogonal to how authorization rules could be set.

We can of course choose 201 for PATCH as with PUT (and yes, it would be appropriate too with the Read case on C/ or C/R), and that would generally be better, but 200 is not ruled out. 202 or 204 would also be okay. I was working with 1) "200 responses to state-changing methods" which is applicable to PATCH 2) PATCH itself didn't make the 200/201 distinction as PUT has 3) giving a bit more attention to authorization / security considerations. Some things didn't add up to me - which is why some see the ramblings as "philosophy".. and some others may look at it in terms of security - we can't pretend that the protocol is all "secure" if we also don't pay attention to minute differences between 200 and 201. See also last paragraph in https://github.com/solid/specification/issues/14#issuecomment-1272166350

CxRes:
@bourgeoa Given that all resources that are being PATCHed on NSS will necessarily have read access on them and @csarven agrees that 201 is an acceptable response in those case, I think we can safely accept and close this PR.

It might be worth considering if NSS might want to allow one to write without read access on certain resources, in which case we will want to revisit and split the response code between 200 and 201 as appropriate.

csarven (approved):
I haven't tested the contents of this PR, so I can't vouch for whether it works as intended or breaks anything. The review is based on what's intended and the discussion in this PR (and elsewhere).

Approving with the understanding that 201 is a valid, but not the only, status code when a resource is created. If the server is also taking other things into account, such as whether the client has the necessary permissions to know the existence of the target resource, as well as the permission to create the resource, then 201 is indeed suitable. Again, the use of 200 is not wrong, and its use could perhaps even be better with `Content-Location` and payload including a representation, but perhaps that's a separate discussion.

jeff-zucker:
I also don't see a problem with sending 201.  The case of something bad happening when a person with write privs but not read privs finding out that the thing didn't exist before they created it seems unlikely and far-fetched.

melvincarvalho:
We had a call on SolidOS yesterday, and this issue came up.  Is there anything still blocking the merge?

michielbdejong:
@CxRes NSS runs the web-access-control-tests from a branch:

https://github.com/solid-contrib/web-access-control-tests/blob/patchAppendNewDocument/test/surface/create.test.ts#L389

If you make a change there in that branch and then retrigger the tests of this PR, I think it should work

timbl:
As discussed on call today.  Overriding as the test are the wrong version of the tests.. except I don't have priv to do so.

CxRes:
@michielbdejong opened PR <https://github.com/solid-contrib/web-access-control-tests/pull/57> on the specific branch. Please review and merge! 

I am not sure if I can trigger a test without the PR!

PR #1801, opened 2024-11-12
Basic prep fix 1

bourgeoa:


bourgeoa:
LGTM
and conformance tests runed on https://solidcommunity.net:8443

CxRes:
Just waiting for @csarven to check the final commit.

CxRes:
Nothing. @bourgeoa prefers it this way, as it allows him to run CTH with solidcommunity.net when the branch is in the repo. And then he merges that.

bourgeoa:
Yes but not only.
It is in general a bad idea to merge directly in main from an outside branch.
There may be issues and main may be polluted, then you may need to revert.

PR #1808, opened 2025-02-11
replace res.set()

bourgeoa:


CxRes:
This is OK for consistency! But I do not understand why it solves the problem.

`res.set()` is a thin wrapper around `res.setHeader()` with only [minor conveniences](https://github.com/expressjs/express/blob/85e48bb8c109d2200077bf8d50f3a163488d0fa5/lib/response.js#L655) which should not matter because the value (2nd argument to `res.set()`) is already a string.

lib/handlers/get.js
```diff
@@ -118,23 +118,25 @@ async function handler (req, res, next) {
...
     }
+    let statusCode = 200
     if (contentRange) {
       headers = {
         ...headers,
         'Content-Range': contentRange,
         'Accept-Ranges': 'bytes',
         'Content-Length': chunksize
       }
-      res.statusCode = 206
```
CxRes, line 128: This needs to be set here on the response because when PREP activates, the status code is still 206 on the response. This code will mean PREP will always use the default `res.statusCode` value of 200 when sending a response.

CxRes (changes requested):

CxRes:
After further testing, I believe the problem happens because `res.set` is setting the `charset` parameter in the `Content-Type` header field, which is allowed. This is an issue therefore with mashlib which fails to render markdown, presumably because it is doing an exact string match to "text/markdown" and failing that rendering text.  

I would revert all changes except L104 and L132, and look for why the render is misunderstanding the media-type.

CxRes (approved):

CxRes:
Made some changes, this will work for now.

Issue #1807, opened 2025-02-10
Since PREP text/markdown is rendered as plain text


bourgeoa:
@CxRes 
Since PREP `text/markdown` resources are rendered as plain text

The image below is from localhost with NSS@latest. I also tried with `src/handlers/get.js` reverted to v5.7.11 same issue
You can  check that https://solidcommunity.net as the same issue

![Image](https://github.com/user-attachments/assets/80087c43-46b1-421a-b072-bc4e17d5e545)

PR #1809, opened 2025-02-14
res.set(headers) and mashlib@1.10.4

bourgeoa:
res.set(headers) sends `charset` parameter
This is now removed in mashlib solid-panes when checking content-type
see #1807


CxRes (approved):
This comment is unnecessary and can be removed! `text/*` should [almost always](https://medium.com/@lefloh/when-to-use-a-charset-attribute-for-a-content-type-header-ffeb5838d3c3) require `charset`. In case of `text/markdown` it is [mandatory](https://www.iana.org/assignments/media-types/text/markdown).

PR #1854, opened 2026-05-06
Add storage description link discovery

### LINKED ISSUE #1805: Implement storage description resource discovery
https://solidproject.org/ED/protocol#server-storage-description details the discovery of the storage description resource.

>Servers MUST include the Link header field with rel="http://www.w3.org/ns/solid/terms#storageDescription" targeting the URI of the storage description resource in the response of HTTP GET, HEAD and OPTIONS requests targeting a resource in a storage.

---

https://github.com/solid/specification/issues/355#issuecomment-1140508784 shows how an application (e.g., https://dokie.li/ ) makes use of such resource to discover the features of the storage. Some common use cases are mentioned in https://github.com/solid/specification/issues/355#use-cases

SarthakDudhe:
Add `storageDescription` Link headers for GET, HEAD, and OPTIONS requests.
Point the relation to the root storage metadata resource (`/.meta`).
Cover resources and containers in HTTP integration tests.

Fixes #1805.

## Tests
npm run lint
npm run mocha-integration


csarven (commented):
Great, thanks for doing this. I haven't run the tests but this generally looks good to me. I've run the code locally and it is referring to the expected URL for storage description resource.

Does this code assume that storage root is always at `req.hostname + '/'`?

I can't remember now but does NSS allow non-root path to be storage root? If so, does the code take that into account?

bourgeoa:
> I can't remember now but does NSS allow non-root path to be storage root? If so, does the code take that into account?

NSS only allow subdomain. Storage starts at root path.

csarven:
Ok, great. If you review / approve, we should merge this.

bourgeoa:
@csarven @SarthakDudhe 

From https://solidproject.org/TR/protocol#storage-description-statements
I think that root/.meta must include the triple 
```
</> a <http://www.w3.org/ns/pim/space#Storage> .
``` 

Which is not actually the case.

If this is correct it implies to 
- update default root/.meta
- prepare a migration script for existing servers

csarven:
In NSS, it'd be (the slash for the storage resource):

`</> a <http://www.w3.org/ns/pim/space#Storage> .`

I don't understand why a migration script is necessary (although I understand it'd be useful to do).

bourgeoa:
Thanks yes you are right I have updated above.
Why a migration ? This is because the triple is created at pod creation. On existing server you need to update all existing pods.
We add the same issue when Solid spec introduced `pim:storage` or `solid:oidcIssuer` in profile/card

csarven:
@bourgeoa Could you please confirm if NSS only allows storage root to be / after hostname? If so, that should address my inquiry in my earlier comment.

bourgeoa:
NSS only allow subdomain pods.

csarven:
@bourgeoa Anything else remaining besides the migration script to include `a pim:Storage` in storage description (`/.meta`)? This PR only covers the storage description link discovery. Could the Storage statement be left for a separate PR?

bourgeoa:
@SarthakDudhe @csarven 
what is missing is updating the templates files 
- default-templates/new-account/.meta
- config/templates/new-account/.meta

PR #1856, opened 2026-05-06
Add Last-Modified header for GET and HEAD

### LINKED ISSUE #1793: Server should include the Last-Modified header
>Server MUST generate a Last-Modified header field in response to GET and HEAD requests.

https://solidproject.org/ED/protocol#server-last-modified-includes

>Servers are strongly discouraged from exposing information beyond the minimum amount necessary to enable a feature.

https://solidproject.org/ED/protocol#exposing-information

So, the implementation should only expose the `Last-Modified` header when access is granted to the request target, i.e., when user or public has Read access to the resource. This is typically aligned with the 200 response.

SarthakDudhe:
Summary
- add `Last-Modified` headers to successful GET and HEAD responses
- use the file or container `mtime` as the header value
- add integration coverage for both resources and containers

Fixes #1793.
 Tests
- npm run mocha-integration
- npm run lint


csarven (changes requested):
Thanks for this. Tested and seems to work as intended. I haven't run the tests.

csarven (approved):

bourgeoa:
@SarthakDudhe Thanks for working on NSS. This is really needed. We may have discussion on what you could work on : implementing the Solid notifications spec !!!! or replacing inrupt OIDC client browser with UVDSL one.

@csarven
I think none of the solid servers does this.
The information is available in the Container not at Resource level
CSS does load the information in .meta (it was added for non filesystem to be able to render in Container).
This has made .meta only Patchable which is not Solid compliant.

csarven:
@bourgeoa , it is required in Solid Protocol. It is also a server-controlled header, as opposed to a resource description, which may or may not be modifiable. The other advantage is being able to get this information using `HEAD` instead of fetching the whole resource payload with `GET`.

PR #1870, opened 2026-07-28
Remove unused resource parameter from getPermissionsFor

### LINKED ISSUE #973: WAC-Allow doesn't seem to depend on resource
Looking at the code in 
https://github.com/solid/node-solid-server/blob/adb0691a787d65e41cfd9e32d4496d8a0b4f98ed/lib/header.js#L132
it doesn't seem like #246 was implemented with `resource`, because that parameter doesn't seem to be used.

NaderMohamed325:
## What
Removes the unused `resource` parameter from `getPermissionsFor` in `lib/header.js`.

## Why
`resource` is passed into `getPermissionsFor` but never used inside the function body — 
`acl.can(user, mode)` doesn't take a resource argument. As confirmed in #973, the 
resource is already implicit in the `acl` object itself (scoped via 
`ACLChecker.createFromLDPAndRequest`, see `lib/acl-checker.js`), so passing `resource` 
again here is dead code.

Closes #973

## Testing
Ran the full test suite locally on Windows. 4 pre-existing failures unrelated to this 
change (CRLF/LF line-ending mismatches in OIDC integration tests, and a signup page 
timeout) — confirmed these fail identically with this change stashed out.


NaderMohamed325:
Hi @bourgeoa / @kjetilk / @RubenVerborgh — this addresses #973 that one of you 
already confirmed. Happy to make any changes if needed, just flagging in case 
it fell off the radar.

PR #1871, opened 2026-08-07
Remove bootstrap dependency

### LINKED ISSUE #1867: remove bootstrap dependency
NSS only use bootstrap icons. We may use SolidOS icons
We use bootstrap@4 latest open source license. This version has a vulnerability

https://matrix.to/#/!wAfwwonbRnRLYejFOR:gitter.im/$xFAiECLG1apc87yny74sGTrW87sA0cI6UNvSl5ctVR4?via=gitter.im&via=matrix.org&via=chat.semantic.works

``` quote
Wednesday
* Noel De Martin
I use icons like this https://icones.js.org/
* Noel De Martin
So basically I look for the icons I want to use in that website, and then in my code I just write the <i- element (I don't have to import any library or anything), then when the app is built the SVGs are auto-replaced :)
Here's an example of using that in Umai: https://github.com/NoelDeMartin/umai/blob/a0d2887e8f70d93ee6f3b3d33b39452cea5874c9/src/components/AppHeaderUserMenu.vue#L18
and regarding colors, they use fillCurrent by default, so they'll have the same color as the text. It can also be overriden by setting the text color with CSS
* Tim Berners-Lee
Our icons are mostly from https://thenounproject.com/
* Noel De Martin
I see, it seems like most of them are plain SVGs (not part of any icon set). You can also add custom SVGs using this, for example I have a few here: https://github.com/NoelDeMartin/umai/tree/main/src/assets/icons
I'll see if I can integrate something like this in SolidOS :). It's also better to use icons from an icon set in order to keep a consistent visual identity, grabbing random SVGs from websites can create some UI inconsistencies
* Timea
this is because Sharon downloaded them from the new design.
I could ask the designer if they use some icon Library....
at the top of the meeting minutes find a link to a ticket where our current designs are. it's all figma.
* Tim Berners-Lee
I don't see why NounProject is not considered an icon set .. it has too many alternatives?
* Tim Berners-Lee
The icons are in https://github.com/SolidOS/solid-ui/tree/main/src/icons ... so we use consistent icons across SolidOS and also apps.
* Noel De Martin
* Tim Berners-Lee
I don't see why NounProject is not considered an icon set .. it has too many alternatives?
yeah, it's not an icon set, it's just a marketplace where anyone can upload their SVGs. Icon sets usually have a consistent style, etc. It's like using a font family, you can use as many font families as you want in a website, but it's usually better to choose a couple that look well together.
* Timea
it's just about the designers choice and the look of them.i can see what they say.
```

mmustafasenoglu:
Closes #1867

Bootstrap 3.4.1 has known vulnerabilities and NSS only uses it for basic CSS layout (grid, panels, forms, buttons, alerts) plus a single `glyphicon-remove` icon. This replaces the entire dependency with a lightweight custom stylesheet.

Changes:
- `common/css/solid.css`: Added replacement CSS for all Bootstrap 3 classes used across templates (grid, panels, forms, buttons, alerts, progress, visibility, typography)
- `common/css/bootstrap.min.css`: Deleted
- `lib/create-app.mjs`: Removed bootstrap node_module route handlers
- `package.json`: Removed `bootstrap` dependency
- 72 template files: Dropped `<link>` to bootstrap.min.css

The `glyphicon-remove` icon is replaced with an inline SVG data URI, so no font files are needed either.

Tests pass (818 passing, 3 pre-existing EADDRINUSE failures).


bourgeoa:
@mmustafasenoglu Thanks for your work
Can you check Copilot comments

mmustafasenoglu:
@bourgeoa Done! All 5 Copilot comments have been addressed:

1. **col-md-\* responsive grid**: Wrapped `.col-md-*` width/float rules inside `@media (min-width: 992px)` to preserve Bootstrap 3's responsive stacking behavior below 992px.

2. **Progress bar color variants**: Added `.progress-bar-success`, `.progress-bar-warning`, and `.progress-bar-danger` definitions so the password strength meter colors work correctly.

3. **Form validation states**: Added `.has-success`, `.has-warning`, and `.has-error` styles including `.form-control` border colors, focus shadows, `.help-block` colors, and feedback icon colors.

4. **Glyphicon icons**: Added `.glyphicon-ok` and `.glyphicon-warning-sign` SVG replacements alongside `.glyphicon-remove`. Switched from hardcoded `fill` colors to `currentColor` so the icon color inherits from the parent `.has-*` state.

5. **flex-wrap fix**: Changed the invalid `flex-wrap: row wrap` to `flex-wrap: wrap`.

bourgeoa:
@mmustafasenoglu why did you close.
I cannot merge your work anymore.

